PR #7, opened 2023-02-23
add some types

BilligsterUser:


src/CashuMint.ts
```diff
@@ -1,4 +1,5 @@
 import axios from "axios";
+import { Proof } from './model/Proof';
```
gandlafbtc, line 2: because of the way the js gets transpiled, the import has to be ./model/Proof.js

check other files like CashuWallet.ts for reference

src/CashuWallet.ts
```diff
@@ -78,16 +78,14 @@ class CashuWallet {
...
     async receive(encodedToken: string): Promise<Array<Proof>> {
-        const { proofs, mints } = getDecodedProofs(encodedToken)
-        const amount = proofs.reduce((total, curr) => {
-            return total + curr.amount
-        }, 0)
+        const { proofs } = getDecodedProofs(encodedToken)
+        const amount = proofs.reduce((total, curr) => total + curr.amount, 0)
         const { payload, amount1BlindedMessages, amount2BlindedMessages } = await this.createSplitPayload(0, amount, proofs)
         const { fst, snd } = await this.mint.split(payload)
+        if(!fst || !snd) { return [] }
```
gandlafbtc, line 85: not sure about this one. maybe better to split this up into two different statements, and only set the given proof array to empty if needed and return at the end.  thoughts?

gandlafbtc (changes requested):
Thank you for the PR! I only request some minor changes.

Are you already in the CASHU R&D telegram group? Feel free to join us if not! you can send me a DM @gandlaf21 on how to join

gandlafbtc (approved):
Thanks! Looks good now :+1:

PR #13, opened 2023-02-26
Extract types

### LINKED ISSUE #11: Extracting reusable types to their own interfaces
I suggest extracting reusable types to their own interfaces to improve the reusability and maintainability of our codebase.

### Proposed Changes
- Identify common types that are used across multiple components or modules
- Extract these types to their own interfaces
- Update existing code to use the new interfaces

### Additional Information
- This change will improve code reusability and maintainability
- We can avoid duplicating code by sharing interfaces across files

### Tasks
- [x] Identify common types to extract
- [x] Create new interfaces for these types
- [x] Update existing components and modules to use the new interfaces

BilligsterUser:
closes #11

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -1,5 +1,6 @@
 import axios from "axios";
 import {
+	CheckSpendablePayload,
     CheckSpendableResponse, MeltPayload, MeltResponse,
     MintKeys, requestMintResponse, SerializedBlindedMessage,
     SerializedBlindedSignature, SplitPayload, SplitResponse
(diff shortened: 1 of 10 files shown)


BilligsterUser:
Naming things is hard 😅

src/CashuMint.ts
```diff
@@ -1,6 +1,9 @@
 import axios from "axios";
-import { Proof } from "./model/Proof.js";
-import { SerealizedBlindedSignature } from "./model/types/SerealizedBlinedSignature.js";
+import {
```
gandlafbtc, line 2: Man, sorry it's my fault, but it should be serialized, with i instead of e

gandlafbtc: https://en.m.wikipedia.org/wiki/Serialization

can also use a different prefix altogether, don't know  how fitting it is.

src/model/types/index.ts
```diff
@@ -0,0 +1,49 @@
+import { Proof } from '../Proof.js'
```
gandlafbtc, line 1: nice! that's better here.

Before we merge, we could add types for the payloads in here

gandlafbtc (commented):
🔥

src/model/types/index.ts
BilligsterUser: I added a CheckSpendablePayload type I don't think there is much sense to add types for the other payloads.

PR #24, opened 2023-03-07
Return unused proofs as change in send function

### LINKED ISSUE #22: Return unused proofs as change in send function
https://github.com/cashubtc/cashu-ts/blob/main/src/CashuWallet.ts?plain=1#L96

It might be better to return proofs that enter the method but are not sent to the mint with the change

BilligsterUser:
fixes #22


BilligsterUser:
done

gandlafbtc:
you're  a machine. beep-boop

gandlafbtc (approved):
stellar

Issue #10, opened 2023-02-25
Error Handling


gandlafbtc:
Right now, error handling is basically non existent. 

Define a strategy and implement it uniformly for the whole library

BilligsterUser:
can we close this?

gandlafbtc:
yes, thanks

Issue #4, opened 2023-02-06
Create remaining unit tests


gandlafbtc:
A lot of the logic still has no unit tests. Specifically the CashuWallet class logic should have a better test coverage

BilligsterUser:
can we close this?

PR #25, opened 2023-03-09
Wallet tests

BilligsterUser:
#4 
#10

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -1,16 +1,18 @@
 import { axios } from './axios.js';
 import {
+	ApiError,
 	CheckSpendablePayload,
 	CheckSpendableResponse,
 	MeltPayload,
 	MeltResponse,
 	MintKeys,
-	requestMintResponse,
+	RequestMintResponse,
 	SerializedBlindedMessage,
 	SerializedBlindedSignature,
 	SplitPayload,
 	SplitResponse
 } from './model/types/index.js';
+import { checkResponse, checkResponseError, isObj } from './utils.js';
 
 /**
  * Class represents Cashu Mint API.
@@ -31,19 +33,30 @@ class CashuMint {
 		}
 	}
 
-	async requestMint(amount: number): Promise<requestMintResponse> {
-		const { data } = await axios.get<requestMintResponse>(`${this.mintUrl}/mint`, {
+	async requestMint(amount: number): Promise<RequestMintResponse> {
+		const { data } = await axios.get<RequestMintResponse>(`${this.mintUrl}/mint`, {
 			params: { amount }
 		});
 		return data;
 	}
 	async mint(payloads: { outputs: Array<SerializedBlindedMessage> }, paymentHash = '') {
-		const { data } = await axios.post<{
-			promises: Array<SerializedBlindedSignature> | { error: string };
-		}>(`${this.mintUrl}/mint`, payloads, {
-			params: { payment_hash: paymentHash }
-		});
-		return data;
+		try {
+			const { data } = await axios.post<
+				{
+					promises: Array<SerializedBlindedSignature>;
+				} & ApiError
+			>(`${this.mintUrl}/mint`, payloads, {
+				params: { payment_hash: paymentHash }
+			});
+			checkResponse(data);
+			if (!isObj(data) || !Array.isArray(data?.promises)) {
+				throw new Error('bad response');
+			}
+			return data;
+		} catch (err) {
+			checkResponseError(err);
+			throw err;
+		}
 	}
 
 	async getKeys(): Promise<MintKeys> {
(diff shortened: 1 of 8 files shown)


gandlafbtc (commented):
nice. Have you seen that there are 2 separate tsconfig files? `tsconfig.es5.json` will not need the `esModuleInterop` option? I'm honestly lost with the ts transpiler options

BilligsterUser:
yes sorry

BilligsterUser:
@gandlafbtc 
Should I still work on this?
Do you now work on this?

gandlafbtc:
I'll review it, i thought you're still working on it

BilligsterUser:
finally finished

BilligsterUser:
yeah

gandlafbtc:
@BilligsterUser Since we changed the tsconfigs, have you tested if it still works for both node and browser builds?

BilligsterUser:
rebased,fixed and tested

gandlafbtc:
one more rebase bls :crying_cat_face:

BilligsterUser:
done

src/CashuMint.ts
```diff
@@ -57,26 +70,63 @@ class CashuMint {
...
+		}
 	}
 	async melt(meltPayload: MeltPayload): Promise<MeltResponse> {
-		const { data } = await axios.post<MeltResponse>(`${this.mintUrl}/melt`, meltPayload);
-		return data;
+		try {
+			const { data } = await axios.post<MeltResponse>(`${this.mintUrl}/melt`, meltPayload);
+			checkResponse(data);
+			checkResponse(data);
+			if (!isObj(data) || typeof data?.paid !== 'boolean' || typeof data?.preimage !== 'string') {
```
gandlafbtc, line 90: since `change` is optional, we don't check it. I think that's ok, we could check it's format if it exists, but I don't think it is strictly necessary

src/CashuWallet.ts
```diff
@@ -35,14 +35,13 @@ class CashuWallet {
 	 * @param proofs
 	 * @returns
 	 */
-	async checkProofsSpent(proofs: Array<Proof>): Promise<Array<Proof>> {
+	async checkProofsSpent(proofs: Array<Proof>, isSpendable = false): Promise<Array<Proof>> {
```
gandlafbtc, line 38: is this so we can filter in both ways? if yes, maybe it would make sense to rename the function, since the function reads "check for spent proofs". I agree the naming is not great to begin with. But maybe the added complexity makes it even harder to understand what's going on. Or at the very least we could add it to the ts-doc comment, explaining the aditional param and update what the function returns. I'm leaning towards the latter, since we don't need to change exposed function signatures

src/CashuWallet.ts
```diff
@@ -94,22 +93,18 @@ class CashuWallet {
...
+			amount1BlindedMessages.rs,
+			amount1BlindedMessages.secrets,
+			this.keys
+		);
+		const proofs2 = dhke.constructProofs(
+			snd,
+			amount2BlindedMessages.rs,
+			amount2BlindedMessages.secrets,
+			this.keys
+		);
```
gandlafbtc, line 107: huh? why did we have that? did the behaviour not change? I guess we shouldn't receive undefined objects as responses, so this solution makes more senes anyway.

src/utils.ts
```diff
@@ -76,6 +77,29 @@ function handleLegacyTokens(token: string): Token {
 	return { token: [{ proofs: obj.proofs, mint: obj?.mints[0]?.url ?? '' }] };
 }
 
+export function isObj(v: unknown): v is object {
+	return typeof v === 'object';
+}
+
+export function checkResponse(data: { error?: string; detail?: string }) {
```
gandlafbtc, line 84: I haven't tested the behaviour, and maybe i'm wrong, but wouldn't this only throw the `error` , if both `detail` and `error` are present? would it make sense to accumulate them?

gandlafbtc (commented):
Great! only a few comments, mainly for helping me understand

src/utils.ts
BilligsterUser: This will throw if detail or error is present. I don't know of a case where both will present.

src/CashuWallet.ts
BilligsterUser: fst and snd could be undefined one case would be our 'could not verify proofs' bug. But now error handling catches this  case.

src/CashuWallet.ts
BilligsterUser: I agree. I will revert this change it may hurt more than it helps

src/utils.ts
gandlafbtc: I think it shouldn't be an issue for now. if it is, we will find out later 👍

gandlafbtc (approved):

PR #30, opened 2023-03-30
Token v3

gandlafbtc:
switching to new token format

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -89,10 +89,10 @@ class CashuWallet {
 	}
 
 	async receive(encodedToken: string): Promise<Array<Proof>> {
-		const { proofs } = getDecodedProofs(encodedToken);
-		const amount = proofs.reduce((total, curr) => total + curr.amount, 0);
+		const { token } = getDecodedProofs(encodedToken);
+		const amount = token[0].proofs.reduce((total, curr) => total + curr.amount, 0);
 		const { payload, amount1BlindedMessages, amount2BlindedMessages } =
-			await this.createSplitPayload(0, amount, proofs);
+			await this.createSplitPayload(0, amount, token[0].proofs);
 		const { fst, snd } = await this.mint.split(payload);
 		const proofs1 = fst
 			? dhke.constructProofs(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,7 +1,7 @@
 import { utils } from '@noble/secp256k1';
 import { encodeBase64ToJson, encodeJsonToBase64 } from './base64.js';
 import { Proof } from './model/Proof.js';
-import { Token } from './model/types/index.js';
+import { Token, TokenV2 } from './model/types/index.js';
 
 function splitAmount(value: number): Array<number> {
 	const chunks: Array<number> = [];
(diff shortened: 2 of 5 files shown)


src/CashuWallet.ts
```diff
@@ -89,10 +89,10 @@ class CashuWallet {
 	}
 
 	async receive(encodedToken: string): Promise<Array<Proof>> {
-		const { proofs } = getDecodedProofs(encodedToken);
-		const amount = proofs.reduce((total, curr) => total + curr.amount, 0);
+		const { token } = getDecodedProofs(encodedToken);
+		const amount = token[0].proofs.reduce((total, curr) => total + curr.amount, 0);
```
gandlafbtc, line 93: #31

src/CashuWallet.ts
```diff
@@ -89,10 +89,10 @@ class CashuWallet {
 	}
 
 	async receive(encodedToken: string): Promise<Array<Proof>> {
-		const { proofs } = getDecodedProofs(encodedToken);
-		const amount = proofs.reduce((total, curr) => total + curr.amount, 0);
+		const { token } = getDecodedProofs(encodedToken);
```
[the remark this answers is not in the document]
callebtc: Since the following code relies on `token[0]` to exist, it should also check in `getDecodedToken` for its existence.

callebtc (approved):
See comments on naming and array length checks, otherwise LGTM!

gandlafbtc:
> See comments on naming and array length checks, otherwise LGTM!

Thanks! yeah we should check for existence with '?' before reading from the objects

renaming also makes sense, since we anyways break compatibility, so let's change it now, better than later.

src/utils.ts
```diff
@@ -28,31 +29,55 @@ function bigIntStringify<T>(_key: unknown, value: T) {
...
  */
-function getEncodedProofs(
-	proofs: Array<Proof>,
-	mints?: Array<{ url: string; ids: Array<string> }>
-): string {
-	const token = {
-		proofs,
-		mints: mints ?? []
+function getEncodedProofs(proofs: Array<Proof>, mint: string, memo?: string): string {
+	let token = {
```
BilligsterUser, line 38: maybe 
```typescript
const token : Token= {
		token: [{ mint, proofs }]
	};

//add memo if exist
if(memo){token.memo=memo;}

```

src/utils.ts
```diff
@@ -28,31 +29,55 @@ function bigIntStringify<T>(_key: unknown, value: T) {
...
 function getDecodedProofs(token: string): Token {
-	const obj = encodeBase64ToJson<Token | Array<Proof>>(token);
+	if (token.startsWith('cashu')) {
+		return getDecodedV3Token(token);
+	}
+	return handleLegacyTokens(token);
+}
+
+function getDecodedV3Token(token: string): Token {
+	const version = token.slice(5, 5);
```
BilligsterUser, line 59: why have version if we dont use it?

BilligsterUser (commented):
would be nice to have v3 decode test

src/utils.ts
gandlafbtc: agree on this

BilligsterUser (approved):
🎊

PR #39, opened 2023-04-05
CashuMint: remove mintApiRoot and mintPort, use mintUrl only

### LINKED ISSUE #38: Request: remove `mintApiRoot` and `mintPort`, use `mintHost` only
I think the fact that you can use both is a bit confusing and redundant and should be handled in the app level, not in the lib.

BilligsterUser:
closes #38


gandlafbtc (approved):
:+1:

PR #41, opened 2023-04-06
support for multiple mints/tokens in one tokens

### LINKED ISSUE #31: support for multiple mints/tokens in one tokens
As for now, we parse only the first token in any given token. At a later time we might want to upgrade to accept multiple entries. This will probably need a refactoring of the library, because it would mean that the tokens can come from different mints and have to be handled separately

BilligsterUser:
closes #31


src/CashuWallet.ts
```diff
@@ -88,25 +88,44 @@ class CashuWallet {
...
+				}
+			} catch (error) {
+				// I'm not sure if this is the best way to handle errors
+				if (proofs.length) {
+					return proofs;
+				}
+				throw error;
+			}
+		}
+		return proofs;
```
gandlafbtc, line 128: It's a good question. Since we are dealing with multiple entpoints, some stuff might go through and some stuff might not. It is a pretty clean solution, my only issue with it is that, we will get some proofs back, but those that couldn't be redeemed (maybe a mint is down or something) are now in an invalid token-set (because those that were redeemed are now invalid) , so we will get another error if we try to redeem again for the tokens that had problems. 

Maybe we could return the failed tokens too, so we can only check those once more? What do you think?

something like:
```suggestion
			} catch (error) {
				// I'm not sure if this is the best way to handle errors
				console.error(error)
				tokensWithErrors.push(token)
			}
		}
		return {proofs, tokensWithErrors};
```

if we have any tokensWithError, we can handle them outside of the lib

src/utils.ts
```diff
@@ -35,16 +35,7 @@ function bigIntStringify<T>(_key: unknown, value: T) {
...
-	const token: Token = {
-		token: [{ mint, proofs }]
-	};
-
-	//add memo if exist
-	if (memo) {
-		token.memo = memo;
-	}
-
+function getEncodedToken(token: Token): string {
```
gandlafbtc, line 38: damn, i guess this will break clients again

gandlafbtc: but we have to do it. I should have built it better in the last PR. that is my mistake

src/CashuWallet.ts
```diff
@@ -118,14 +121,11 @@ class CashuWallet {
...
-				if (proofs.length) {
-					return proofs;
-				}
-				throw error;
+				console.error(error);
+				tokensWithErrors.push(token);
 			}
 		}
-		return proofs;
+		return { proofs, tokensWithErrors };
```
gandlafbtc, line 128: should we return `tokensWithErrors` as a `Token` object? so that the client doesn't have to construct it himself. We can then only return it if `tokensWithErrors.tokens.length is > 0`  So we can set the `memo` as well in the returned object, and if not, return `undefined`/`null`.

gandlafbtc (commented):
what do you think?

src/CashuWallet.ts
```diff
@@ -87,9 +87,12 @@ class CashuWallet {
 		return this.payLnInvoice(invoice, encodeBase64ToJson(token));
 	}
 
-	async receive(encodedToken: string): Promise<Array<Proof>> {
+	async receive(
+		encodedToken: string
+	): Promise<{ proofs: Proof[]; tokensWithErrors: { mint: string; proofs: Proof[] }[] }> {
```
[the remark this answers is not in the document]
BilligsterUser: [eslint add array-type rule & fix array syntax](https://github.com/cashubtc/cashu-ts/pull/41/commits/a45e2da55c2002d8e3971ff4976f31213ef396a6)

gandlafbtc:
> It's not pretty, but okay so far. Maybe we'll come up with something more elegant later.

perfect is the enemy of the good

PR #42, opened 2023-04-13
fix melt response bug

gandlafbtc:
While testing the new release version with nutstash, i encountered a bug when trying to melt. The issue is, that the preimage might be null sometimes, and if that's the case, we get a 'bad response', eventhough the payment went through.

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -82,8 +82,7 @@ class CashuMint {
 		try {
 			const { data } = await axios.post<MeltResponse>(`${this.mintUrl}/melt`, meltPayload);
 			checkResponse(data);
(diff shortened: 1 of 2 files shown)


gandlafbtc:
@BilligsterUser

BilligsterUser:
do you know why the preimage is null?

gandlafbtc:
i think because some LNbits backends don't provide it.
At least that's what i remember from asking @callebtc

BilligsterUser:
then we should update the type

```typescript
export type MeltResponse = {
	paid: boolean;
	preimage: string|null;
	change?: Array<SerializedBlindedSignature>;
} & ApiError;
```

PR #46, opened 2023-04-17
Check if keys have changed before creating proofs

### LINKED ISSUE #45: BUG: Check if keys have changed before creating proofs
If the mint has changed keys, but the wallet does not recognize it, it might create broken proofs.  

Possible solution: 
* Calculate the current (wallet) mint keyset id from all keys according to [NUT-02](https://github.com/cashubtc/nuts/blob/main/02.md)
* compare with the id inside the received `BlindedSignature`
* If they don't match, we have to load the new keys and perform the unblinding after that

### LINKED ISSUE #52: NUT-02: Keyset ID URL Safety
https://github.com/cashubtc/nuts/issues/12

### LINKED ISSUE #53: `CashuWallet()` wants `keys` and `mint`, `keys` seems redundant
`CashuWallet()` wants `keys` and `mint` as arguments. Isn't`keys` argument not redundant? 

I think it would be nicer to have only one and get the keys, either in the constructor of the `CashuMint` or (maybe better) when they are needed in `CashuWallet` (right after `step3_Alice` afaik).

```typescript
const mint = new CashuMint(MINTURL);
const keys = await mint.getKeys();
const wallet = new CashuWallet(keys, mint);
```

BilligsterUser:
add and test deriveKeysetId

closes #45 
closes #52 
closes #53


test/wallet.test.ts
gandlafbtc, line 1: hm why did this change?

src/CashuWallet.ts
```diff
@@ -63,7 +64,14 @@ class CashuWallet {
...
 			preimage: payData.preimage,
-			change: payData?.change ? dhke.constructProofs(payData.change, rs, secrets, this.keys) : []
+			change: payData?.change
+				? dhke.constructProofs(
+						payData.change,
+						rs,
+						secrets,
+						payData.change.length ? await this.getKeys(payData.change[0].id) : []
+				  )
+				: []
```
gandlafbtc, line 74: i wonder if it would make sense to return something that indicates if a keyset has changed. We're already returning a bunch of params, what's one more, eh?

src/CashuWallet.ts
```diff
@@ -176,7 +184,23 @@ class CashuWallet {
 		const { blindedMessages, secrets, rs } = await this.createRandomBlindedMessages(amount);
 		const payloads = { outputs: blindedMessages };
 		const { promises } = await this.mint.mint(payloads, hash);
-		return dhke.constructProofs(promises, rs, secrets, this.keys);
+		return dhke.constructProofs(
+			promises,
+			rs,
+			secrets,
+			promises.length ? await this.getKeys(promises[0].id) : []
+		);
```
gandlafbtc, line 192: same here

gandlafbtc (commented):
take a look at the comments, I'm not sure about my proposal, but i feel like we have to let the caller know about the updated keyset, so he can reflect the changes in his state. Maybe there is a better way of doing it tho.

test/wallet.test.ts
BilligsterUser: because i use fake keys here
`mockedAxios.get.mockResolvedValueOnce({
			data: {
				1: '02f970b6ee058705c0dddc4313721cffb7efd3d142d96ea8e01d31c2b2ff09f181'
			}
		}) `
	`	
now we are checking the keys/keysetId so i had to update the ids

src/CashuWallet.ts
BilligsterUser: should we return just a bool like keysChanged or the keys?

gandlafbtc: i think returning the keys themselves would make the most sense.

src/utils.ts
```diff
@@ -72,6 +72,12 @@ function handleTokens(token: string): Token {
 	return { token: [{ proofs: obj.proofs, mint: obj?.mints[0]?.url ?? '' }] };
 }
 
+export async function deriveKeysetId(keys: MintKeys) {
+	const pubkeysConcat = Object.values(keys).join('');
+	const hash = await utils.sha256(new TextEncoder().encode(pubkeysConcat));
+	return Buffer.from(hash).toString('base64').slice(0, 12);
+}
+
```
gandlafbtc, line 80: let's expose this function via index.ts

BilligsterUser:
I think it is better to remove the key parameter from the CashuWallet constructor and not to return changed keys handle all key staff inside the lib.
What do you think?

BilligsterUser:
https://github.com/cashubtc/cashu-ts/pull/46/commits/612e1229322aaddd1b2c4c8e696c4983ccbe091f
to fix #52

BilligsterUser:
https://github.com/cashubtc/cashu-ts/pull/46/commits/5d54e65037d0964bd736e9ade4498b560e0a76ec
to fix #53

src/CashuWallet.ts
```diff
@@ -8,27 +8,31 @@ import { Proof } from './model/Proof.js';
...
 	mint: CashuMint;
 
 	/**
 	 *
 	 * @param keys public keys from the mint
 	 * @param mint Cashu mint instance is used to make api calls
 	 */
-	constructor(keys: MintKeys, mint: CashuMint) {
-		this.keys = keys;
+	constructor(mint: CashuMint) {
```
gandlafbtc, line 33: hm keys no longer get set on construction. is this on purpose?

src/CashuWallet.ts
```diff
@@ -56,14 +60,25 @@ class CashuWallet {
 	 * @param proofsToSend the exact amount to send including fees
 	 * @returns
 	 */
-	async payLnInvoice(invoice: string, proofsToSend: Array<Proof>) {
+	async payLnInvoice(
+		invoice: string,
+		proofsToSend: Array<Proof>
+	): Promise<{
```
gandlafbtc, line 66: Can we create a type for this :grin:

src/CashuWallet.ts
```diff
@@ -72,29 +87,48 @@ class CashuWallet {
 		return fee;
 	}
 
-	static getDecodedLnInvoice(invoice: string) {
+	static getDecodedLnInvoice(invoice: string): {
```
gandlafbtc, line 90: lol. i guess we don't really need this function. should we remove it and call `decode` directly?

src/CashuWallet.ts
```diff
@@ -72,29 +87,48 @@ class CashuWallet {
...
+		newKeys?: MintKeys | undefined;
+	}> {
 		return this.payLnInvoice(invoice, encodeBase64ToJson(token));
 	}
 
-	async receive(
-		encodedToken: string
-	): Promise<{ proofs: Array<Proof>; tokensWithErrors: Token | undefined }> {
+	async receive(encodedToken: string): Promise<{
+		proofs: Array<Proof>;
```
gandlafbtc, line 123: would also make sense to make this a type i think

src/CashuWallet.ts
```diff
@@ -128,14 +165,15 @@ class CashuWallet {
 		}
 		return {
 			proofs,
-			tokensWithErrors: tokensWithErrors.length ? { token: tokensWithErrors } : undefined
+			tokensWithErrors: tokensWithErrors.length ? { token: tokensWithErrors } : undefined,
+			newKeys
```
gandlafbtc, line 169: this function has become so weird. If it does receive tokens from multiple mints, it will only return the token from the 'active' mint. I'm not complaining about your implementation, i think it is good, but maybe it warrants the question if the protocol should support tokens from multiple mints inside a single request, or if that is too complicated. I guess it's functionality vs simplicity. 

Would love your thoughts on that @callebtc

src/CashuWallet.ts
BilligsterUser: should we export it in index.ts ?
```typescript
import { CashuMint } from './CashuMint.js';
import { CashuWallet } from './CashuWallet.js';
import { Proof } from './model/Proof.js';
import { getEncodedToken, getDecodedToken, deriveKeysetId } from './utils.js';
import { decode as getDecodedLnInvoice } from '@gandlaf21/bolt11-decode';
export {
	CashuMint,
	CashuWallet,
	Proof,
	getDecodedToken,
	getEncodedToken,
	deriveKeysetId,
	getDecodedLnInvoice
};
```

BilligsterUser:
I really don't like this PR.I think I should restart from scratch after #48 is merged

gandlafbtc:
fair enough.. i think it got a little hacky there.

src/CashuWallet.ts
callebtc: It's a good question. I think receiving from multiple mints should be pretty straightforward and actually one of the reasons why we ended up with V3 tokens.

As you can see [here in NUT-00](https://github.com/cashubtc/nuts/blob/main/00.md#example-json-2), we assume that each entry of `token.token` is from one mint, that means you should be able to loop through entries `token.token` and receive nuts. In pseudo-python:

```python
for t in token.token:
    activateMint(t.mint)
    redeem(t.proofs)
```

Does that make sense?

gandlafbtc: yes this does make sense. @BilligsterUser maybe we have to refactor the receive function to only take proofs and a mint as an argument? and then let the wallets/implementations handle the complexity?

gandlafbtc:
Thanks for the inputs!

yes, that's how it was planned, but the PR got too weird together with trying to handle the multiple tokens stuff. I think we should reduce the complexity in that function, by taking only proofs and a mint as params and then either let wallets implement their own solution on how they want to deal with multiple tokens, or we can also implement a separate function in the lib that can handle this.

PR #50, opened 2023-04-19
update Proof type

### LINKED ISSUE #49: Proofs Class adds friction
Some of the methods on CashuWallet expect or return proofs as Class. While I understand that OOP leads to cleaner code, it makes working (and especially storing proofs more difficult as those instances are not serializable. So in order to store proofs I have to extract data out of the instance and if I want to work with previously stored ones, I have to create new instances from the data.

BilligsterUser:
closes #49


BilligsterUser:
ready to review @gandlafbtc

gandlafbtc (approved):
Perfect :grin:  zero complaints

PR #51, opened 2023-04-20
added accumulateProof fn

Egge7:
Added a utility function that will iterate over an array of proofs and gather them until a certain amount is reached. The function will take a strategy argument, which determines in which order proofs are selected:

- middle: Will alternate between lowest and highest amount
- ascending: Will start with the lowest amount and work its way up
- descending: Will start with the highest amount and work its way down.

The function returns an object with three properties:

- base: An array that contains all proofs but the last one (Proof[)]
- exceeds: The last Proof that was added to the result (which will make the total amount of gathered proofs greater than the required amount (Proof)
- excess: The amount that the total amount of gathered proofs exceeds the required amount. (number)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -95,11 +95,63 @@ export function checkResponseError(err: unknown) {
 		}
 	}
 }
+
+function accumulateProofs(
+	proofs: Proof[],
+	requiredAmount: number,
+	strategy: 'middle' | 'ascending' | 'descending'
+) {
+	const result: Proof[] = [];
+	const temp = proofs.slice();
+	let total = 0;
+	switch (strategy) {
+		case 'middle': {
+			while (temp.length && total < requiredAmount) {
+				const first = temp.shift();
+				total += first!.amount;
+				result.push(first!);
+				if (total >= requiredAmount) {
+					break;
+				}
+				const last = temp.pop();
+				total += last!.amount;
+				result.push(last!);
+			}
+			break;
+		}
+		case 'ascending': {
+			for (let i = 0; i < temp.length; i++) {
+				total += temp[i].amount;
+				result.push(temp[i]);
+				if (total >= requiredAmount) {
+					break;
+				}
+			}
+			break;
+		}
+		case 'descending': {
+			for (let i = 0; i < temp.length; i++) {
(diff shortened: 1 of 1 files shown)


BilligsterUser (commented):
could you pls run 
```bash
npm run lint
```
and 
```bash
npm run format
```

src/utils.ts
```diff
@@ -95,11 +95,63 @@ export function checkResponseError(err: unknown) {
...
 	}
 }
+
+function accumulateProofs(
+	proofs: Proof[],
+	requiredAmount: number,
+	strategy: 'middle' | 'ascending' | 'descending'
+) {
+	const result: Proof[] = [];
+	const temp = proofs.slice().sort((a, b) => a.amount - b.amount);
```
BilligsterUser, line 105: why not 
```typescript
const temp = proofs.slice()
	if (strategy === 'ascending' || strategy === 'middle') {
		temp.sort((a, b) => a.amount - b.amount);
	} else {
		temp.sort((a, b) => b.amount - a.amount);
	}
```

and than
```typescript
case 'descending':
case 'ascending':
			for (let i = 0; i < temp.length; i++) {
				total += temp[i].amount;
				result.push(temp[i]);
				if (total >= requiredAmount) {
					break;
				}
			}
			break;

```

Egge7:
> could you pls run
> 
> ```shell
> npm run lint
> ```

eslint yells at me for using `!`... The reason for non-null assertion in this case is that shift is typed to return a value or undefined if the array is empty. In my case I know that the array will not be empty, because the loop will only run if temp.length is truthy... Unfortunately the TS does not catch this and will complain about `first` possibly being undefined...

gandlafbtc:
We can expose this function on `index.ts`. maybe at some point it will make sense to expose an util object or something, it's becoming more and more helper methods. Starts to get confusing

gandlafbtc:
@BilligsterUser @Egge7 what shall we do with this PR? Has everything been resolved?

BilligsterUser:
I´m still waiting for @Egge7

gandlafbtc:
@Egge21M what you wanna do with this?

PR #55, opened 2023-04-25
add deriveKeysetId function

### LINKED ISSUE #43: returning proofs on receive is now ambiguous
It used to return proofs only from the mint the wallet is attached to, but since we allow tokens from multiple mints, we could receive proofs from many mints. Clients might receive these tokens without knowledge/(storage of the mint keys) which could lead to confusion. I think we should return a `Token` instead of `proofs`

BilligsterUser:
CashuWallet: remove getDecodedLnInvoice, update payLnInvoiceWithToken & receive now returns a token ( #43  )

Index: export decode as getDecodedLnInvoice and export deriveKeysetId

Utils: add deriveKeysetId function

Types: add PaymentPayload, SendResponse, ReceiveResponse & PayLnInvoiceRespons

closes #43

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -3,6 +3,7 @@ import { CashuMint } from './CashuMint.js';
 import * as dhke from './DHKE.js';
 import { BlindedMessage } from './model/BlindedMessage.js';
 import {
+	BlindedMessageData,
 	BlindedTransaction,
 	MintKeys,
 	PayLnInvoiceResponse,
@@ -129,7 +130,7 @@ class CashuWallet {
 			}
 		}
 		return {
-			token: { token: [{ proofs, mint: this.mint.mintUrl }] },
+			proofs: proofs,
 			tokensWithErrors: tokensWithErrors.length ? { token: tokensWithErrors } : undefined
 		};
 	}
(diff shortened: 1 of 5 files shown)


BilligsterUser:
review pls @gandlafbtc

src/CashuWallet.ts
```diff
@@ -128,15 +125,12 @@ class CashuWallet {
 			}
 		}
 		return {
-			proofs,
+			token: { token: [{ proofs, mint: this.mint.mintUrl }] },
```
gandlafbtc, line 128: what is the benefit of doing it this way?

src/CashuWallet.ts
```diff
@@ -206,13 +200,21 @@ class CashuWallet {
...
 		const amount2: number = amount;
 		return { amount1, amount2 };
 	}
 
-	private async createRandomBlindedMessages(amount: number) {
+	private async createRandomBlindedMessages(amount: number): Promise<{
+		blindedMessages: Array<SerializedBlindedMessage>;
+		secrets: Array<Uint8Array>;
+		rs: Array<bigint>;
+		amounts: Array<number>;
```
gandlafbtc, line 216: maybe type worthy? bellow in createBlankOutputs we have the same object with only `amounts` missing

gandlafbtc (commented):
nutty

src/CashuWallet.ts
BilligsterUser: https://github.com/cashubtc/cashu-ts/issues/43 ?

gandlafbtc: yeah but we're still only returning tokens from one mint here, right? It would only make sense if we return all tokens.

but also: https://github.com/cashubtc/cashu-ts/pull/46#issuecomment-1527733894

maybe we should do it differently.

BilligsterUser: OK

gandlafbtc:
LGTM, we can refactor the receive method later. After we do that, let's make a release. Breaking changes...

gandlafbtc:
@BilligsterUser can merge after rebase

gandlafbtc (approved):

BilligsterUser:
done

gandlafbtc (approved):
lgtm

PR #56, opened 2023-04-26
make checkProofsSpent generic and less strict

BilligsterUser:


src/CashuWallet.ts
```diff
@@ -37,7 +37,7 @@ class CashuWallet {
 	 * @param proofs
```
gandlafbtc, line 37:
```suggestion
	 * @param proofs (only the 'secret' field is required)
```

gandlafbtc (approved):
i think i've seen the same formatting changes in another PR, FIY when rebaseing

gandlafbtc (approved):

PR #62, opened 2023-05-01
CashuWallet: update receive function

BilligsterUser:


BilligsterUser:
Review pls @gandlafbtc

src/CashuWallet.ts
```diff
@@ -218,32 +243,32 @@ class CashuWallet {
...
+	private createBlankOutputs(feeReserve: number): BlindedMessageData {
 		const blindedMessages: Array<SerializedBlindedMessage> = [];
 		const secrets: Array<Uint8Array> = [];
 		const rs: Array<bigint> = [];
 		const count = Math.ceil(Math.log2(feeReserve));
 		for (let i = 0; i < count; i++) {
 			const secret = randomBytes(32);
 			secrets.push(secret);
-			const { B_, r } = await dhke.blindMessage(secret);
+			const { B_, r } = dhke.blindMessage(secret);
```
gandlafbtc, line 271: did this change because we use the new curve lib? or how come this is no longer async

src/utils.ts
```diff
@@ -72,14 +72,43 @@ function handleTokens(token: string): Token {
...
+/**
+ * merge proofs from same mint,
+ * removes TokenEntrys with no proofs or no mint field
+ * and sorts proofs by id
+ *
+ * @export
+ * @param {Token} token
+ * @return {*}  {Token}
+ */
+export function cleanToken(token: Token): Token {
```
gandlafbtc, line 92: would be great to have a test for that funtion!

src/CashuWallet.ts
```diff
@@ -177,24 +168,58 @@ class CashuWallet {
...
+				amount2BlindedMessages.secrets,
+				keys
+			);
+			proofs.push(...proofs1, ...proofs2);
+		} catch (error) {
+			console.error(error);
+			proofsWithError.push(...tokenEntry.proofs);
+		}
+		return {
+			proofs,
```
gandlafbtc, line 206: I'm not even sure if we need this. Can there be single proofs that fail in a token, while others succeed?

gandlafbtc (commented):
Amazing PR :fire:

src/CashuWallet.ts
BilligsterUser: Yes

src/CashuWallet.ts
gandlafbtc: oh nevermind.

gandlafbtc: sorry for the confusion, i see now that you need them in the receive function to determine wether the token was faulty. all good!

src/utils.ts
BilligsterUser: done

BilligsterUser:
after this is merged i will work on checking for changed keys before creating proofs

src/utils.ts
BilligsterUser: thank you for reviewing my prs ❤️

PR #63, opened 2023-05-04
export `setupAxios` for end user

ebrakke:
Adding onto #34, not sure if this was missed or intentional, but I need to setup some custom axios config in cashu.me. Saw this function, but it was not exported at the top level.


gandlafbtc (approved):
very good, was probably missed. Thanks!

PR #64, opened 2023-05-04
CashuWallet: Check if keys have changed before creating proofs

BilligsterUser:


src/CashuWallet.ts
```diff
@@ -30,10 +33,21 @@ class CashuWallet {
...
 	}
 
+	get keys(): MintKeys {
+		return this._keys;
+	}
+	set keys(keys: MintKeys) {
+		this._keys = keys;
+		this._keysetId = deriveKeysetId(this._keys);
+	}
+	get keysetId(): string {
```
gandlafbtc, line 48: We could also only store the keys, and derive the keyset from the keys each time we want to read it.

Pros:
We don't need to keep the keyset variable in sync

Cons:
Might be computationally intensive

src/CashuWallet.ts
```diff
@@ -102,30 +119,34 @@ class CashuWallet {
...
 			} catch (error) {
 				console.error(error);
 				tokenEntriesWithError.push(tokenEntry);
 			}
 		}
 		return {
 			token: { token: tokenEntries },
-			tokensWithErrors: tokenEntriesWithError.length ? { token: tokenEntriesWithError } : undefined
+			tokensWithErrors: tokenEntriesWithError.length ? { token: tokenEntriesWithError } : undefined,
+			newKeys
```
gandlafbtc, line 149: shouldn't this be a map/array of MintKeys? Seems like we could get a weird case, where keys in the second token changed, and now we return it as the new keys of token 1

src/CashuWallet.ts
```diff
@@ -266,7 +336,7 @@ class CashuWallet {
 		const blindedMessages: Array<SerializedBlindedMessage> = [];
 		const secrets: Array<Uint8Array> = [];
 		const rs: Array<bigint> = [];
-		const count = Math.ceil(Math.log2(feeReserve));
+		const count = Math.ceil(Math.log2(feeReserve)) || 1;
```
gandlafbtc, line 339: should we back propagate this change to the other implementations? Or are they already aware of this

src/CashuWallet.ts
BilligsterUser: If we return newKeys the currently active keyset changed .Token 1 could use an old keyset and token 2 could use a new keyset we now return the newKeys (used for token2) because newKeys are now the currently used keyset.

src/CashuWallet.ts
BilligsterUser: True. I'm ok with both ways .You decide.

src/CashuWallet.ts
BilligsterUser: not sure about this

```typescript
Math.ceil(Math.log2(1)) => 0
```

can the LN fee be 1?
If it can and we pay 0 sats fee, the we won't get a change back with `Math.ceil(Math.log2(1))` but we a should get 1 sat back or I'm missing something?

src/CashuWallet.ts
BilligsterUser: but they shouldn´t get out of sync as the keys are a private property
 - keys can be set in the constructor - we recalc the keysetId 
 - keys can be set by the setter - we recalc the keysetId 
 
 also hashing is 'slow' / computationally intensive

src/CashuWallet.ts
gandlafbtc: what if the 2nd tokenentry is from a different mint

BilligsterUser: we only return newKeys from the mint the wallet is using 
see https://github.com/cashubtc/cashu-ts/blob/874a372697077c29347fee6e6a704fb8ffea79da/src/CashuWallet.ts#L236

gandlafbtc: I missed that! thanks for clarifying 🫡

src/CashuWallet.ts
gandlafbtc: the change is good👍 i was just wondering how other implementations handle it, and if we should let them know

src/CashuWallet.ts
gandlafbtc: i agree 👍 we just have to be careful within the class to not mess it up 😁

gandlafbtc (approved):
lgtm!

PR #65, opened 2023-05-05
CashuWallet: make keys constructor param optional

### LINKED ISSUE #53: `CashuWallet()` wants `keys` and `mint`, `keys` seems redundant

BilligsterUser:
closes #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,7 @@ npm i @cashu/cashu-ts
 ```typescript
 import { CashuMint, CashuWallet, getEncodedToken } from '@cashu/cashu-ts';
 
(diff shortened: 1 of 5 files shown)


gandlafbtc (approved):
It's starting to look real good!

PR #66, opened 2023-05-07
update Buffer imports

BilligsterUser:
update Buffer imports from
```
import { Buffer } from 'buffer';
```
to
```
import { Buffer } from 'buffer/';
```
to use the Buffer npm pkg and not nodes Buffer

from https://www.npmjs.com/package/buffer

```
To depend on this module explicitly (without browserify), require it like this:

var Buffer = require('buffer/').Buffer  // note: the trailing slash is important!

To require this module explicitly, use require('buffer/') which tells the node.js module lookup algorithm (also used by browserify) to use the npm module named buffer instead of the node.js core module named buffer!
```

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -278,19 +278,14 @@ class CashuWallet {
 		if (this.keysetId === keysetId) {
 			return this.keys;
 		}
-		// let newKeys = this.keysMap.get(keysetId);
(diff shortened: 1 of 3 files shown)


BilligsterUser:
review pls @gandlafbtc

gandlafbtc (approved):
interesting! good catch 

LGTM

PR #69, opened 2023-05-16
make token entry public

gandlafbtc:
I think we can make this function public. It gives some more flexibility to wallets, on how they would like to receive tokens.


BilligsterUser (approved):
OK

PR #72, opened 2023-05-28
Axios: `/melt` without timeout

### LINKED ISSUE #78: Increase timeouts to infinite
The axios requests needs infinite timeouts for the endpoints `/split, /melt, POST /mint` so that in case of a slow mint, things don't go south (for example, mint responds to `POST /mint` with BlindSignatures after the client has already timed out, then refuses to issue tokens again (because they've been issued already).

See #72

callebtc:
Removes `/melt` timeout. LN payments can take a long time and its best to keep the connection to the mint as long as possible so we can receive change after the payment is done. SoonTM we would have a "pending" state that could make this unnecessary.

closes #78

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -49,7 +49,8 @@ class CashuMint {
 	 */
 	public static async requestMint(mintUrl: string, amount: number): Promise<RequestMintResponse> {
 		const { data } = await axios.get<RequestMintResponse>(`${mintUrl}/mint`, {
-			params: { amount }
+			params: { amount },
+			timeout: 0
 		});
 		return data;
 	}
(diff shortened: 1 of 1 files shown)


BilligsterUser:
Hmm forever is a very long time

gandlafbtc:
yeah will that leave the connection open forever? I think we shouldn't do that. Maybe we can increase the limit, and then as @callebtc said, move to a pending-token system where we can poll the mint for 'invoice paid'

callebtc:
Yes! We literally need infinite here because we don't know how long the Lightning payment will take. The wallet should keep the connection open as long as the mint takes to respond. Even if it doesn't respond at all because something went wrong, I think it's better to sacrifice one connection for a session.

gandlafbtc:
we're removing axios

PR #77, opened 2023-06-20
Split custom amounts

gandlafbtc:
When minting, we can select to give custom instructions to the mint as for what token amounts we want to mint.

With the following approach we can pass an `AmountPreference`:

* give a list of tokens we would like to mint ```[{amount: <number>, count: <number>}, ....]```
* `amount` has to be a power of 2 number, and the count represents how many tokens of this instance should be minted
* if the provided ln payment is greater than  the `AmountPreference` the rest of the tokens will be created in the most optimized way (as one-each powers of 2)
* if the provided payment is smaller than the `AmountPreference` it will cut off before overflowing
* if the cut off leaves a gap between payment and `AmountPreference`, the gap will be refilled with the optimal powers of 2s


callebtc:
In Nutshell, I implemented this as an explicit list of each amount requested. In my opinion this could be better suited here as it allows the implementer of the library to add helpers like `(amount, count)`. It would also result in less code with token selection logic etc...

```python
# in case the user wants a specific split, we create a list of amounts
optional_split = None
if split:
    assert amount % split == 0, "split must be divisor or amount"
    assert amount >= split, "split must smaller or equal amount"
    n_splits = amount // split
    optional_split = [split] * n_splits
    logger.debug(f"Requesting split with {n_splits} * {split} sat tokens.")

 r = await wallet.mint(amount, split=optional_split)
```

gandlafbtc:
merged with #89

PR #81, opened 2023-09-02
Remove axios

SDBowen:
Hey folks, took a look at completing the work to remove axios. Thought this a good way to explore the library a bit.

- Adds a wrapper around the native fetch to type response data and handle errors thrown from bad responses.
- `setupAxios` was replaced with `setGlobalRequestOptions`, which will allow custom settings on fetch.   
- Update to the tests to replace the axios mock. I used [nock](https://github.com/nock/nock) here to mock the HTTP requests themselves. Nock does not support Node's fetch yet and requires `node-fetch` at the moment.


callebtc:
I needed to add `application/json` headers to the requests, otherwise the lib wouldn't work with the validator in @cashu/cashu mint

SDBowen:
Ah, that makes since. Axios adds those common headers.

Made a change to leave off the content-type headers when there is no request body.

callebtc (approved):
Tested with https://github.com/cashubtc/cashu.me

gandlafbtc (approved):
Amazing! 

Just reviewed and did some testing on bun/node

src/CashuMint.ts
```diff
@@ -74,27 +69,17 @@ class CashuMint {
...
-			checkResponse(data);
-			if (!isObj(data) || !Array.isArray(data?.promises)) {
-				throw new Error('bad response');
-			}
-			return data;
-		} catch (err) {
-			checkResponseError(err);
-			throw err;
+		const data = await request<{ promises: Array<SerializedBlindedSignature> }>({
+			endpoint: `${mintUrl}/mint?hash=${hash}`,
```
gandlafbtc, line 73: i guess we can move on from payment_hash now, right @callebtc ?

BilligsterUser:
Pls wait before merging this PR .
Will review this PR tomorrow

BilligsterUser:
😥

gandlafbtc:
sorry, should I revert?

BilligsterUser:
Nop😀

PR #83, opened 2023-09-06
Comments and minor improvements for utils.ts

vnnkl:
Added some comments to the function to make them more consistent with the others

The `isObj` function in the `utils.ts` file has been updated to provide more accurate type checking. Instead of simply checking if the value is an object, it now checks if it is an object with a string key and unknown value. This change improves the reliability of type checking in the codebase because in TypeScript, null is considered an object, which can lead to unexpected behavior.

Made checkResponseError more concise


callebtc (approved):

gandlafbtc (changes requested):
Thanks for the PR!

Sorry i got to it so late. 

We removed axios, do you want to adapt the PR? 
I think you can remove the changes in the `checkResponseError` function. No need to change anything there, we want to revamp the error handling anyway

src/utils.ts
```diff
@@ -31,10 +40,21 @@ function splitAmount(value: number, amountPreference?: Array<AmountPreference>):
 	return chunks;
 }
 
+/**
+ * Checks if a number is a power of two.
+ * @param number The number to check
+ * @returns True if the number is a power of two, false otherwise
+ */
 function isPowerOfTwo(number: number) {
```
gandlafbtc, line 48: while we're at it, can you add the return type to this function signature?

src/utils.ts
```diff
@@ -121,10 +162,11 @@ function handleTokens(token: string): Token {
...
+
 /**
- * Returns the keyset id of a set of keys
- * @param keys keys object to derive keyset id from
- * @returns
+ * Derives the keyset id from a set of keys.
+ * @param keys The keys to derive the keyset id from
+ * @returns The derived keyset id
  */
 export function deriveKeysetId(keys: MintKeys) {
```
gandlafbtc, line 171: Can you add the missing return type here?

src/utils.ts
```diff
@@ -166,34 +205,44 @@ export function cleanToken(token: Token): Token {
 		}))
 	};
 }
+
+/**
+ * Sorts an array of proofs by id.
+ * @param proofs The proofs to sort
+ * @returns The sorted proofs
+ */
 export function sortProofsById(proofs: Array<Proof>) {
```
gandlafbtc, line 214: and here

gandlafbtc (commented):
Hi! Can you please add the missing return types? I think then after this we merge it. I'm still waiting on feedback on other PRs, but IDK how long this takes, so let's merge this first

gandlafbtc:
closing due to inactivity

PR #84, opened 2023-09-12
CashuWallet.ts: add comments

callebtc:
Adds some comments to `CashuWallet.ts`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -263,6 +263,12 @@ class CashuWallet {
 		return { returnChange: change, send: proofsToSend };
 	}
 
+	/**
+	 * Request tokens from the mint
+	 * @param amount amount to request
+	 * @param hash hash to use to identify the request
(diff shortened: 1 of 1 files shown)


src/CashuWallet.ts
```diff
@@ -263,6 +263,12 @@ class CashuWallet {
 		return { returnChange: change, send: proofsToSend };
 	}
 
+	/**
+	 * Request tokens from the mint
+	 * @param amount amount to request
```
gandlafbtc, line 268: amount doesn't exist anymore

gandlafbtc (changes requested):

src/CashuWallet.ts
callebtc: It still exists in this file though.

gandlafbtc: correct, please forgive my stupidity

gandlafbtc (approved):

PR #87, opened 2023-09-18
improve: handle trailing and duplicate `/` in mint urls

thesimplekid:
## Description

I've been having an issue where mint urls with a trailing `/` will cause errors in this library as it ends up with duplicate `/` in the mint urls. This adds a `joinUrls` function to remove duplicate `/` when joining url instead of just appending the path to the mint url string, 

The other way I thought to do this is to strip the trailing `/` when creating a new mint. However, since these functions are publicly exposed think its best to handle mint urls that maybe passed in with a trailing `/`.

...

## Changes

Add: Util `joinUrls`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -12,7 +12,7 @@ import {
 	SplitResponse
 } from './model/types/index.js';
 import request from './request.js';
-import { isObj } from './utils.js';
+import { isObj, joinUrls } from './utils.js';
 
 /**
  * Class represents Cashu Mint API. This class contains Lower level functions that are implemented by CashuWallet.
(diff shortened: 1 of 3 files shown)


gandlafbtc (approved):
LGTM, Thanks for the PR!

There is one thing that might be a footgun, which is that the mint URL can act as kind-of an ID. and by allowing this we could end up with 2 "IDs" for 1 mint, if not handled correctly by the client. one with `/` and one without. 

I still think it's better to make this the clients problem, and not the libraries. The lib should accept both :+1:

thesimplekid:
>There is one thing that might be a footgun, which is that the mint URL can act as kind-of an ID. and by allowing this we could end up with 2 "IDs" for 1 mint, if not handled correctly by the client. one with `/` and one without. 

Yes, I had this problem in the cashu-crab rust crate, and handle it similarly where it is up to the client to not duplicate mints.

callebtc (approved):
Nice one

Issue #86, opened 2023-09-17
Error details for failed receives


dipunm:
Currently, the library will make some assumptions, and will hide details of what went wrong when a proof fails to be `split`.

Specifically, I am looking at the following methods:

- https://github.com/cashubtc/cashu-ts/blob/main/src/CashuWallet.ts#L136
- https://github.com/cashubtc/cashu-ts/blob/main/src/CashuWallet.ts#L176
- https://github.com/cashubtc/cashu-ts/blob/main/src/CashuMint.ts#L139

I would like to at the very least be able to identify the difference between a server rejecting a request (token has already been claimed) vs network issues (eg. http timeout or unexpected response)

There are a few ways to achieve this, but with the current design, I am not sure what considerations I am missing. Also with https://github.com/cashubtc/cashu-ts/pull/85 in progress, I think its too early to attempt a PR.

That said, hopefully this can kick off a discussion at the very least.

My initial thoughts here, the least breaking change, is to add a new property to `ReceiveTokenEntryResponse` and `ReceiveResponse` to include the original error(s) that caused the failure(s).

That said, I would also suggest a change in design if possible. Returning an object with both `proofs` or `tokens` _and_ `____WithErrors` implies that you can have  partial success. If this is not the case (as I assume is not the case based on the current implementation), then throwing an error seems like a more straightforward and standard solution. If not throwing an error, then an object that more succinctly identifies as an error (eg. a `Failed....Response`) or just adding a simple flag like `success: true|false` would help to solidify that there is no partial success.

If there is partial success, I think this becomes a pretty difficult thing to manage, so I hope that's not the case honestly, but if so, then we should explore how clients should handle this and what type of design we need to support that.

gandlafbtc:
Thanks for opening the issue! 

Yes, correct the error handling is very bad right now. That is partialy due to the mints only recent upgrade to expressive errors, and partially due to a switch from axios to fetch in the library that this has not been implemented propperly yet. The mints are upgraded now, and we've switched to fetch, so error handling should be next on the list! Feel free to open a PR regarding this, I will handle the merging.
----
The thing is, we can have partial success because there can be tokens from different mints inside a token. But as far as i can tell from the discussions from the TG channel, we will switch to a simpler data structure that only contains one token per `token` (duh). is that correct @callebtc ? 

this would solve our problem

dipunm:
> Proofs inside a token are transactional, if there is a single invalid proof, the token can't be claimed.

🔥

[...]

PR #88, opened 2023-09-19
Refactored `request` error handling

dipunm:
Exception is now identifiable as HttpResponseError providing a status code. 
**Breaking?** We no longer look for error messages on `ok` responses. (HTTP2XX)

# Fixes: Not yet.

## Description
This is a small PR as part of an ongoing effort to address https://github.com/cashubtc/cashu-ts/issues/86
...

## Changes

- Exception is now identifiable as HttpResponseError providing a status code. 
- **Breaking?** We no longer look for error messages on `ok` responses. (HTTP2XX)

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,4 +1,4 @@
-import { checkResponse } from './utils';
+import { HttpResponseError } from './model/Errors';
 
 type RequestArgs = {
 	endpoint: string;
@@ -23,7 +23,7 @@ async function _request({
 	requestBody,
 	headers: requestHeaders,
 	...options
-}: RequestOptions): Promise<Response> {
+}: RequestOptions): Promise<unknown> {
 	const body = requestBody ? JSON.stringify(requestBody) : undefined;
 	const headers = {
 		...{ Accept: 'application/json, text/plain, */*' },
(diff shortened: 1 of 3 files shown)


gandlafbtc:
awesome! looking forward to this

dipunm:
Should I include the second part into this PR?

I was planning on getting this reviewed and merged and then following up with a second PR.

gandlafbtc:
what kind of files would the 2nd pr touch?

dipunm:
The CashuWallet.ts file

It will change the signature of the methods. I've put details in the issue.

gandlafbtc:
okay... i have a PR coming that changes a bunch in `CashuWallet.ts`. If you can wait a couple days i think we'll have an easier time merging

gandlafbtc:
i'll rebase this branch after i merged the branch `deterministic-secrets`

src/request.ts
```diff
@@ -23,7 +23,7 @@ async function _request({
 	requestBody,
 	headers: requestHeaders,
 	...options
-}: RequestOptions): Promise<Response> {
+}: RequestOptions): Promise<unknown> {
```
gandlafbtc, line 26: Is unknown the best option for a type here? Would love your thoughts on that

gandlafbtc (commented):
Finally rebased onto main after merging other commits! 

"Couple days" TM...

I have one comment on the PR! other than that, looks good to me

src/request.ts
dipunm: Here are my thoughts on this:

1. This is a private function (not exported and denoted with an _ to further emphasise this)
2. It is used by the exported utility function `request` which has an unbounded `T`
3. `response.json()` returns `Promise<any>`
4. `any` is typically discouraged, many linters warn about it. `unknown` is the "type safe" alternative, it makes no assumptions about the type but doesn't give you the freedom to make your own assumptions about it either -- which would lead to a runtime error.
5. An alternative type could be `object`, but the response could technically be a number which is valid JSON I think.
6. The method that is given the type `T` will be able to cast with no resistance.

Did you have another type in mind?

dipunm: Casting as unknown also has no runtime cost, it's all thrown away after type checking and compilation.

gandlafbtc: i was thinking maybe something like `Either response or error` but maybe that makes things too complicated.
```
type Only<T, U> = {
    [P in keyof T]: T[P];
} & {
    [P in keyof U]?: never;
};

type Either<T, U> = Only<T, U> | Only<U, T>
```

Not that important, was just curious to pick your brain

gandlafbtc (approved):

PR #91, opened 2023-10-04
Deterministic secrets and restore from seed phrase :fire: 

gandlafbtc:
# new feature: Deterministic secrets and restore from seed phrase :fire: 

## Description

This PR introduces the functionality for creating ecash derived from a seed. This allows for a backup and restore scheme where the wallet owner only needs to store the 12 words seed phrase to restore his ecash.

## Changes

**External changes**
- Add and expose function `generateNewMnemonic()` for creating new mnemonics
- Add optional `mnemonic` param to pass mnemonic to `CashuWallet` constructor
- Add optional `count` param to all functions that create tokens, to pass the current state of the derivation path
- Add restore() function that recreates ecash that has previously been signed by the mint (doesn't check for validity!)

**Internal changes**
- Add paths for deterministic secret and blinding factor if `mnemonic` and `count` are present
- Implementation secret and blinding factor derivation from seed 
- refactored `createblankOutputs` to reduce duplicate code 

There are no breaking changes, deterministic secrets is opt in. By ommiting  `count` or `mnemonic` params, ecash will be generated randomly.

## how it works

presentation slides --> https://det-sec.gandlaf.com/

## Usage

```typescript
import { CashuMint, CashuWallet, generateNewMnemonic } from '@cashu/cashu-ts';

//generate seed phrase
const seed = generateNewMnemonic();

const cashuMint = new CashuMint('https://.........');

// initialize wallet with seed phrase
const wallet = new CashuWallet(cashuMint, mint.keys, seed);

// as the implementer we have to keep track of the current counter of a keyset as an example, 
// we asume we store the counter in keysetID.counter. (The counter needs to be persistent)
// pass counter into function that creates proofs. 
const {proofs, newKeys} = await wallet.receive({{ some token }}, undefined, keysetID.counter) 

// You need to update your wallet internal counter,
// for the keysetID by the number of proofs that were created
keysetID.counter += proofs.length

// restore. can be called multiple times as long as we find new signatures. 
//In this case it would recreate the blinded messages from 100 to 200
const { proofs, newKeys } = await wallet.restore(100, 100);

// check for valid ecash after creating proofs
const {proofs: validProofs} = await wallet.checkProofsSpent(proofs)

```

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -511,8 +511,10 @@ class CashuWallet {
 		keysetId?: string
 	): BlindedMessageData & { amounts: Array<number> } {
 		// if we atempt to create deterministic messages without a _seed, abort.
-		if (count!=undefined && !this._seed) {
-			throw new Error('Cannot create deterministic messages without seed. Instantiate CashuWallet with a mnemonic, or omit count param.')
+		if (count != undefined && !this._seed) {
+			throw new Error(
+				'Cannot create deterministic messages without seed. Instantiate CashuWallet with a mnemonic, or omit count param.'
+			);
 		}
 		const blindedMessages: Array<SerializedBlindedMessage> = [];
 		const secrets: Array<Uint8Array> = [];
@@ -521,8 +523,10 @@ class CashuWallet {
 			let deterministicR = undefined;
 			let secret = undefined;
 			if (this._seed && count != undefined) {
-				secret = deriveSecret(this._seed, keysetId??this.keysetId, count + i);
-				deterministicR = bytesToNumber(deriveBlindingFactor(this._seed, keysetId??this.keysetId, count + i));
+				secret = deriveSecret(this._seed, keysetId ?? this.keysetId, count + i);
+				deterministicR = bytesToNumber(
+					deriveBlindingFactor(this._seed, keysetId ?? this.keysetId, count + i)
+				);
 			} else {
 				secret = randomBytes(32);
 			}
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -553,6 +553,7 @@ class CashuWallet {
 	 */
 	private createBlankOutputs(feeReserve: number, count?: number): BlindedMessageData {
 		let counter = Math.ceil(Math.log2(feeReserve)) || 1;
+		//Prevent counter from being -Infinity
 		if (counter < 0) {
 			counter = 0;
 		}
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -43,20 +43,26 @@ class CashuWallet {
 	/**
 	 * @param keys public keys from the mint
 	 * @param mint Cashu mint instance is used to make api calls
-	 * @param mnemonic mnemonic seed phrase to initial derivation key for this wallets deterministic secrets
+	 * @param mnemonicOrSeed mnemonic phrase or Seed to initial derivation key for this wallets deterministic secrets. When the mnemonic is provided, the seed will be derived from it. 
+	 * This can lead to poor performance, in which case the seed should be directly provided
 	 */
-	constructor(mint: CashuMint, keys?: MintKeys, mnemonic?: string) {
+	constructor(mint: CashuMint, keys?: MintKeys, mnemonicOrSeed?: string | Uint8Array) {
 		this._keys = keys || {};
 		this.mint = mint;
 		if (keys) {
 			this._keysetId = deriveKeysetId(this._keys);
 		}
-		if (mnemonic) {
-			if (!validateMnemonic(mnemonic, wordlist)) {
-				throw new Error('Tried to instantiate with mnemonic, but mnemonic was invalid');
-			}
-			this._seed = deriveSeedFromMnemonic(mnemonic);
+		if (!mnemonicOrSeed) {
+			return
 		}
+		if (mnemonicOrSeed instanceof Uint8Array) {
+			this._seed = mnemonicOrSeed
+			return
+		}
+		if (!validateMnemonic(mnemonicOrSeed, wordlist)) {
+			throw new Error('Tried to instantiate with mnemonic, but mnemonic was invalid');
+		}
+		this._seed = deriveSeedFromMnemonic(mnemonicOrSeed);
 	}
 
 	get keys(): MintKeys {
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -352,13 +352,14 @@ class CashuWallet {
 	 */
 	async restore(
 		count: number,
-		startIndex = 0
+		startIndex = 0,
+		keysetId?: string
 	): Promise<{ proofs: Array<Proof>; newKeys?: MintKeys }> {
 		if (!this._seed) {
 			throw new Error('CashuWallet must be initialized with mnemonic to use restore');
 		}
 		const amounts = Array(count).fill(0);
-		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, startIndex);
+		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, startIndex, keysetId);
 
 		const { outputs, promises } = await this.mint.restore({ outputs: blindedMessages });
 
@@ -504,7 +505,8 @@ class CashuWallet {
 	 */
 	private createBlindedMessages(
 		amounts: Array<number>,
-		count?: number
+		count?: number,
+		keysetId?: string
 	): BlindedMessageData & { amounts: Array<number> } {
 		const blindedMessages: Array<SerializedBlindedMessage> = [];
 		const secrets: Array<Uint8Array> = [];
@@ -513,8 +515,8 @@ class CashuWallet {
 			let deterministicR = undefined;
 			let secret = undefined;
 			if (this._seed && count != undefined) {
-				secret = deriveSecret(this._seed, this.keysetId, count + i);
-				deterministicR = bytesToNumber(deriveBlindingFactor(this._seed, this.keysetId, count + i));
+				secret = deriveSecret(this._seed, keysetId??this.keysetId, count + i);
+				deterministicR = bytesToNumber(deriveBlindingFactor(this._seed, keysetId??this.keysetId, count + i));
 			} else {
 				secret = randomBytes(32);
 			}
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -43,7 +43,7 @@ class CashuWallet {
 	/**
 	 * @param keys public keys from the mint
 	 * @param mint Cashu mint instance is used to make api calls
-	 * @param mnemonicOrSeed mnemonic phrase or Seed to initial derivation key for this wallets deterministic secrets. When the mnemonic is provided, the seed will be derived from it. 
+	 * @param mnemonicOrSeed mnemonic phrase or Seed to initial derivation key for this wallets deterministic secrets. When the mnemonic is provided, the seed will be derived from it.
 	 * This can lead to poor performance, in which case the seed should be directly provided
 	 */
 	constructor(mint: CashuMint, keys?: MintKeys, mnemonicOrSeed?: string | Uint8Array) {
@@ -53,11 +53,11 @@ class CashuWallet {
 			this._keysetId = deriveKeysetId(this._keys);
 		}
 		if (!mnemonicOrSeed) {
-			return
+			return;
 		}
 		if (mnemonicOrSeed instanceof Uint8Array) {
-			this._seed = mnemonicOrSeed
-			return
+			this._seed = mnemonicOrSeed;
+			return;
 		}
 		if (!validateMnemonic(mnemonicOrSeed, wordlist)) {
 			throw new Error('Tried to instantiate with mnemonic, but mnemonic was invalid');
@@ -375,7 +375,7 @@ class CashuWallet {
 
 		const { outputs, promises } = await this.mint.restore({ outputs: blindedMessages });
 
-		// Collect and map the secrets and blinding factors with the blinded messages that were returned from the mint 
+		// Collect and map the secrets and blinding factors with the blinded messages that were returned from the mint
 		const validRs = rs.filter((r, i) => outputs.map((o) => o.B_).includes(blindedMessages[i].B_));
 		const validSecrets = secrets.filter((s, i) =>
 			outputs.map((o) => o.B_).includes(blindedMessages[i].B_)
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -357,8 +357,7 @@ class CashuWallet {
 		if (!this._seed) {
 			throw new Error('CashuWallet must be initialized with mnemonic to use restore');
 		}
-		const numberOfMessages = count - startIndex;
-		const amounts = Array(numberOfMessages).fill(0);
+		const amounts = Array(count).fill(0);
 		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, startIndex);
 
 		const { outputs, promises } = await this.mint.restore({ outputs: blindedMessages });
@@ -540,7 +539,7 @@ class CashuWallet {
 		if (counter < 0) {
 			counter = 0;
 		}
-		const amounts = counter ? Array(counter).fill(0) : [];
+		const amounts = counter ? Array(counter).fill(1) : [];
 		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, count);
 		return { blindedMessages, secrets, rs };
 	}
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -510,6 +510,10 @@ class CashuWallet {
 		count?: number,
 		keysetId?: string
 	): BlindedMessageData & { amounts: Array<number> } {
+		// if we atempt to create deterministic messages without a _seed, abort.
+		if (count!=undefined && !this._seed) {
+			throw new Error('Cannot create deterministic messages without seed. Instantiate CashuWallet with a mnemonic, or omit count param.')
+		}
 		const blindedMessages: Array<SerializedBlindedMessage> = [];
 		const secrets: Array<Uint8Array> = [];
 		const rs: Array<bigint> = [];
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -346,9 +346,9 @@ class CashuWallet {
 
 	/**
 	 * Regenerates
-	 * @param count? set number of blinded messages that should be generated
+	 * @param count set number of blinded messages that should be generated
 	 * @param startIndex optionally set starting point for count (default is 0)
-	 * @returns proofs and newKeys if they have changed
+	 * @returns proofs (and newKeys, if they have changed)
 	 */
 	async restore(
 		count: number,
@@ -358,6 +358,7 @@ class CashuWallet {
 		if (!this._seed) {
 			throw new Error('CashuWallet must be initialized with mnemonic to use restore');
 		}
+		// create blank amounts for unknown restore amounts
 		const amounts = Array(count).fill(0);
 		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, startIndex, keysetId);
 
@@ -484,7 +485,7 @@ class CashuWallet {
 	/**
 	 * Creates blinded messages for a given amount
 	 * @param amount amount to create blinded messages for
-	 * @param preference optional preference for splitting proofs into specific amounts. overrides amount param
+	 * @param amountPreference optional preference for splitting proofs into specific amounts. overrides amount param
 	 * @param count? optionally set count to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
 	 * @returns blinded messages, secrets, rs, and amounts
 	 */
@@ -501,6 +502,7 @@ class CashuWallet {
 	 * Creates blinded messages for a according to @param amounts
 	 * @param amount array of amounts to create blinded messages for
 	 * @param count? optionally set count to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
+	 * @param keyksetId? override the keysetId derived from the current mintKeys with a custom one. This should be a keyset that was fetched from the `/keysets` endpoint
 	 * @returns blinded messages, secrets, rs, and amounts
 	 */
 	private createBlindedMessages(
@@ -526,7 +528,7 @@ class CashuWallet {
 			const blindedMessage = new BlindedMessage(amounts[i], B_);
 			blindedMessages.push(blindedMessage.getSerializedBlindedMessage());
 		}
-		return { amounts, blindedMessages, rs, secrets };
+		return { blindedMessages, secrets, rs, amounts };
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


dipunm:
First note, can we update the example in the description? I don't really know where `keysetID.COUNTER` comes from. I can't see any references.

test/wallet.test.ts
dipunm, line 1: Note to readers: This file is identical, it only has whitespace changes.

src/model/types/index.ts
```diff
@@ -336,6 +336,13 @@ export type GetInfoResponse = {
 	motd?: string;
 	parameter: { peg_out_only: boolean };
 };
+/**
+ * Response from mint at /info endpoint
```
dipunm, line 340: This is from the /restore endpoint.
```suggestion
 * Response from mint at /restore endpoint
```

src/CashuWallet.ts
```diff
@@ -309,6 +344,36 @@ class CashuWallet {
 		};
 	}
 
+	/**
+	 * Regenerates
+	 * @param count? set number of blinded messages that should be generated
```
dipunm, line 349: Why `?`

src/CashuWallet.ts
```diff
@@ -309,6 +344,36 @@ class CashuWallet {
...
+	 * @param startIndex optionally set starting point for count (default is 0)
+	 * @returns proofs and newKeys if they have changed
```
dipunm, line 351: why `if they have changed` ? Seems unconditional.

src/CashuWallet.ts
```diff
@@ -364,22 +429,31 @@ class CashuWallet {
...
-		const keepBlindedMessages = this.createRandomBlindedMessages(totalAmount - amount);
-		const sendBlindedMessages = this.createRandomBlindedMessages(amount, preference);
+		const keepBlindedMessages = this.createRandomBlindedMessages(
+			totalAmount - amount,
+			undefined,
+			count
+		);
+		if (this._seed && count) {
+			count = count + keepBlindedMessages.secrets.length;
+		}
```
dipunm, line 455: We should probably throw an error if `count` is provided, but `_seed` is not.

src/CashuWallet.ts
```diff
@@ -410,47 +484,65 @@ class CashuWallet {
 	/**
 	 * Creates blinded messages for a given amount
 	 * @param amount amount to create blinded messages for
+	 * @param preference optional preference for splitting proofs into specific amounts. overrides amount param
```
dipunm, line 487:
```suggestion
	 * @param amountPreference optional preference for splitting proofs into specific amounts. overrides amount param
```

src/CashuWallet.ts
```diff
@@ -410,47 +484,65 @@ class CashuWallet {
...
-		amountPreference?: Array<AmountPreference>
+		amountPreference?: Array<AmountPreference>,
+		count?: number
+	): BlindedMessageData & { amounts: Array<number> } {
+		const amounts = splitAmount(amount, amountPreference);
+		return this.createBlindedMessages(amounts, count);
+	}
+
+	/**
+	 * Creates blinded messages for a according to @param amounts
```
dipunm, line 501: review message description, `... for a according ...`

src/CashuWallet.ts
```diff
@@ -410,47 +484,65 @@ class CashuWallet {
...
 		for (let i = 0; i < amounts.length; i++) {
-			const secret = randomBytes(32);
+			let deterministicR = undefined;
+			let secret = undefined;
+			if (this._seed && count != undefined) {
+				secret = deriveSecret(this._seed, keysetId??this.keysetId, count + i);
+				deterministicR = bytesToNumber(deriveBlindingFactor(this._seed, keysetId??this.keysetId, count + i));
+			} else {
+				secret = randomBytes(32);
+			}
```
dipunm, line 522: We should throw or terminate early if `count` is provided but `_seed` is not. This would give unexpected results otherwise. We may put error checking higher up, but putting it here is going to catch more edge cases than expecting all callers to do the validation from now into the future.

src/CashuWallet.ts
```diff
@@ -410,47 +484,65 @@ class CashuWallet {
...
 			secrets.push(secret);
-			const { B_, r } = dhke.blindMessage(secret);
+			const { B_, r } = dhke.blindMessage(secret, deterministicR);
 			rs.push(r);
 			const blindedMessage = new BlindedMessage(amounts[i], B_);
 			blindedMessages.push(blindedMessage.getSerializedBlindedMessage());
 		}
-		return { blindedMessages, secrets, rs, amounts };
+		return { amounts, blindedMessages, rs, secrets };
```
dipunm, line 529: Why is this re-arranged? Does it have a meaningful impact on anything or just formatting?

src/secrets.ts
```diff
@@ -0,0 +1,58 @@
...
+import { generateMnemonic, validateMnemonic, mnemonicToSeedSync } from '@scure/bip39';
+import { wordlist } from '@scure/bip39/wordlists/english';
+import { encodeBase64toUint8 } from './base64';
+import { bytesToNumber } from './utils';
+import { hexToNumber } from '@noble/curves/abstract/utils';
+export const generateNewMnemonic = (): string => {
+	const mnemonic = generateMnemonic(wordlist, 128);
+	if (!validateMnemonic(mnemonic, wordlist)) {
+		return generateNewMnemonic();
+	}
```
dipunm, line 11: Why this recursive behaviour?

dipunm: If this really is a weird edge case, then probably better to do this in a while loop?

src/secrets.ts
```diff
@@ -0,0 +1,58 @@
...
+export const deriveSecret = (seed: Uint8Array, keysetId: string, counter: number): Uint8Array => {
+	return derive(seed, keysetId, counter, 0);
+};
+
+export const deriveBlindingFactor = (
+	seed: Uint8Array,
+	keysetId: string,
+	counter: number
+): Uint8Array => {
+	return derive(seed, keysetId, counter, 1);
```
dipunm, line 29: Use a const to make it clear what `1` represents

src/secrets.ts
```diff
@@ -0,0 +1,58 @@
...
+	return mnemonic;
+};
+
+export const deriveSeedFromMnemonic = (mnemonic: string): Uint8Array => {
+	const seed = mnemonicToSeedSync(mnemonic);
+	return seed;
+};
+
+export const deriveSecret = (seed: Uint8Array, keysetId: string, counter: number): Uint8Array => {
+	return derive(seed, keysetId, counter, 0);
```
dipunm, line 21: use a const to make it clear what `0` represents

src/secrets.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const derive = (
+	seed: Uint8Array,
+	keysetId: string,
+	counter: number,
+	secretOrBlinding: 0 | 1
+): Uint8Array => {
+	const hdkey = HDKey.fromMasterSeed(seed);
+	const keysetIdInt = getKeysetIdInt(keysetId)
+	const derivationPath = `m/129372'/0'/${keysetIdInt}'/${counter}'/${secretOrBlinding}`;
```
dipunm, line 40: How are these numbers decided? `129372'`

test/secrets.test.ts
```diff
@@ -0,0 +1,63 @@
...
+	const seed =
+		'dd44ee516b0647e80b488e8dcc56d736a148f15276bef588b37057476d4b2b25780d3688a32b37353d6995997842c0fd8b412475c891c16310471fbc86dcbda8';
+	const seed_uint8_array = Uint8Array.from(Buffer.from(seed, 'hex'));
+	const hdkey = HDKey.fromMasterSeed(seed_uint8_array);
+	const expected_privatekey = '9d32fc57e6fa2942d05ee475d28ba6a56839b8cb8a3f174b05ed0ed9d3a420f6';
+	const derivation_path = "m/129372'/0'/2004500376'/0'/0";
+	const derived = hdkey.derive(derivation_path);
+	test('derive Secret', async () => {
+		expect(derived.privateKey).not.toBeNull();
+		const privateKey = derived.privateKey || new Uint8Array();
```
dipunm, line 60:
```suggestion
		const privateKey = derived.privateKey;
```

dipunm: Not sure why fallback value is needed here.

src/CashuWallet.ts
```diff
@@ -309,6 +344,36 @@ class CashuWallet {
...
 
+	/**
+	 * Regenerates
+	 * @param count? set number of blinded messages that should be generated
+	 * @param startIndex optionally set starting point for count (default is 0)
+	 * @returns proofs and newKeys if they have changed
+	 */
+	async restore(
+		count: number,
+		startIndex = 0,
```
dipunm, line 355: I understand that `startIndex` is optional here, but from a readability aspect, I would use a signature similar to `slice()`: `restore(start, count)` making neither optional.

src/CashuWallet.ts
```diff
@@ -309,6 +344,36 @@ class CashuWallet {
...
+		keysetId?: string
+	): Promise<{ proofs: Array<Proof>; newKeys?: MintKeys }> {
+		if (!this._seed) {
+			throw new Error('CashuWallet must be initialized with mnemonic to use restore');
+		}
+		const amounts = Array(count).fill(0);
+		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, startIndex, keysetId);
```
dipunm, line 362: What is confusing me here are these `amounts`. Aren't these really indexes for deterministic secrets?

dipunm:
I was getting confused with some of the variable names, help me out here:

- `count`: is this really like a `deterministicNoteIndex`? (`dni`?). It represents the index used to generate both blinding factor and secret used for an ecash token.
- `amount`: when restoring, it seems amount 0 means unknown. The server provides the amount in its response.

One thing to note, what happens if there is a conflict? I imagine that the same note can be signed for with different amounts (i.e. `count=1,amount=1` and `count=1,amount=4` can both exist. So if you send a `count=1,amount=0` token to a mint, how should it respond if it finds multiple signatures for that token?

Should mints be guarding against this case themselves?

src/CashuWallet.ts
```diff
@@ -410,47 +484,65 @@ class CashuWallet {
...
+		count?: number
+	): BlindedMessageData & { amounts: Array<number> } {
+		const amounts = splitAmount(amount, amountPreference);
+		return this.createBlindedMessages(amounts, count);
+	}
+
+	/**
+	 * Creates blinded messages for a according to @param amounts
+	 * @param amount array of amounts to create blinded messages for
+	 * @param count? optionally set count to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
```
gandlafbtc, line 503:
```suggestion
	 * @param count? optionally set count to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
	 * @param keyksetId? override the keysetId derived from the current mintKeys with a custom one. This should be a keyset that was fetched from the `/keysets` endpoint
```

src/CashuWallet.ts
gandlafbtc: I agree. If count is set, but not the `_seed`, we should not allow the function to run. I would say we check for that at the very top of the function, and throw an error. I think it is fair to say that if you provide a counter, you expect it to be deterministic. But now, the code will just return random, because oops the `_seed` was not set.

src/secrets.ts
gandlafbtc: looking at this again, i think this is entirely unnecessary, and the `generateMnemonic` function will always return a valid mnemonic.

gandlafbtc, line 11:
```suggestion
export const generateNewMnemonic = (): string => {
	const mnemonic = generateMnemonic(wordlist, 128);
```

src/secrets.ts
gandlafbtc: very good. I was looking at the code and was like wrf is this `0`

src/secrets.ts
gandlafbtc: in the protocol. It's the standard derivation path for cashu. We can add this path to a const to make it more clear

src/CashuWallet.ts
gandlafbtc:
```suggestion
	 * @param count set number of blinded messages that should be generated
```

src/CashuWallet.ts
gandlafbtc: the idea behind this is that a lot of times you might want to check if you have any ecash created at a certain mint with a seedphrase first: `restore(25)` , and then maybe look for more: restore(25,25).

I agree it reads worse, even though it can be a bit more convenient.
I'm down to change it.

src/CashuWallet.ts
gandlafbtc: lol, I have no explanation for this. Maybe my OCD made me order them alphabetically.

gandlafbtc:
```suggestion
		return { blindedMessages, secrets, rs, amounts };
```

gandlafbtc (commented):
Thanks for the review!

src/CashuWallet.ts
gandlafbtc: let's look at this again after we change the function

gandlafbtc:
> I was getting confused with some of the variable names, help me out here:
> 
>     * `count`: is this really like a `deterministicNoteIndex`? (`dni`?). It represents the index used to generate both blinding factor and secret used for an ecash token.
> 
>     * `amount`: when restoring, it seems amount 0 means unknown. The server provides the amount in its response.
> 
> 
> One thing to note, what happens if there is a conflict? I imagine that the same note can be signed for with different amounts (i.e. `count=1,amount=1` and `count=1,amount=4` can both exist. So if you send a `count=1,amount=0` token to a mint, how should it respond if it finds multiple signatures for that token?
> 
> Should mints be guarding against this case themselves?

Great question! and valid points about the naming.

Yes you could call it a `dni` , i feel like count is simpler. Because that's what it is, a counter. The `count`, together with the `keysetID` and the `cashu standard derivation path` (plus the aditional 0/1 switch for secret/blinding factor) will result in an unique derivation. The amounts in this context don't matter, because they get overriden by the mint with the amount that corresponds with the `BlindedMessage` in their DB. We still have to send amounts, because the `BlindedMessage` model requires them. 

We would need to change it on the protocol level to make the amounts optional on `BlindedMessages`. I think we discussed this before in the cashu telegram group, and we decided against it, but I don't remember the reason :)

src/secrets.ts
dipunm: Can we add a comment to the page where this derivation path is defined?

src/CashuWallet.ts
dipunm: Another good alternative is using an options object kind of pattern:

`restore(count: number, { startIndex?: number } = {})`

src/CashuWallet.ts
```diff
@@ -33,18 +35,23 @@ import {
...
 	 */
-	constructor(mint: CashuMint, keys?: MintKeys) {
+	constructor(mint: CashuMint, keys?: MintKeys, mnemonic?: string) {
 		this._keys = keys || {};
 		this.mint = mint;
 		if (keys) {
 			this._keysetId = deriveKeysetId(this._keys);
 		}
+		if (mnemonic) {
+			this._seed = deriveSeedFromMnemonic(mnemonic);
```
callebtc, line 53: What happens if `mnemonic` is not a valid bip39 string?

src/CashuWallet.ts
```diff
@@ -85,17 +92,19 @@ class CashuWallet {
 	 * @param invoice
 	 * @param proofsToSend the exact amount to send including fees
 	 * @param feeReserve? optionally set LN routing fee reserve. If not set, fee reserve will get fetched at mint
+	 * @param count? optionally set count to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
```
callebtc, line 95: To me, "count" sounds like more like "how many secrets would you like to derive". 

"counter" might be more appropriate for "I want to derive the n-th secret"

src/CashuWallet.ts
```diff
@@ -289,17 +320,21 @@ class CashuWallet {
 	/**
 	 * Request tokens from the mint
 	 * @param amount amount to request
-	 * @param hash hash to use to identify the request
+	 * @param hash hash to use to identify the request*
```
callebtc, line 323: Just a minor nit pick: Would it be possible to rename "hash" to "id" without breaking things since we're not using a hash of anything anymore? 

Afaik, in TS the name of the variable has no effect for the caller, right? I mean, implementors don't use `function(hash=<some_value>)` but only `function(<some_value>)`.

src/CashuWallet.ts
callebtc: After reading to the end of the PR, I've noticed that "count" indeed correctly refers to the first option, so I'll take this back! :)

callebtc: It seems like, I must take my taking back bak again.

The term "count" is indeed used in two different ways which leads to the confusion I've mentioned above:

In `restore` it is used in the first meaning (a range).
In `createBlindedMessages` it is used in the second meaning (a number).

src/CashuWallet.ts
```diff
@@ -309,6 +344,37 @@ class CashuWallet {
...
+		// create blank amounts for unknown restore amounts
+		const amounts = Array(count).fill(0);
+		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, start, keysetId);
+
+		const { outputs, promises } = await this.mint.restore({ outputs: blindedMessages });
+
+		const validRs = rs.filter((r, i) => outputs.map((o) => o.B_).includes(blindedMessages[i].B_));
+		const validSecrets = secrets.filter((s, i) =>
+			outputs.map((o) => o.B_).includes(blindedMessages[i].B_)
+		);
```
callebtc, line 370: I would appreciate a comment on what logic these two steps are based on!

src/CashuWallet.ts
```diff
@@ -364,22 +430,31 @@ class CashuWallet {
...
 		const totalAmount = proofsToSend.reduce((total, curr) => total + curr.amount, 0);
-		const keepBlindedMessages = this.createRandomBlindedMessages(totalAmount - amount);
-		const sendBlindedMessages = this.createRandomBlindedMessages(amount, preference);
+		const keepBlindedMessages = this.createRandomBlindedMessages(
+			totalAmount - amount,
+			undefined,
+			count
+		);
+		if (this._seed && count) {
+			count = count + keepBlindedMessages.secrets.length;
```
callebtc, line 455: Here it is a bit more obvious why this should be "counter".

src/CashuWallet.ts
```diff
@@ -435,22 +539,16 @@ class CashuWallet {
...
-			secrets.push(secret);
-			const { B_, r } = dhke.blindMessage(secret);
-			rs.push(r);
-			const blindedMessage = new BlindedMessage(0, B_);
-			blindedMessages.push(blindedMessage.getSerializedBlindedMessage());
+	private createBlankOutputs(feeReserve: number, count?: number): BlindedMessageData {
+		let counter = Math.ceil(Math.log2(feeReserve)) || 1;
+		if (counter < 0) {
+			counter = 0;
 		}
```
callebtc, line 549: What's the idea reasoning this?

src/secrets.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const getKeysetIdInt = (keysetId:string): bigint =>{
+	let keysetIdInt: bigint
+	if (/^[a-fA-F0-9]+$/.test(keysetId)) {
+		keysetIdInt = (hexToNumber(keysetId)) % BigInt(2 ** 31 - 1);
+	}
+	else {
+		//legacy keyset compatibility
+		keysetIdInt = bytesToNumber(encodeBase64toUint8(keysetId)) % BigInt(2 ** 31 - 1);
+	}
```
callebtc, line 61: *chef's kiss*

callebtc (changes requested):
Great work! Only nitpicks I have are about variable naming!

src/CashuWallet.ts
dipunm: Plus 1, I wanted to make this same suggestion... Either index or counter. Either convey the correct sentiment.

dipunm (approved):
Awesome work. Only nitpick is that `count` variable, but consider this my ACK.

prusnak:
What bip32 paths were used for secrets and blinding factors?

callebtc:
> What bip32 paths were used for secrets and blinding factors?

We haven't officially specced this in a NUT yet but we're now using the same path for two implementations. 

image

Source: https://github.com/cashubtc/nutshell/pull/131

minibits-cash:
Performance issue - creating a CashuWallet instance with seed takes 9 - 10 seconds on mobile device.

```
 LOG  2023-12-01T19:45:46.462Z CashuWallet.constructor deriving seed from mnemonic
 LOG  2023-12-01T19:45:55.831Z CashuWallet.constructor deriving seed from mnemonic - complete
```
caused by:

`this._seed = deriveSeedFromMnemonic(mnemonic);`

In case of more then one instances for different mints, method runs mutiple times even seed is the same.

gandlafbtc:
Wow that's pretty long. I guess we should give the option to instantiate from seed instead from mnemonic. Or what do you suggest?

minibits-cash:
> It is measired in dev mode, but still will be half of that in prod. I would suggest to accept seed instead of mnemonic in the wallet constructor and if there is a reverse function in deps cashu-ts already uses, (seed2mnemonic) use it to show the mnemonic on the UI which is a rare usecase. Just a note, to store uint seed in mobile secure storage, it needs to be converted to string and back but this is not a perf issue.

Actually seed2mnemonic makes no sense as the seed is derived using PBKDF function. Sorry for confusion. So perhaps to store both the seed and the derived key in wallet secure store is a way to go.

gandlafbtc:
That means, the wallet needs to be initialized with seed, and not mnemonic, and we need to expose the seed somehow, maybe as a secondary step after creating a mnemonic as a separate function

src/CashuWallet.ts
gandlafbtc: nothing! but we should check for that. good catch!

src/CashuWallet.ts
gandlafbtc: makes sense, renamed to id

src/CashuWallet.ts
gandlafbtc: agree! restore should be count, everything else should be counter.

minibits-cash:
I found out the hopefully final issue causing the duplicate _B messages passed from wallet to the mint causing failure.

In case there are gaps in proof sequence vs the counter  it is not enough to set the initial counter after the restore to the number of recreated proofs (spent or not).

To be safe against gaps, I believe the counter needs to be set to upper limit of the used recovery interval (e.g. 50 for 1...50) and then further incremented by the number of recovered (only unspent/pending) proofs.

If ack, this should be explicitly named / mentioned in the docs as not following it breaks further wallet operations. 

There is related fix on the nutshell mint side pending (to prevent such duplicate messages account for as spent by mint)

gandlafbtc:
@callebtc there are still requested changes pending from you. Can you take a look and check if it's ok now? And then let's merge this badboy!

a1denvalu3 (approved):
The vulnerability reported in this PR has been fixed in the latest commits of the target PR branch.

Issue #94, opened 2023-11-18
Add utility function to decode LN-invoice 


gandlafbtc:
especially amount would be nice. Now it has to be done manually outside of the lib:

```javascript
const amount = decode(invoice).sections[2].value / 1000;
```

Egge21M:
I have written this exact utility a million times and would be happy to add it here. 

But should this really be a utility function, or should the wallet API change in a future major update to return an invoice object that has pr, amount, memo etc?

PR #98, opened 2024-01-17
added decodeInvoice util

Egge21M:
# Fixes: #94 

## Description
Added a decodeInvoice utility function that takes in a bolt11 invoice and returns an object of type InvoiceData
```ts
type InvoiceData = {
	paymentRequest: string;
	amountInSats: number;
	amountInMSats: number;
	timestamp: number;
	paymentHash: string;
	memo: string;
	expiry: number;
};
```

## Changes

- Added type InvoiceData
- Added decodeInvoice to utils.ts
- Added test

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -3,7 +3,7 @@
 
 name: Node.js CI
 
-on: 
+on:
   push:
     branches:
       - main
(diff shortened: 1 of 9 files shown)


gandlafbtc (commented):
nice! We could add an import and export for this function in `index.ts` so it is easily accessible when importing the lib

gandlafbtc (approved):

Issue #90, opened 2023-09-21
Order outputs by amount and not in buckets to keep and send


callebtc:
For privacy reasons, the outputs should be ordered in a deterministic fashion so that the mint can not deduce what amount we are trying to send or keep. 

See the note in NUT-06:

> Note: In order to preserve privacy around the amount that a client might want to send to another user and keep the rest as change, the client **SHOULD** ensure that the list of `BlindedMessage`s is ordered by amount in ascending order. As an example of what to avoid, a request for tokens expressed like so: `[16, 8, 2, 64, 8]` might imply the client is building a payment for 26 sat; the client should instead order the list like so: `[2, 8, 8, 16, 64]` to mitigate this privacy leak to the mint.

The following piece of code in `CashuWallet.ts` and especially `this.createSplitPayload` should return an Array of `blindedMessages` that is sorted by amount and a boolean vector of which ones to keep and which ones to send so that we don't have to do the sum at the end of the block but simply filter the returned proofs by this vector.

```typescript
if (amount < amountAvailable || preference) {
	const { amountKeep, amountSend } = this.splitReceive(amount, amountAvailable);
	const { payload, blindedMessages } = this.createSplitPayload(amountSend, proofsToSend, preference);
	const { promises } = await this.mint.split(payload);
	const proofs = dhke.constructProofs(
		promises,
		blindedMessages.rs,
		blindedMessages.secrets,
		await this.getKeys(promises)
	);
	// sum up proofs until amount2 is reached
	const splitProofsToKeep: Array<Proof> = [];
	const splitProofsToSend: Array<Proof> = [];
	let amountSendCounter = 0;
	proofs.forEach((proof) => {
		if (amountSendCounter >= amountSend) {
			splitProofsToKeep.push(proof);
			return;
		}
		amountSendCounter = amountSendCounter + proof.amount;
		splitProofsToSend.push(proof);
	});
	return {
		returnChange: [...splitProofsToKeep, ...proofsToKeep],
		send: splitProofsToSend,
		newKeys: await this.changedKeys([...(promises || [])])
	};
}
```

Egge21M:
@callebtc started implementing this in https://github.com/cashubtc/cashu-ts/pull/100

Is this what you had in mind?

Egge21M:
This has been implemented together with the big blinding data refactor

PR #100, opened 2024-01-18
Sorting blindedMessages by amount

Egge21M:
# Fixes: #90

## Description
Ordered blindedMessages returned by createSplitPayload by amount and added a keepVector

## Changes

- Ordered blindedMessages by amount
- added keepVector as return value

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -299,15 +299,15 @@ class CashuWallet {
 			throw new Error('Not enough funds available');
 		}
 		if (amount < amountAvailable || preference) {
-			const { amountKeep, amountSend } = this.splitReceive(amount, amountAvailable);
-			const { payload, blindedMessages } = this.createSplitPayload(
+			const { amountSend } = this.splitReceive(amount, amountAvailable);
+			const { payload, blindedMessages, keepVector } = this.createSplitPayload(
 				amountSend,
 				proofsToSend,
 				preference,
 				counter
 			);
 			const { promises } = await this.mint.split(payload);
-			const proofs = dhke.constructProofs(
+			const splitProofs = dhke.constructProofs(
 				promises,
 				blindedMessages.rs,
 				blindedMessages.secrets,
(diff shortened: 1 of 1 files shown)


Egge21M:
This is now ready for review. As outline above I changed createSplitPayload to return a payload containing a list of proofs ordered by amount, as well as a list of booleans indicating which proof to keep and which to send.
I then changed send to use the list of booleans (keepVector) to /split and keep the right proofs.

@gandlafbtc @callebtc

src/CashuWallet.ts
```diff
@@ -474,22 +473,43 @@ class CashuWallet {
...
+
+		const convertedSendMessages = sendBlindedMessages.blindedMessages.map((_, i) => ({
+			msg: sendBlindedMessages.blindedMessages[i],
+			secret: sendBlindedMessages.secrets[i],
+			rs: sendBlindedMessages.rs[i],
+			amount: sendBlindedMessages.amounts[i],
+			keep: false
+		}));
+		const keepVector: Array<boolean> = [];
 		const blindedMessages: BlindedTransaction = {
```
gandlafbtc, line 492: let's rename this variable to blindedTransaction. `blindedMessages` seems kinda confusing. What do you think?

src/CashuWallet.ts
```diff
@@ -474,22 +473,43 @@ class CashuWallet {
...
-				...keepBlindedMessages.blindedMessages,
-				...sendBlindedMessages.blindedMessages
-			],
-			secrets: [...keepBlindedMessages.secrets, ...sendBlindedMessages.secrets],
-			rs: [...keepBlindedMessages.rs, ...sendBlindedMessages.rs],
-			amounts: [...keepBlindedMessages.amounts, ...sendBlindedMessages.amounts]
+			blindedMessages: [],
+			secrets: [],
+			rs: [],
+			amounts: []
```
gandlafbtc, line 496: we could add the keep vector here instead

gandlafbtc (commented):
Looks good! Only thing missing is a Unit test testing the behaviour.

src/CashuWallet.ts
Egge21M: Oh yes absolutely. Will do

src/CashuWallet.ts
Egge21M: True, I was not sure how many lines of code depend on the type  `BlindedTransaction` and wanted to reduce this PRs footprint. But I will check that out in my next iteration.

Egge21M:
> Looks good! Only thing missing is a Unit test testing the behaviour.

I was under the assumption that the old units tests for this method would still suffice, but we should probably add one to check wether the amount is actually sorted (will do asap). Or am I missing something?

Egge21M:
> Looks good! Only thing missing is a Unit test testing the behaviour.

What would be good way to easily generate secrets and signatures for the Unit tests?

gandlafbtc:
@Egge21M I think this PR might be outdated, especially after https://github.com/cashubtc/cashu-ts/pull/159.. shall we close it?

PR #101, opened 2024-01-24
add missing paymentRequest, update InvoiceData type

KKA11010:
## Description

After updating to 0.9.0 and testing out the util function "decodeInvoice", I noticed that the property "paymentRequest" is undefined. I assume that it should hold the invoice string.

Since the variable "invoiceData" is initialized as empty object, I've also updated the type "InvoiceData" because all properties (except "paymentRequest") could be undefined for whatever reason.

Please let me know if I am missing something @Egge21M @BilligsterUser @gandlafbtc

### DIFF
--- a/src/model/types/index.ts
+++ b/src/model/types/index.ts
@@ -351,10 +351,10 @@ export type AmountPreference = {
 
 export type InvoiceData = {
 	paymentRequest: string;
-	amountInSats: number;
-	amountInMSats: number;
-	timestamp: number;
-	paymentHash: string;
-	memo: string;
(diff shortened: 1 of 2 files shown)


Egge21M:
Good catch. I forgot about paymentRequest...

Making all properties optional also makes sense, as the underlying library does not actually validate an invoice. So while all of these should be present in a valid invoice, there is nothing inside decode that will throw if they are not.

However this means that consumers will need to guard to make the compiler happy.

gandlafbtc (approved):
looks good!

PR #105, opened 2024-02-19
switch to hex secrets

gandlafbtc:
:warning: This PR will affect the restore process of wallets that implement determinsitic secrets. Wallets that implement this PR will have to initiate a 'self-spend' of all ecash after upgrading, in order to ensure the ecash will be recoverable.  :warning: 

# Fixes: base64 secret format 

## Description
The secrets that cashu-ts uses is a base64 string, which makes it incompatible for restoring ecash accross clients that use hex secrets. The spec recommends to use a hex string, so that's why we switch.

## Changes

- change from b64 to hex
- fix tests

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -1,4 +1,4 @@
-import { randomBytes } from '@noble/hashes/utils';
+import { bytesToHex, hexToBytes, randomBytes } from '@noble/hashes/utils';
 import { CashuMint } from './CashuMint.js';
 import * as dhke from './DHKE.js';
 import { BlindedMessage } from './model/BlindedMessage.js';
@@ -548,6 +548,8 @@ class CashuWallet {
 			} else {
 				secret = randomBytes(32);
 			}
+			secret = bytesToHex(secret);
+			secret = new TextEncoder().encode(secret);
 			secrets.push(secret);
 			const { B_, r } = dhke.blindMessage(secret, deterministicR);
 			rs.push(r);
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -539,17 +539,17 @@ class CashuWallet {
 		const rs: Array<bigint> = [];
 		for (let i = 0; i < amounts.length; i++) {
 			let deterministicR = undefined;
-			let secret = undefined;
+			let secretBytes = undefined;
 			if (this._seed && counter != undefined) {
-				secret = deriveSecret(this._seed, keysetId ?? this.keysetId, counter + i);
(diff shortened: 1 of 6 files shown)


src/CashuWallet.ts
```diff
@@ -548,6 +548,8 @@ class CashuWallet {
 			} else {
 				secret = randomBytes(32);
 			}
+			secret = bytesToHex(secret);
+			secret = new TextEncoder().encode(secret);
```
Egge21M, line 552: This might be a breaking change for NodeJS implementations, as TextEncoder is not part of the global scope iirc.

Egge21M: PS: Why are we encoding bytes to hex and then reading hex string as utf8 to bytes again?

callebtc: I would suggest to use different variables with fixed types like `secret_bytes` and `secret_str`.

callebtc (changes requested):
LGTM, with a suggestion to avoid redefining the `secret` variable and changing its type

src/CashuWallet.ts
gandlafbtc: > This might be a breaking change for NodeJS implementations, as TextEncoder is not part of the global scope iirc.
We're already using `TextEncoder`
https://github.com/cashubtc/cashu-ts/blob/bc6724aa0fae2454a338d22b85dd04c21b9ba6d3/src/utils.ts#L136

I wonder if NodeJS breaks, I haven't tested in a while. Also, isn't jest runing in the nodeJS context? It should break there also, no?
@Egge21M

gandlafbtc: > What am I missing here?
1. we convert the bytes to a hex string
2. we encode the hex string to bytes

It's not the same bytes that are coming out as those that wen't in. It's a bit confusing, but thats mainly due to it being JS

Egge21M: > 1. we convert the bytes to a hex string
> 
> 2. we encode the hex string to bytes
> 


> It's not the same bytes that are coming out as those that wen't in. It's a bit confusing, but thats mainly due to it being JS

It's not about JS. You encode an Arraybuffer into hex (e.g. 32 bytes = 64 hex characters) and then read that as utf-8 (e.g. 64 characters = 64 bytes)

gandlafbtc: yes, this is according to spec: https://github.com/cashubtc/nuts/blob/main/00.md#proof

callebtc (approved):

Egge21M (approved):
LGTM :)

gandlafbtc:
I'll merge it, we'll probably have to fix a test though.

PR #106, opened 2024-02-19
Marge & add unit field to CashuWallet instance

starbackr-dev:
- Merged and resolved code from main branch
 - Added unit field to pass on 'sat' or 'usd'. It defaults to 'sat'
 - Added integration test case to get active quote from mint.current.fyi

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -36,21 +36,28 @@ import { bytesToHex } from '@noble/curves/abstract/utils';
 class CashuWallet {
 	private _keys = {} as MintKeys;
 	private _keysetId = '';
+	private _unit='sat';
 
 	mint: CashuMint;
-	unit = 'sat';
 
 	/**
 	 * @param keys public keys from the mint
 	 * @param mint Cashu mint instance is used to make api calls
 	 */
-	constructor(mint: CashuMint, keys?: MintKeys) {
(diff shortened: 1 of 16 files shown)


src/CashuWallet.ts
```diff
@@ -503,75 +420,50 @@ class CashuWallet {
...
 	 * @param feeReserve amount to cover with blank outputs
 	 * @param counter? optionally set counter to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
 	 * @returns blinded messages, secrets, and rs
 	 */
-	private createBlankOutputs(feeReserve: number, counter?: number): BlindedMessageData {
-		let count = Math.ceil(Math.log2(feeReserve)) || 1;
-		//Prevent count from being -Infinity
-		if (count < 0) {
-			count = 0;
+	private createBlankOutputs(feeReserve: number): BlindedMessageData {
```
gandlafbtc, line 453: where did the counters go?

we need them for determinstic secrets

gandlafbtc (changes requested):
Thanks for the PR!

I think you might have mistakenly removed the counters. They are very important though, since we need them for creating non-random/deterministic secrets

gandlafbtc:
gonna close this since it will get merged here #110

PR #115, opened 2024-03-25
Major Version 1.0.0

gandlafbtc:
# New version 1.0.0

- [x] Open PR
- [x] run `npm run test-integration` --> no failing integration tests
- [x] run `npm run format`
- [x] update readme
- [x] create/publish release candidate
- [x] create migration guide
- [ ] get feedback from wallet devs

------ After Merge

- [ ] Rebase branches
- [ ] Prepare next release items


minibits-cash:
As I've reviewed CashuWallet I noticed that although the CashuMint provides the way to check for different states of proofs with the mint, CashuWallet api still provides method to check only for SPENT status.

As the check api is by far the most called api by the wallets, retrieving PENDING status should not involve making another check api call. Another reason is, that PENDING by mint status is quite rare, so such overhead would be really wasteful.

I suggest to adapt CashuWallet.checkSpent method to more generic CashuWallet.check, that would follow CashuMint method signature, retrieving various statuses in the response.

gandlafbtc:
> As I've reviewed CashuWallet I noticed that although the CashuMint provides the way to check for different states of proofs with the mint, CashuWallet api still provides method to check only for SPENT state.
> 
> As the check api is by far the most called api by the wallets, retrieving PENDING state should not involve making another check api call. Another reason is, that PENDING by mint state is quite rare, so such overhead would be really wasteful.
> 
> I suggest to adapt CashuWallet.checkSpent method to more generic CashuWallet.check, that would follow CashuMint method signature, retrieving various states in the response.

That's a good suggestion, and we should definitely add it in the 1.0.0 release. Would you be interested in PR-ing it?

gandlafbtc:
@minibits-cash Sounds right to me. I think we need to get the unit from the to-be-received `TokenEntry`s keyset, in case the wallet was initialized with a different unit. Good catch! There might be more room for improvement of the receive flow, If you have any good ideas, feel free to PR them!

callebtc (approved):
Ready to merge?

gandlafbtc:
i think we should add the changes @minibits-cash suggested, before we merge and release as v1 . There is already a prerelease, RC3 that can be used to test the new version

Egge21M:
Seems like one of the later RCs broke browser support. I suspect this happened when crypto related stuff got moved to the crypto lib.

`ReferenceError: Buffer is not defined`

Egge21M:
> Seems like one of the later RCs broke browser support. I suspect this happened when crypto related stuff got moved to the crypto lib.
> 
> `ReferenceError: Buffer is not defined`

I was just able to confirm this. Importing `buffer` in a consuming project fixes this. The crypto lib should handle this on its own without polyfilling a global IMHO

callebtc:
> The crypto lib should handle this on its own without polyfilling a global IMHO

Done by you, here: https://github.com/cashubtc/cashu-crypto-ts/releases/tag/v0.2.7

Egge21M:
I did a few manual browser tests today and it seems that polyfilling the crypto lib fixed the issues. I think we can finally release :)

Egge21M (approved):

gandlafbtc:
Let's go! Thanks @callebtc and @Egge21M  for doing the heavy lifting

PR #123, opened 2024-04-19
Bug/send keyset + make meltQuote optional in payLnInvoice

gudnuf:
# Fixes: 

- calling `CashuWallet.send()`  with a keyset for units other than 'sat'
-  `payLnInvoice` required a `meltQuote` even though it creates one.

## Description
Very small changes. I just added `keysetId` to the `send` function and passed it to `getKeys`.

I also made it so that in `getKeys`, if the keysetId parameter is`undefined` it will not be compared to `this._keys.id`

...

## Changes

- I add
- ...

## PR Tasks

NOTE: `npm run test` failed all tests even before I made changes:

image

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -449,7 +449,7 @@ class CashuWallet {
 	async payLnInvoice(
 		invoice: string,
 		proofsToSend: Array<Proof>,
-		meltQuote: MeltQuoteResponse,
+		meltQuote?: MeltQuoteResponse,
 		options?: {
 			keysetId?: string;
 			counter?: number;
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -222,12 +222,13 @@ class CashuWallet {
 			counter?: number;
 			pubkey?: string;
 			privkey?: string;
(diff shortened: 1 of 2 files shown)


callebtc:
I encountered a similar problem with `receiveTokenEntry()` where `getKeys()` gets the wrong keyset for a unit other than `sat`. What do you think of adding additional support to pass the `keyset` (not just a `keysetId`), in which case `getKeys()` could also be skipped? That would get rid of one round trip if the implementor already knows the keyset.

gandlafbtc (approved):
LGTM, thanks!

gandlafbtc:
> I encountered a similar problem with `receiveTokenEntry()` where `getKeys()` gets the wrong keyset for a unit other than `sat`. What do you think of adding additional support to pass the `keyset` (not just a `keysetId`), in which case `getKeys()` could also be skipped? That would get rid of one round trip if the implementor already knows the keyset.

We could do that, but wouldn't it make more sense to just have another instance of cashuWallet instantiated with different keys?

Issue #124, opened 2024-04-23
Connect to mint fails because of trailing slash


Dayvvo:
When the mint URL supplied to the CashuMint class contains a trailing URL. the connect request would fail as shown here:
<img width="1680" alt="image" src="https://github.com/cashubtc/cashu-ts/assets/45407706/4062cfe9-2730-4d74-b27d-5d5496da46f5">. This could cause some confusion should for a regular user without technical knowhow

Solution: the CashuMint Class should handle trailing URLS by removing the trailing slash in the class if it exists

Dayvvo:
@callebtc please let me know your thoughts towards this. Looking forward to your response!

callebtc:
Agreed! Please feel free to open an PR!

Dayvvo:
Alright thanks! on it

PR #125, opened 2024-04-25
Fix: Handle trailing slashes in mint urls supplied to the CashuMint class

Dayvvo:
# Fixes: #124 

## Description
Mint URLS containing a trailing slash at the end fail to connect because of that slash. this fix includes a sanitizeurl function that removes the trailing slash when it it exists
...

## Changes

- ...sanitizeURL function in /src/utils.ts for handling trailing slash
- ...utilize sanitizeUrl function in /src/CashuMint.ts in the `get MintUrl` method in the CashuMint class

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -19,7 +19,7 @@ import type {
 	MeltQuoteResponse
 } from './model/types/index.js';
 import request from './request.js';
-import { isObj, joinUrls } from './utils.js';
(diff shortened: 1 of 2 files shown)


Dayvvo:
@callebtc worked on this. Please check this out and provide some feedback

gandlafbtc:
thanks for the PR @Dayvvo , i took the liberty to change the url sanitizing to the constructor, so we don't need to call it every time we access the getter.

PR #126, opened 2024-04-27
Bug: fix `keyset` or `keysetId` for `receiveTokenEntry()`

callebtc:
Bug related to #123. 

`getKeys()` used to fetch `sat` unit keysets per default which leads to errors when receiving a different unit.

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -163,6 +163,8 @@ class CashuWallet {
 			counter?: number;
 			pubkey?: string;
 			privkey?: string;
+			keysetId?: string;
+			keyset?: MintKeys;
 		}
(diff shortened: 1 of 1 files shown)


gandlafbtc:
should we add optional keyset to other functions too? Usually it was thought that the keys get initialized with the wallet, so make a different instance for each keyset. But I am interested to know how other people are using it

callebtc:
> should we add optional keyset to other functions too? Usually it was thought that the keys get initialized with the wallet, so make a different instance for each keyset. But I am interested to know how other people are using it

We could do both: take those the wallet got initialized with or an optional defined keyset and / or unit. Ideally this would be coherent across all relevant methods. I could do that, would you like me to?

gudnuf:
seems to be resolved in latest version fyi

PR #128, opened 2024-05-02
Fix: rename `amountPreference` to `preference`

callebtc:
Option in `mintTokens` was called `amountPreference` but it's `preference` everywhere else.


gandlafbtc (approved):
nice catch... implementors will have to adapt to this change

PR #131, opened 2024-05-17
Add `GET quote` methods and rename methods

callebtc:
This PR adds the `GET quote` methods from the v1 api and the `expiry` and `paid` attributes to the models that are defined in NUT-4 and 5. 

**Attention! This PR changes method names!!!!** 

### Rename `getMintQuote` and `getMintQuote` to `mintQuote` and `meltQuote`

`getMintQuote` is **REPLACED** as this method creates a new quote and the old naming is confusing with relation to the API.

```ts
getMintQuote(amount: number); -> mintQuote(amount: number)
getmeltQuote(request: string) -> meltQuote(request: string)
```

### Add new 

New GET methods
```ts
getMintQuote(quoteId: string)
getMeltQuote(quoteId: string)
```

### Rename Split to Swap

```ts
SplitPayload -> SwapPayload
SplitResponse -> SwapResponse
```

# Upgrade notes

For minting new tokens, most wallets used this flow: request mint quote, and then keep calling the `mint` function until it doesn't return an error anymore (but returns ecash). This is not very elegant since we're using an error case to control the flow of the application. 

It's better to request a mint (with `mintQuote()`), then keep checking if the mint quote is `paid === true` by calling `getMintQuote(quoteId: string)`. If the quote is `paid`, then call `mint()`. This should not entail any errors and can be controlled more precisely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Implemented [NUTs](https://github.com/cashubtc/nuts/):
 - [x] [NUT-07](https://github.com/cashubtc/nuts/blob/main/07.md)
 - [x] [NUT-08](https://github.com/cashubtc/nuts/blob/main/08.md)
 - [x] [NUT-09](https://github.com/cashubtc/nuts/blob/main/09.md)
+- [x] [NUT-11](https://github.com/cashubtc/nuts/blob/main/11.md)
 
 Supported token formats:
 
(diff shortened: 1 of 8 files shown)


gandlafbtc (approved):
LGTM

PR #134, opened 2024-05-20
Use cashu crypto library functions, remove dhke and secrets module

capitalist42:
## Description
All functions related to BDHKE and restoring signatures (from secrets) were copied to the "@cashu/crypto" library. 
This pull request aims to refactor code to use functions from the library.
...

## Changes

- remove 'dhke' module and related test cases 
- remove 'secrets' module and related test cases 
- replace functions with "@cashu/crypto" 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


src/CashuWallet.ts
```diff
@@ -682,6 +690,31 @@ class CashuWallet {
 		const { blindedMessages, rs, secrets } = this.createBlindedMessages(amounts, keysetId, counter);
 		return { blindedMessages, secrets, rs };
 	}
+
```
gandlafbtc, line 693: It would probably make sense to move this to the cashu-crypto lib too. We can do it in a later PR though

gandlafbtc (approved):
Nice Work!

I added a comment on the `constructProofs` function, but we can do it later. We can merge this as is :partying_face:

Issue #129, opened 2024-05-02
A set of amount `preference` fixes


callebtc:
Some issues I think are worth fixing:

- Allow setting the preference for all outputs of a swap (`createSplitPayload`), not only for `proofsToSend` (but also `proofsToKeep`)
- `splitAmount` seems to "fill up the rest" with 2^n even if `amountPreference` is set. Is this intended?
- Do not check if all amounts are `2^n`, since this is not generally the case. Replace this check by checking if every amount has a key.
- Consider replacing the `AmountPrefernce` object `{ amount: int, count: int }` with a simple `number[]` that holds the requested amounts (feels cleaner to me).

lollerfirst:
My new PR addresses first 3 points. I take it that point 4 is to be discussed.

Egge21M:
@callebtc this is done, right?

gandlafbtc:
will be closed with #159

PR #140, opened 2024-06-18
Amount Preferences Fixes

lollerfirst:
# Fixes: #129 

## Changes

- getPreferences checks input amouts against the amounts of the keys
- keepPreferences and sendPreferences as different optional parameters in createSplitPayload.
- modify splitAmount behaviour -in case of no specified preference- to create amounts not from powers of 2
    but from the amounts of the keyset instead.
- splitAmount reorders its outputs by descending/ascending order.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -21,7 +21,8 @@ import {
 	CheckStateEnum,
 	SerializedBlindedSignature,
 	MeltQuoteState,
-	MintQuoteResponse
+	MintQuoteResponse,
+	Preferences
 } from './model/types/index.js';
 import {
 	bytesToNumber,
@@ -171,11 +172,12 @@ class CashuWallet {
 		if (!preference) {
 			preference = getDefaultAmountPreference(amount, keys);
 		}
+		const pref: Preferences = {sendPreference: preference};
 		const { payload, blindedMessages } = this.createSwapPayload(
 			amount,
 			tokenEntry.proofs,
 			keys,
-			preference,
+			pref,
 			options?.counter,
 			options?.pubkey,
 			options?.privkey
@@ -207,7 +209,7 @@ class CashuWallet {
 		amount: number,
 		proofs: Array<Proof>,
 		options?: {
-			preference?: Array<AmountPreference>;
+			preference?: Preferences;
 			counter?: number;
 			pubkey?: string;
 			privkey?: string;
@@ -215,7 +217,7 @@ class CashuWallet {
 		}
 	): Promise<SendResponse> {
 		if (options?.preference) {
-			amount = options?.preference?.reduce((acc, curr) => acc + curr.amount * curr.count, 0);
+			amount = options?.preference?.sendPreference.reduce((acc, curr) => acc + curr.amount * curr.count, 0);
 		}
 		const keyset = await this.getKeys(options?.keysetId);
 		let amountAvailable = 0;
@@ -538,7 +540,7 @@ class CashuWallet {
 		amount: number,
 		proofsToSend: Array<Proof>,
 		keyset: MintKeys,
-		preference?: Array<AmountPreference>,
+		preference?: Preferences,
 		counter?: number,
 		pubkey?: string,
 		privkey?: string
@@ -550,7 +552,7 @@ class CashuWallet {
 		const keepBlindedMessages = this.createRandomBlindedMessages(
 			totalAmount - amount,
 			keyset,
-			undefined,
+			preference?.keepPreference,
 			counter
 		);
 		if (this._seed && counter) {
@@ -559,7 +561,7 @@ class CashuWallet {
 		const sendBlindedMessages = this.createRandomBlindedMessages(
 			amount,
 			keyset,
-			preference,
+			preference?.sendPreference,
 			counter,
 			pubkey
 		);
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -167,10 +167,10 @@ class CashuWallet {
 		const proofs: Array<Proof> = [];
 		const amount = tokenEntry.proofs.reduce((total, curr) => total + curr.amount, 0);
 		let preference = options?.preference;
-		if (!preference) {
-			preference = getDefaultAmountPreference(amount);
-		}
 		const keys = await this.getKeys(options?.keysetId);
+		if (!preference) {
+			preference = getDefaultAmountPreference(amount, keys);
+		}
 		const { payload, blindedMessages } = this.createSwapPayload(
 			amount,
 			tokenEntry.proofs,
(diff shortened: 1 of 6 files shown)


lollerfirst:
Now the changes are isolated.

src/model/types/index.ts
```diff
@@ -556,6 +556,11 @@ export type AmountPreference = {
 	count: number;
 };
 
+export type Preferences = {
+	sendPreference: Array<AmountPreference>;
+	keepPreference?: Array<AmountPreference>;
+};
```
callebtc, line 562: This is an awesome PR!

If we do introduce breaking changed is on how the preference is passed, what do you guys think of getting rid of the (rather cumbersome?) format for the preference as `sendPreference: [{ amount: 1, count: 3 }]` and instead use a pure array of numbers a la `sendPreference: [1, 1, 1]`? I think I would prefer that.

src/CashuWallet.ts
```diff
@@ -214,15 +216,18 @@ class CashuWallet {
 		amount: number,
 		proofs: Array<Proof>,
 		options?: {
-			preference?: Array<AmountPreference>;
+			preference?: Preferences;
```
callebtc, line 219: How about we deprecate `preference`, do the new thing in a new option called `outputAmounts`, and write a thin translation method that maps `preference` to `outputAmounts`, and issue a deprecation warning?

callebtc (commented):
Really like this PR. We have to decide wether this should be a breaking change in the API or not. We could also deprecate the old method for setting the amount preference (see my comment).

callebtc:
Maybe this is out of scope for this PR but we should also get rid of this behavior of the lib where it overwrites `amount` if `preference` is set. We should throw an error instead of overwriting things and force the implementor to properly use the library.

Example:

```ts
if (options?.preference) {
    amount = options?.preference?.reduce((acc, curr) => acc + curr.amount * curr.count, 0);
}
```

src/CashuWallet.ts
lollerfirst: I second this

src/model/types/index.ts
lollerfirst: Probably a good idea. I think the original motive behind it was noble (not having an array with a billion useless 1s in it) but this is more practical.

gandlafbtc: Yes that was exactly the thought behind it, but I came around to the idea of simpler data structures

gandlafbtc:
looks like this PR needs some conflict resolution. What's the status on this? @lollerfirst

src/model/types/index.ts
lollerfirst: Is this the matter for another PR, thought? Or shall we do it here?

lollerfirst:
> looks like this PR needs some conflict resolution. What's the status on this? @lollerfirst

Conflicts resolved and I made this into a non-breaking change. Now issues a deprecation warning and converts `Array<AmountPreference>` to `Preferences` when needed.

src/model/types/index.ts
gandlafbtc: we can do it in a different PR

src/utils.ts
```diff
@@ -19,37 +19,52 @@ import { bytesToHex, hexToBytes } from '@noble/curves/abstract/utils';
 import { sha256 } from '@noble/hashes/sha256';
 import { decodeCBOR, encodeCBOR } from './cbor.js';
 
-function splitAmount(value: number, amountPreference?: Array<AmountPreference>): Array<number> {
+function splitAmount(
+	value: number,
+	keyset: Keys,
+	amountPreference?: Array<AmountPreference>,
+	order?: string
```
gandlafbtc, line 26: It looks like order is not really used yet, and if it will be used later, typing it as string is a bit ambiguous to me
```suggestion
	isDesc?: boolean
```

src/utils.ts
```diff
@@ -19,37 +19,52 @@ import { bytesToHex, hexToBytes } from '@noble/curves/abstract/utils';
...
-	return chunks;
+	const sortedKeyAmounts: Array<number> = Object.keys(keyset)
+		.map((k) => parseInt(k))
+		.sort((a, b) => b - a);
+	sortedKeyAmounts.forEach((amt) => {
+		let q = Math.floor(value / amt);
+		for (let i = 0; i < q; ++i) chunks.push(amt);
+		value %= amt;
+	});
+	return chunks.sort((a, b) => (order === 'desc' ? b - a : a - b));
```
gandlafbtc, line 45:
```suggestion
	return chunks.sort((a, b) => ( isDesc ? b - a : a - b));
```

test/utils.test.ts
```diff
@@ -1,48 +1,85 @@
...
+	const undef = undefined;
+	test('testing undefined', async () => {
+		const chunks = utils.splitAmount(5, keys, undef);
 		expect(chunks).toStrictEqual([1, 4]);
 	});
 });
 
+describe('test split different key amount', () => {
+	test('testing amount 68251', async () => {
+		const chunks = utils.splitAmount(68251, keys_base10, undefined, 'desc');
```
gandlafbtc, line 68:
```suggestion
		const chunks = utils.splitAmount(68251, keys_base10, undefined, true);
```

test/utils.test.ts
```diff
@@ -1,48 +1,85 @@
...
 	});
 	const illegal: Array<AmountPreference> = [{ amount: 3, count: 2 }];
 	test('testing non pow2', async () => {
-		expect(() => utils.splitAmount(6, illegal)).toThrowError();
+		expect(() => utils.splitAmount(6, keys, illegal)).toThrowError();
 	});
 	const empty: Array<AmountPreference> = [];
 	test('testing empty', async () => {
-		const chunks = utils.splitAmount(5, empty);
+		const chunks = utils.splitAmount(5, keys, empty, 'desc');
```
gandlafbtc, line 56:
```suggestion
		const chunks = utils.splitAmount(5, keys, empty, true);
```

test/utils.test.ts
```diff
@@ -1,48 +1,85 @@
...
-		const chunks = utils.splitAmount(10, tenToOneAndTwo);
+		const chunks = utils.splitAmount(10, keys, tenToOneAndTwo);
 		expect(chunks).toStrictEqual([1, 1, 2, 2, 2, 2]);
 	});
 	const fiveTwelve: Array<AmountPreference> = [{ amount: 512, count: 2 }];
-	test('testing amount 516', async () => {
-		const chunks = utils.splitAmount(518, fiveTwelve);
-		expect(chunks).toStrictEqual([512, 2, 4]);
+	test('testing amount 518', async () => {
+		const chunks = utils.splitAmount(518, keys, fiveTwelve, 'desc');
```
gandlafbtc, line 47:
```suggestion
		const chunks = utils.splitAmount(518, keys, fiveTwelve, true);
```

gandlafbtc (approved):
Nice work!

To me it looks good. What do you think about the suggestions @lollerfirst ? 

I think it would be a bit more expressive, but maybe that's just my opinion. You can either accept or reject all suggestions, and then we'll merge this

lollerfirst:
@gandlafbtc Thank you! I applied the suggestions.

PR #147, opened 2024-07-15
Nut-06-fix-contact

callebtc:
## Description

Implements changes for the contact field in NUT-06 (https://github.com/cashubtc/nuts/pull/117) and monkey-patches responses from old mints so that they can be deserialized. Deprecated code block can be removed once all mints upgraded.

## Changes

- new wallet method `getMintInfo` 
- tests

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -51,12 +51,17 @@ class CashuMint {
 		const data = await requestInstance<GetInfoResponse>({
 			endpoint: joinUrls(mintUrl, '/v1/info')
 		});
-		// BEGIN DEPRECATED 
+		// BEGIN DEPRECATED
 		// Monkey patch old contact field ["email", "me@mail.com"] Array<[string, string]>; to new contact field [{method: "email", info: "me@mail.com"}] Array<MintContactInfo>
 		// This is to maintain backwards compatibility with older versions of the mint
 		if (Array.isArray(data?.contact) && data?.contact.length > 0) {
 			data.contact = data.contact.map((contact: any) => {
-				if (Array.isArray(contact) && contact.length === 2 && typeof contact[0] === 'string' && typeof contact[1] === 'string') {
+				if (
+					Array.isArray(contact) &&
+					contact.length === 2 &&
+					typeof contact[0] === 'string' &&
+					typeof contact[1] === 'string'
+				) {
 					return { method: contact[0], info: contact[1] };
 				}
 				return contact;
--- a/src/CashuMint.ts
+++ b/src/CashuMint.ts
@@ -29,7 +29,7 @@ import {
 	MintQuoteResponsePaidDeprecated,
 	handleMintQuoteResponseDeprecated
 } from './legacy/nut-04.js';
-import { handeMintInfoContactFieldDeprecated } from './legacy/nut-06.js';
+import { handleMintInfoContactFieldDeprecated } from './legacy/nut-06.js';
 /**
  * Class represents Cashu Mint API. This class contains Lower level functions that are implemented by CashuWallet.
  */
@@ -60,7 +60,7 @@ class CashuMint {
 		const response = await requestInstance<GetInfoResponse>({
 			endpoint: joinUrls(mintUrl, '/v1/info')
 		});
-		const data = handeMintInfoContactFieldDeprecated(response);
+		const data = handleMintInfoContactFieldDeprecated(response);
 		return data;
 	}
 	/**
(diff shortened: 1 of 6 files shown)


src/base64.ts
```diff
@@ -20,12 +20,12 @@ function encodeBase64ToJson<T extends object>(base64String: string): T {
 }
 
 function base64urlToBase64(str: string) {
-	return str.replace(/-/g, '+').replace(/_/g, '/').split('=')[0]
+	return str.replace(/-/g, '+').replace(/_/g, '/').split('=')[0];
```
callebtc, line 23: change due to `npm run format`

src/CashuMint.ts
```diff
@@ -48,7 +48,28 @@ class CashuMint {
...
-		return requestInstance<GetInfoResponse>({ endpoint: joinUrls(mintUrl, '/v1/info') });
+		const data = await requestInstance<GetInfoResponse>({
+			endpoint: joinUrls(mintUrl, '/v1/info')
+		});
+		// BEGIN DEPRECATED
+		// Monkey patch old contact field ["email", "me@mail.com"] Array<[string, string]>; to new contact field [{method: "email", info: "me@mail.com"}] Array<MintContactInfo>
+		// This is to maintain backwards compatibility with older versions of the mint
+		if (Array.isArray(data?.contact) && data?.contact.length > 0) {
+			data.contact = data.contact.map((contact: any) => {
+				if (
```
Egge21M, line 59: I have the feeling that this check id a bit to verbose. We should assume that mint info responses implement either the old way or the new way.

src/CashuMint.ts
```diff
@@ -48,7 +48,28 @@ class CashuMint {
...
 		const requestInstance = customRequest || request;
-		return requestInstance<GetInfoResponse>({ endpoint: joinUrls(mintUrl, '/v1/info') });
+		const data = await requestInstance<GetInfoResponse>({
+			endpoint: joinUrls(mintUrl, '/v1/info')
+		});
+		// BEGIN DEPRECATED
+		// Monkey patch old contact field ["email", "me@mail.com"] Array<[string, string]>; to new contact field [{method: "email", info: "me@mail.com"}] Array<MintContactInfo>
+		// This is to maintain backwards compatibility with older versions of the mint
+		if (Array.isArray(data?.contact) && data?.contact.length > 0) {
+			data.contact = data.contact.map((contact: any) => {
```
BilligsterUser, line 58: It would be better to use the `unknown` type instead of `any` for the `contact` parameter in the map function. The `unknown` type is safer because it requires us to explicitly check the type before performing operations on the variable, thus avoiding potential runtime errors. Here’s how you can update the code:

```typescript
if (Array.isArray(data?.contact) && data?.contact.length > 0) { 
    data.contact = data.contact.map((contact: unknown) => {
        // Perform necessary type checks or type assertions here
    });
}
```
---

BilligsterUser (approved):

src/CashuMint.ts
```diff
@@ -48,7 +48,28 @@ class CashuMint {
 		customRequest?: typeof request
 	): Promise<GetInfoResponse> {
 		const requestInstance = customRequest || request;
-		return requestInstance<GetInfoResponse>({ endpoint: joinUrls(mintUrl, '/v1/info') });
+		const data = await requestInstance<GetInfoResponse>({
+			endpoint: joinUrls(mintUrl, '/v1/info')
+		});
+		// BEGIN DEPRECATED
+		// Monkey patch old contact field ["email", "me@mail.com"] Array<[string, string]>; to new contact field [{method: "email", info: "me@mail.com"}] Array<MintContactInfo>
```
Egge21M, line 55: This comment could be removed and the part of the function that handles the deprecated info parsing extracted in a function called `handleDeprecatedInfo`, that way code would be documenting itself. However I am unsure about this repos policies about code comments. cc @gandlafbtc

src/CashuMint.ts
callebtc: I don't understand the issue.

src/CashuMint.ts
callebtc: I think this should be fixed in [ff1f316](https://github.com/cashubtc/cashu-ts/pull/147/commits/ff1f3163ac99c32b2d0f0b799da257ab86d2251c)

src/CashuMint.ts
callebtc: Done

Egge21M (approved):
LGTM

PR #148, opened 2024-07-16
NUT-04 and NUT-05: Add `state` enum field

callebtc:
## Description

Implements: https://github.com/cashubtc/nuts/pull/141

and https://github.com/cashubtc/nuts/pull/136

## Changes

- nothing changes for the implementor 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/types/index.ts
+++ b/src/model/types/index.ts
@@ -146,6 +146,12 @@ export type MeltQuotePayload = {
 	request: string;
 };
 
+export enum MeltQuoteState {
+	UNPAID = 'UNPAID',
+	PENDING = 'PENDING',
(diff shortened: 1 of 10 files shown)


src/model/types/index.ts
```diff
@@ -280,6 +276,13 @@ export type MintQuotePayload = {
 	 */
 	amount: number;
 };
+
+export enum MintQuoteState {
```
Egge21M, line 280: Usually I consider enums harmful, but in this case it should be fine

Egge21M (approved):
LGTM

PR #154, opened 2024-07-18
Migrate to esbuild

Egge21M:
# Fixes: None

## Description

The current build process leaves room for improvement, so the goal of this PR is to get rid of TSC as a transpiler and migrate to using esbuild. TSC will still be part of the pipeline and handle type checking.

## Changes

- Remove TSC transpile step
- Use esbuild to transpile TS -> JS and bundle
- Change exports in `package.json` to the latest standards

## PR Tasks

- [x] Open PR
- [ ] run `npm run format`


Egge21M:
This PR is so much out of date that it will be easier to reopen.

PR #158, opened 2024-07-22
Deprecate `payLnInvoice` helpers, rename `CashuWallet.meltTokens` to `CashuWallet.melt`

### LINKED ISSUE #157: Discuss: deprecate `payLnInvoice` and `payLnInvoiceWithToken` from lib
These helper methods are not needed, the API is now simple enough to leave it up to the implementor to figure out how to do it.

Thoughts?

callebtc:
Removes LnInvoice helper functions (closes #157) and renames `CashuWallet.meltTokens` to `CashuWallet.melt`.

Todo:
- [ ] re-add removed methods and deprecate them so that we don't break implementations. 
- [ ] do the same with `meltTokens`
- [ ] `CashuMint.split` -> `CashuMint.swap` and deprecate split

### DIFF
--- a/migration-1.0.0.md
+++ b/migration-1.0.0.md
@@ -63,7 +63,7 @@ type MeltQuoteResponse = {
 };
 ```
 
-where `quote` is the identifier to pass to `meltTokens()`
+where `quote` is the identifier to pass to `melt()`
 
 ---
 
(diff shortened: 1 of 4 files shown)


Egge21M:
@callebtc I guess we can close this, right?

PR #159, opened 2024-07-26
Add fees and coin selection

callebtc:
Introduces support for fees for spending ecash with the mint. Fees are published in the `/v1/keysets` response (see NUT-02). Also introduces a new input coinselection algorithm to minimize fees when spending coins and an output selection, to maintain a diverse set of amounts in the user's wallet, increasing the chance for being able to make offline transactions in the future.

### Automatic input selection
There is a rather sophisticated coinselection algorithm now that will select with fees in mind. The library will automatically fetch the keyset of the mint in order to determine the fees during selection. The implementor can decide to use `includeFees = true` in various class methods in order to include fees in the selection so that the selection can be spent. This can be used in situations where the sender pays the fees instead of the receiver (and the receiver can receive a predictable amount). It is necessary to use if we perform coin selection to pay a Lightning invoice afterwards:

```ts
const { keep: proofsToKeep, send: proofsToSend } = await wallet.send(amountToSend, proofs, {
	includeFees: true
});
const meltResponse = await wallet.meltProofs(meltQuote, proofsToSend);
```

### Output amounts
`AmountPreference` is removed, use `OutputAmounts` now, which can be used to define `sendAmounts` and `keepAmounts` which are now both simple arrays of numbers. Previously we could only define amounts to send through `AmountPreference`. 

### Automatic output amount selection
Users can use `proofsWeHave` to specify in various methods which proofs are present in the wallet. This will make the library to choose output amounts to refill the wallet with missing denominations:
```ts
const { keep: proofsToKeep, send: proofsToSend } = await wallet.send(amountToSend, proofs, {
            proofsWeHave: proofs
});
```
The target defaults to `DEFAULT_DENOMINATION_TARGET = 3;` but can be defined in the constructor of the `CashuWallet` class.

## Renaming 
- in `SendResponse`, `returnChange` is now called `keep`
- `mintTokens` is now called `mintProofs`
- `meltTokens` is now called `meltProofs`
- `CashuMint.split` is now called `CashuMint.swap`

## Legacy code
I suggest we get rid of all functions a la `PayLNInvoice` and force the implementor to understand how to use the library themselves instead of hiding crucial details. This should be done in a separate PR after merging this.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


callebtc:
I think now only tests are missing.

callebtc:
Both remaining todos are done, all integration tests now run with a mint with fees. Ready for review 🙏

callebtc:
@lollerfirst I can't tag you as a reviewer but if you find time, please have a look as well

lollerfirst (approved):
Amazing work! ♥️

I don't think anything needs changing except maybe in `splitAmount` we could put a boolean to specify which order instead of  `"asc"` and `"desc"` (That was dumb of me :) )

src/utils.ts
```diff
@@ -25,68 +24,100 @@ import { PaymentRequest } from './model/PaymentRequest.js';
...
+	sortedKeyAmounts.forEach((amt: number) => {
 		const q = Math.floor(value / amt);
 		for (let i = 0; i < q; ++i) chunks.push(amt);
 		value %= amt;
 	});
-	return chunks.sort((a, b) => (isDesc ? b - a : a - b));
+	return chunks.sort((a, b) => (order === 'desc' ? b - a : a - b));
+}
+
+function getKeepAmounts(
```
Egge21M, line 56: This function seems to not be covered by unit tests. I will not enforce a 100% coverage, but please lmk if this is on purpose.

Egge21M (changes requested):
I think these are great changes and additions! Thank you so much for tackling this @callebtc <3

This PR addresses many issues that existed, however it does so in a non-backwards compatible way. I have just created `v2.0.0` branch off of `development`. I would like to merge this into the `v2.0.0` branch, so that we can base the next major release off that.

src/utils.ts
```diff
@@ -25,68 +24,100 @@ import { PaymentRequest } from './model/PaymentRequest.js';
...
 	return amount in keyset;
 }
 
-function getPreference(
-	amount: number,
-	keyset: Keys,
-	preferredAmounts: Array<AmountPreference>
-): Array<number> {
+function getPreference(amount: number, keyset: Keys, split: Array<number>): Array<number> {
 	const chunks: Array<number> = [];
```
gudnuf, line 111: is this function still needed? it's only used in `splitAmount` and seems all that it does is make sure all the split amounts have a corresponding key.
Also, the `amount` parameter is not used any more in this function

callebtc:
> I think these are great changes and additions! Thank you so much for tackling this @callebtc <3
> 
> This PR addresses many issues that existed, however it does so in a non-backwards compatible way. I have just created `v2.0.0` branch off of `development`. I would like to merge this into the `v2.0.0` branch, so that we can base the next major release off that.

Not sure I quite get it but I'll let you do the branch management :) Should I reopen this PR? Are there any other changes we want to releave in a 1.x update? Otherwise, I suggest we start working on the 2.0.0 release. I hope that I won't have to rebase this PR anymore 😅

Egge21M:
> Not sure I quite get it but I'll let you do the branch management :) Should I reopen this PR? Are there any other changes we want to release in a 1.x update? Otherwise, I suggest we start working on the 2.0.0 release. I hope that I won't have to rebase this PR anymore 😅

I see your point. We are close enough to a new major release to do it on ´development´ as well. Either way, there should not be any rebase necessary :)

gandlafbtc (changes requested):
Yuge PR! thanks.

There are a lot of new functions. Please make sure that functions in tthe CashuWallet class that should not be used externally are marked `private`

Also, please add the minimal docs to the functions that are accessible to the implementoors

README.md
```diff
@@ -77,21 +77,30 @@ if (mintQuoteChecked.state == MintQuoteState.PAID) {
...
 // store proofsToKeep in wallet ..
 
-const meltResponse = await wallet.meltTokens(meltQuote, proofsToSend);
+const meltResponse = await wallet.meltProofs(meltQuote, proofsToSend);
 // store meltResponse.change in wallet ..
 ```
 
+#### Create a token and receive it
+
+```typescript
```
gandlafbtc, line 100: is there something missing here?

src/CashuWallet.ts
```diff
@@ -365,26 +626,44 @@ class CashuWallet {
...
-			preference?: Array<AmountPreference>;
+			outputAmounts?: OutputAmounts;
+			proofsWeHave?: Array<Proof>;
 			counter?: number;
 			pubkey?: string;
 		}
 	): Promise<{ proofs: Array<Proof> }> {
 		const keyset = await this.getKeys(options?.keysetId);
+		if (!options?.outputAmounts && options?.proofsWeHave) {
+			options.outputAmounts = {
```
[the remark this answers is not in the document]
gandlafbtc: `options` should be defined, due to `&& options?.proofsWeHave` conditional. Mutating state outside of the function is real though, which I agree should be avoided. In this case though, it might not be as bad, since it will update the state to what it should have been in the first place.

If you got a good idea for a cleaner solution, feel free to add it to the PR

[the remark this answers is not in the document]
Egge21M: Yep, you are right. Looking at this again this assignment might even be pretty safe, as it only happens when `outputAmounts` is falsy.

README.md
callebtc: Oops

callebtc:
@gandlafbtc you're the best. thank you for doing all that work

gandlafbtc (approved):
LGTM! Nice work!

src/CashuWallet.ts
```diff
@@ -47,87 +41,232 @@ import {
...
-		if (options?.keys) {
-			this._keys = options.keys;
-			this._unit = options.keys.unit;
+		if (options?.keysets) this._keysets = options.keysets;
+		if (options?.denominationTarget) {
+			this._denominationTarget = options.denominationTarget;
+		}
+
+		if (options?.loadMint) {
+			this.loadMint()
```
gandlafbtc, line 107: we should either add events here, or do it do it async outside of the constructor. right now, we don't know when the mint has been loaded

src/CashuWallet.ts
```diff
@@ -194,96 +331,286 @@ class CashuWallet {
...
+			privkey?: string;
+			keysetId?: string;
+			offline?: boolean;
+			includeFees?: boolean;
+		}
+	): Promise<SendResponse> {
+		if (sumProofs(proofs) < amount) {
+			throw new Error('Not enough funds available to send');
+		}
+		const { keep: keepProofsOffline, send: sendProofOffline } = this.selectProofsToSend(
```
gandlafbtc, line 365: doesn't this need to consider `outputAmounts`, in case `includeFees` is set to `true`? Or how do we make sure the fees are covered? `IncludeFees` is thought to cover the fees for the receiver, right?

gandlafbtc: we also should add a function, to return the fees to the user, before we perform the split. I think most wallets will need to show that:

1. user enters amount
2. display amount + fees = total leaving the wallet
3. perform send

src/CashuWallet.ts
Egge21M: I agree. Methods that require a up to date key set could call this method and await its return. Potentially with a cache invalidation logic

callebtc: > Methods that require a up to date key set could call this method and await its return. Potentially with a cache invalidation logic

out of scope

src/CashuWallet.ts
callebtc: > doesn't this need to consider `outputAmounts`, in case `includeFees` is set to `true`? Or how do we make sure the fees are covered? `IncludeFees` is thought to cover the fees for the receiver, right?

I don't understand. `includeFees` should includes the fees in `sendProofOffline`. What does this have to do with outputs?

src/CashuWallet.ts
callebtc: > Methods that require a up to date key set could call this method and await its return.

They do: `await loadKeys()` when needed etc.

> Potentially with a cache invalidation logic

out of scope

Egge21M: Then this should be removed from the constructor.

callebtc: Done

src/CashuWallet.ts
Egge21M: I think this can be resolved. `outputAmounts` does only determine changeOutputs kept by the sender. To cover the fees for the receiving end we either include then in the sent proofs in `selectProofsToSend` or in `swap`. If `offline` is true, fee is included in `sendProofOffline` and if we swap it is included in `send` returned by `swap`

[ab2fe7e5-d4f9-47f8-9a81-1ef6b0ec332a](https://github.com/cashubtc/cashu-ts/pull/159/files#diff-ed2a551c11fe95a44ccc731d64f73eea7ced34250da86272f7ec5a0ad6a8781bR371)

Egge21M (approved):
LGTM! Thank you again for this @callebtc !

PR #161, opened 2024-07-28
Update Nutshell Docker image for integration tests

callebtc:
This pull request updates the Nutshell Docker image used for integration tests to the latest version. The previous version was 0.15.2, and this PR updates it to the latest version available.

### DIFF
--- a/.github/workflows/nutshell-integration.yml
+++ b/.github/workflows/nutshell-integration.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - name: Pull and start mint
         run: |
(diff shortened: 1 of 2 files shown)


gandlafbtc (approved):
might break the tests unexpectedly sometimes though...

Egge21M:
TBH I am not a fan of using a moving target version for the pipeline. As already mentioned this might break tests from time, especially because it will also make major version jumps. I would suggest that we increment versions manually from time to time instead

callebtc:
I opened this PR to minimize manual changes necessary when new nutshell releases are out in the future. We can always fix the tests if they break (which they shouldn't if everything goes right). 

If this is still undesirable, I'm fine with that. Please close the PR in that case @Egge21M.

gandlafbtc:
I also lean more towards NACK actually. Imagine the scenario where we introduce a new feature that breaks the tests, but at the same time the `latest` version of nutshell also breaks the tests. Then we have chaos. I don't think it's too much of a hassle to update it manually and in return we are able to determine where breaking changes come from

PR #176, opened 2024-09-08
NUT-05: description for invoice, integration test without bolt11 decode

callebtc:
Implements https://github.com/cashubtc/nuts/pull/144/

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -339,7 +339,7 @@ class CashuWallet {
 		const mintQuotePayload: MintQuotePayload = {
 			unit: this._unit,
 			amount: amount,
-			description: description || null,
+			description: description || null
 		};
 		return await this.mint.createMintQuote(mintQuotePayload);
 	}
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -339,7 +339,7 @@ class CashuWallet {
 		const mintQuotePayload: MintQuotePayload = {
 			unit: this._unit,
 			amount: amount,
-			description: description || null
+			description: description
 		};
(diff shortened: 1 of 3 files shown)


src/CashuWallet.ts
```diff
@@ -331,12 +332,14 @@ class CashuWallet {
...
+	 * @param description optional description for the mint quote
 	 * @returns the mint will return a mint quote with a Lightning invoice for minting tokens of the specified amount and unit
 	 */
-	async createMintQuote(amount: number) {
+	async createMintQuote(amount: number, description?: string) {
 		const mintQuotePayload: MintQuotePayload = {
 			unit: this._unit,
-			amount: amount
+			amount: amount,
+			description: description || null
```
Egge21M, line 342: Is the description field mandatory on future mint requests? If not, I would prefer to not pass the null value here. If left undefined, description will be dropped from the payload completely, while if set null it will be `{"description":null}`

src/model/types/wallet/payloads.ts
```diff
@@ -98,6 +98,10 @@ export type MintQuotePayload = {
 	 * Amount to be minted
 	 */
 	amount: number;
+	/**
+	 * Description for the invoice
+	 */
+	description: string | null;
```
Egge21M, line 104: Same here. If not mandatory this should be `description?: string`

Egge21M (changes requested):

Egge21M:
LGTM. What's the support on this? Should we validate the reponse and add a warning if the invoice memo does not contain the expected message?

callebtc:
> LGTM. What's the support on this? Should we validate the reponse and add a warning if the invoice memo does not contain the expected message?

No, that would be overkill. This is enough.

PR #180, opened 2024-09-21
add pipeline to build RC from staging branch

gandlafbtc:
# Fixes: Adding pipeline to automatically create RC from staging branch

## Changes

- Add GH action

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/.github/workflows/nextVersion.yml
+++ b/.github/workflows/nextVersion.yml
@@ -1,0 +1,24 @@
+name: Publish Package to npmjs
+permissions:
+  contents: write
+  id-token: write
+on:
+  push:
+    branches:
+      - staging
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm i
+      - run: npm run compile
+      - run: npm version prerelease --preid=rc --no-git-tag-version
+      - run: git push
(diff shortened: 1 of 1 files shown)


callebtc:
Good work, I know too little about NPM to give a good judgement here though :)

Egge21M:
@gandlafbtc great addition! Could you explain the flow for future prereleases?

We would increment the minor version number on development and then merge to staging? We would still have to add a tag ourselves, right?

I wonder if it wouldn't be better if we had version specific branches. That way we could we could work on separate versions at the same time (e.g. current release and version 2.x).

gandlafbtc:
It would automatically create the incremented minor version and create an RC release under the `next` npm tag. I have git tags turned off, but maybe it's better to turn it on, so it's easier to reference? 

I'm open to the idea of version branches, I haven't had too much experience with it though. So you would have to take the lead to structure that @Egge21M

Egge21M (approved):

Egge21M:
> It would automatically create the incremented minor version and create an RC release under the `next` npm tag. I have git tags turned off, but maybe it's better to turn it on, so it's easier to reference?

Awesome! I was not aware that `version prerelease` would automatically bump to the next minor version. I think having automatic git tags could be very helpful. Ideally we would come up with a system where a merge to a branch kicks of the whole release, and tags are part of that IMO.

> I'm open to the idea of version branches, I haven't had too much experience with it though. So you would have to take the lead to structure that @Egge21M

Happy to do that. Lets get this one merge first, and discuss the best way forward in a different PR. I have put this on my list and will propose something next week.

PR #188, opened 2024-10-14
P2PK for multiple pubkeys in one operation

gudnuf:
adds an optional `pubkey` field to `AmountPreference`

now when the blinded messages are constructed, a P2PK lock can be created for each preference creating ecash locked to many pubkeys


gandlafbtc (changes requested):
In v2, amount preferences will be changed to simple number arrays. I think we also should change this PR accordingly to accommodate for that, so it is not deprecated right away. So it would probably just be an array of pubkeys, to go along with the preferred output amounts.

new amount preference format will be enforced here:
https://github.com/cashubtc/cashu-ts/pull/159

You can rebase onto there, if that helps you make sure everything works ok

PR #190, opened 2024-10-16
Wallet: Remove bip39 dependency

Egge21M:
## Description

Depending on the BIP39 package means bundling the wordlist most of the time, which increases bundle size dramatically. This removed the dependency on bip39. People that want to use deterministic secrets can import a seed that they derive from bip39 words themselves.

## Changes

- Removed bip39 package
- *BREAKING CHANGE*: Removes option to instantiate `CashuWallet` with a mnemonic

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


gudnuf (approved):
lgtm

callebtc:
I noticed that we need to make the same adjustments in [cashubtc/cashu](https://github.com/cashubtc/cashu-crypto-ts)

prusnak:
I think it might be worth asking @scure/bip39 to provide a minimal version which bundles only english wordlist to decrease the size and keep this minimal dependency in

prusnak:
Hm checking this more, it seems the non-English wordlist should not be used unless explicitly imported. 

Is this PR really about removing 37kB of code and 12kB of English wordlist (49kB in total)?

Egge21M:
> Hm checking this more, it seems the non-English wordlist should not be bundled unless explicitly imported.

I think this actually comes down to the bundler used and how it is able to resolve the imports.

> Is this PR really about removing 37kB of code and 12kB of English wordlist (49kB in total)?

49kb might not seem too much, however it is about 10% of the current bundle size and we are talking about a cashu-unrelated dependency that can easily be managed by the consumer.

callebtc (approved):

gandlafbtc (commented):
hmm not sure if I agree with removing the library from cashu-ts. Shouldn't a SDK come with all the required tooling pre-installed? I understand that it's a big hit on the bundle size. But I assume that 99% of projects that use cashu-ts will need to install a bip39 lib anyway, so in the end, what is the gain?

Egge21M:
> hmm not sure if I agree with removing the library from cashu-ts. Shouldn't a SDK come with all the required tooling pre-installed? I understand that it's a big hit on the bundle size. But I assume that 99% of projects that use cashu-ts will need to install a bip39 lib anyway, so in the end, what is the gain?

I don't agree with that reasoning. cashu-ts is not a full fleshed SDK. It keeps stuff simple, but leaves many things in the responsibility of the consumer. Like for example proof storage or multi mint management. I think it should focus on the things that are absolutely necessary to build a cashu-enabled application.

BIP39 might be adjacent to cashu, but in the end cashu just cares about entropy, not seed words. In the same way nostr can be considered adjacent to cashu, but bundling a nostr dependency would not be the right way either.

If I was to find a middle path here, then I would say we need to fix the bundling, move from package.json's `main` to `exports` and have the BIP39 logic in a separate export, which ensures correct tree-shaking. However older bundlers will still bundle it regardless.

However I still think removing this dependency is the right now and we should strive for less dependencies at all times, as they do not only bloat, but are also potential security risks.

gandlafbtc:
Fair points. If we end up removing it, let's make at least sure that it is well documented on how devs can chose and implement bip39 libraries to use with cashu-ts. 

I do believe that deterministic nuts are important for most usecases, and my main concern is that we elevate the barrier to entry for using them if we make it more ambiguous on how to implement (by forcing devs to choose their own implementations and such), that they will just skip it, especially if they don't know what to do. It should maybe even be mentioned on the main readme page then, something like "to use seed phrase backup, you need a bip39 library of your choosing". Maybe we can also mention things there like: "you have to handle persistence for proofs and counters yourself etc."

Du you think you could add this to the PR?

callebtc:
> "to use seed phrase backup, you need a bip39 library of your choosing"

That's not completely true. You need bip39 to convert a mneominc to a seed but you could also use any other private key. 

We can add implementation details in the readme, I think that's a good idea. But we should not delay PRs due to bikeshedding over details. @gandlafbtc you can push your updated README here or after the merge, whatever comes first.

gandlafbtc (approved):
After thinking more about it, this PR fits the overall direction the library is taking. 

LGTM

Issue #186, opened 2024-10-12
Types: Rewrite Token type


Egge21M:
The Token type that is a central part of this library is a leftover from multi-mint/token times and does not reflect the actual modern cashu token.

The TokenEntry array is outdated, no longer used and most parts of cashu-ts navigate around this by only accessing the first index of a TokenEntry array.

There is no reason to keep this and we should remove it. However this requires a big change that will touch big parts of the lib and should probably be done for a major version bump.

Egge21M:
Looking further into this I would use the structure of the v4 token as basis. That would leave use with a type like this:

```ts
type Token = {
  mint: string;
  memo: string;
  unit: string;
  proofs: Array<Proof>
}
```

PR #193, opened 2024-10-23
Restructured Token type

Egge21M:
# Fixes: #186

## Description

The old `Token` type is a leftover from MultiToken and MultiMint. Both got never really got handled by cashu-ts (it most of the time extracted the first token inside a token object and worked with that). Now that v4 is the new standard it is time to restructure the `Token`.

## Changes

- Restructured the `Token` type
- Adjust all methods to work with the new type

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


Egge21M:
FYI: I realised that two things are still missing:

- adjust integration tests
- adjust token v3 encoding (not sure why we do not have a unit test for this, will add them)

src/CashuWallet.ts
```diff
@@ -264,54 +263,27 @@ class CashuWallet {
...
-			pubkey?: string;
-			privkey?: string;
-		}
-	): Promise<Array<Proof>> {
-		const proofs: Array<Proof> = [];
 		const keys = await this.getKeys(options?.keysetId);
 		const amount =
-			tokenEntry.proofs.reduce((total: number, curr: Proof) => total + curr.amount, 0) -
-			this.getFeesForProofs(tokenEntry.proofs);
+			token.proofs.reduce((total: number, curr: Proof) => total + curr.amount, 0) -
```
gudnuf, line 268: we now how `sumProofs` in src/utils.ts fyi

src/utils.ts
```diff
@@ -178,22 +183,13 @@ export function getEncodedToken(token: Token): string {
...
-			}
-		}
-		for (let j = 0; j < token.token[i].proofs.length; j++) {
-			const proof = token.token[i].proofs[j];
-			if (idMap[proof.id]) {
-				idMap[proof.id].push(proof);
-			} else {
-				idMap[proof.id] = [proof];
-			}
+			idMap[proof.id] = [proof];
```
gudnuf, line 192: What do you think about adding a check to make sure the `proof.id` is hex? 

On boardwalk, I was receiving V3 tokens with Cashu V0 keyset Id's, then when I would try to send and encode as a V4 token the `hexToBytes(id)` would error and say something like "Error: expected hex". I only suggest adding a check so that there can be a more clear error message: "Cannot encode proofs with a Cashu V0 keyset ID into a V4 token".

Egge21M: Yes I totally agree. I wanted to implement something like this in https://github.com/cashubtc/cashu-ts/pull/194 to keep things separated.

PR #199, opened 2024-10-29
Add `CashuWallet.checkProofsStates()`

callebtc:
Add `checkProofsStates` method to `CashuWallet` to get the spent enums from the mint, should retire `checkProofsSpent` which throws most information away.

I would prefer getting rid of `checkProofsSpent` entirely. If you agree, please delete it or let me know.

Edit: we're getting rid of `checkProofsSpent`.

### DIFF
--- a/migration-2.0.0.md
+++ b/migration-2.0.0.md
@@ -36,7 +36,7 @@ export type OutputAmounts = {
 
 #### `checkProofsStates` replaces `checkProofsSpent`
 
-To check the state of a `Proof`, call `CashuWallet.checkProofsStates`. `checkProofsStates` now returns an array of `CheckStateEnum` for each proof provided. States are now `CheckStateEnum.SPENT`, `CheckStateEnum.UNSPENT`, and `CheckStateEnum.PENDING`.
+To check the state of a `Proof`, call `CashuWallet.checkProofsStates`. `checkProofsStates` now returns an array of `ProofState`'s, one for each `Proof` provided. The spent states are in `ProofState.state` and can have the values `CheckStateEnum.SPENT`, `CheckStateEnum.UNSPENT`, and `CheckStateEnum.PENDING`. `ProofState` also contains a `witness` if present.
 
 #### renamed functions
 
--- a/migration-2.0.0.md
+++ b/migration-2.0.0.md
@@ -34,6 +34,10 @@ export type OutputAmounts = {
 };
 ```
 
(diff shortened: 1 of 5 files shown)


Egge21M:
Can we be sure that a mint will ALWAYS return the state in the same order as it received the proofs? This is not explicitly defined in NUT07 and if the mint does actually not do that, we need to take care of this in order to avoid lost nuts.

callebtc:
> Can we be sure that a mint will ALWAYS return the state in the same order as it received the proofs? This is not explicitly defined in NUT07 and if the mint does actually not do that, we need to take care of this in order to avoid lost nuts.

Agreed, good point. Fixed in code and fixed in the spec: https://github.com/cashubtc/nuts/pull/181

callebtc:
> Can we be sure that a mint will ALWAYS return the state in the same order as it received the proofs? This is not explicitly defined in NUT07 and if the mint does actually not do that, we need to take care of this in order to avoid lost nuts.

Good point. Fixed in the code and also fixed in the spec: https://github.com/cashubtc/nuts/pull/181

Egge21M:
Thanks for updating! What do you think about this instead?

```ts
async checkProofsStates(proofs: Array<Proof>): Promise<Array<ProofState>> {
	const enc = new TextEncoder();
	const Ys = proofs.map((p: Proof) => hashToCurve(enc.encode(p.secret)).toHex(true));
	// TODO: Replace this with a value from the info endpoint of the mint eventually
	const BATCH_SIZE = 100;
	const states: Array<ProofState> = [];
	for (let i = 0; i < Ys.length; i += BATCH_SIZE) {
		const YsSlice = Ys.slice(i, i + BATCH_SIZE);
		const { states: batchStates } = await this.mint.check({
			Ys: YsSlice
		});
		const stateMap: { [y: string]: ProofState } = {};
		batchStates.forEach((s) => {
			stateMap[s.Y] = s;
		});
		for (let j = 0; j < YsSlice.length; j++) {
			const state = stateMap[YsSlice[j]];
			if (!state) {
				throw new Error('Could not find state for proof with Y: ' + YsSlice[j]);
			}
		states.push(state);
		}
	}
	return states;
}
```

By introducing a map for the returned states, we can check whether a state is present in constant time. Before the worst case scenario was a quadratic + running time

callebtc:
> By introducing a map for the returned states, we can check whether a state is present in constant time. Before the worst case scenario was a quadratic + running time

sure go ahead and push and merge it

Issue #156, opened 2024-07-22
[Feature request] Support for DLEQ proofs


callebtc:
[DLEQ proof spec](https://github.com/cashubtc/nuts/blob/main/12.md). We should support: 
1) verify DLEQ proofs whenever blind signatures are received from a mint that include proofs
2) option to send DLEQ proofs to another user
3) ability to verify DLEQ proofs that we received from another user

lollerfirst:
@callebtc I figured that before we do this we need to implement the necessary tools in the cashu-crypto-ts module.
See my PR: https://github.com/cashubtc/cashu-crypto-ts/pull/2

PR #200, opened 2024-10-29
Support for NUT-12

lollerfirst:
# Fixes: #156 

## Changes

- added optional `dleq` and `dleqValid` fields to `Proof`. `dleq` holds a `SerializedDLEQ` proof, while `dleqValid`, if present, signals whether it has been correctly verified.
- `CashuWallet.meltProofs` and `CashuWallet.createSwapPayload` strip the `dleq` field off of each `Proof`, if there is any. (No privacy leak).
- `CashuWallet.construct_proofs` now unpacks and verifies DLEQ proofs returned by the Mint, if any. It will create `Proof` with:
   * `dleqValid == true` if the DLEQ is present and verified
   * `dleqValid == undefined` if DLEQ not present
   * `dleqValid == false` if DLEQ present and verification failed. 
- added boolean option `includeDleq` to `CashuWallet.send`:
   * if `includeDleq == true`, makes sure the selected `send` proofs each have a defined `dleq` field;
   * if `includeDleq == false`, strip off the `dleq` field from `send`;
   * if `includeDleq == undefined`, do not intervene.
- added boolean option `requireDleq` to `CashuWallet.receive`:
   * if `requireDleq == true`, verifies each proof's dleq for correctness.
   * if `requireDleq == false || requireDleq == undefined`, does not verify dleq.
- added private helper function `CashuWallet.requireDLEQ`. Separates the logic for requiring valid dleqs from a token from the function `receive`.
- added `numberToHexPadded64` to the utilities. This function takes in a `bigint` and returns an a hex string prefixed to 64 characters with '0'.
- tests 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -1,4 +1,4 @@
-import { bytesToHex, randomBytes } from '@noble/hashes/utils';
+import { bytesToHex, hexToBytes, randomBytes } from '@noble/hashes/utils';
 import { CashuMint } from './CashuMint.js';
 import { BlindedMessage } from './model/BlindedMessage.js';
 import {
@@ -38,7 +38,8 @@ import {
 	deriveSeedFromMnemonic
 } from '@cashu/crypto/modules/client/NUT09';
 import { createP2PKsecret, getSignedProofs } from '@cashu/crypto/modules/client/NUT11';
-import { type Proof as NUT11Proof } from '@cashu/crypto/modules/common/index';
+import { type Proof as NUT11Proof, DLEQ } from '@cashu/crypto/modules/common/index';
+import { verifyDLEQProof_reblind } from '@cashu/crypto/modules/client/NUT12';
 
 /**
  * The default number of proofs per denomination to keep in a wallet.
@@ -276,6 +277,7 @@ class CashuWallet {
 	 * @param options.counter? optionally set counter to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
 	 * @param options.pubkey? optionally locks ecash to pubkey. Will not be deterministic, even if counter is set!
 	 * @param options.privkey? will create a signature on the @param tokenEntry secrets if set
+	 * @param options.requireDLEQ? optionally require a DLEQ proof from the mint
 	 * @returns {Promise<Array<Proof>>} New token entry with newly created proofs, proofs that had errors
 	 */
 	async receiveTokenEntry(
@@ -286,6 +288,7 @@ class CashuWallet {
 			counter?: number;
 			pubkey?: string;
 			privkey?: string;
+			requireDLEQ?: boolean;
 		}
 	): Promise<Array<Proof>> {
 		const proofs: Array<Proof> = [];
@@ -303,11 +306,13 @@ class CashuWallet {
 			options?.privkey
 		);
 		const { signatures } = await this.mint.swap(payload);
+		const requireDleq = options?.requireDLEQ;
 		const newProofs = this.constructProofs(
 			signatures,
 			blindingData.blindingFactors,
 			blindingData.secrets,
-			keys
+			keys,
+			requireDleq ?? false
 		);
 		proofs.push(...newProofs);
 		return proofs;
@@ -490,6 +495,7 @@ class CashuWallet {
 	 * @param options.proofsWeHave? optionally provide all currently stored proofs of this mint. Cashu-ts will use them to derive the optimal output amounts
 	 * @param options.pubkey? optionally locks ecash to pubkey. Will not be deterministic, even if counter is set!
 	 * @param options.privkey? will create a signature on the @param proofs secrets if set
+	 * @param options.requireDLEQ? optionally require a DLEQ proof from the mint.
 	 * @returns promise of the change- and send-proofs
 	 */
 	async swap(
@@ -503,6 +509,7 @@ class CashuWallet {
 			privkey?: string;
 			keysetId?: string;
 			includeFees?: boolean;
+			requireDLEQ?: boolean;
 		}
 	): Promise<SendResponse> {
 		if (!options) options = {};
@@ -575,11 +582,13 @@ class CashuWallet {
 			options?.privkey
 		);
 		const { signatures } = await this.mint.swap(payload);
+		const requireDleq = options?.requireDLEQ;
 		const swapProofs = this.constructProofs(
 			signatures,
 			blindingData.blindingFactors,
 			blindingData.secrets,
-			keyset
+			keyset,
+			requireDleq ?? false
 		);
 		const splitProofsToKeep: Array<Proof> = [];
 		const splitProofsToSend: Array<Proof> = [];
@@ -603,6 +612,7 @@ class CashuWallet {
 	 * @param start set starting point for count (first cycle for each keyset should usually be 0)
 	 * @param count set number of blinded messages that should be generated
 	 * @param options.keysetId set a custom keysetId to restore from. keysetIds can be loaded with `CashuMint.getKeySets()`
+	 * @param options.requireDLEQ require a DLEQ proof
 	 * @returns proofs
 	 */
 	async restore(
@@ -610,6 +620,7 @@ class CashuWallet {
 		count: number,
 		options?: {
 			keysetId?: string;
+			requireDLEQ?: boolean;
 		}
 	): Promise<{ proofs: Array<Proof> }> {
 		const keys = await this.getKeys(options?.keysetId);
@@ -633,9 +644,15 @@ class CashuWallet {
 		const validSecrets = secrets.filter((_: Uint8Array, i: number) =>
 			outputs.map((o: SerializedBlindedMessage) => o.B_).includes(blindedMessages[i].B_)
 		);
-
+		const requireDleq = options?.requireDLEQ;
 		return {
-			proofs: this.constructProofs(promises, validBlindingFactors, validSecrets, keys)
+			proofs: this.constructProofs(
+				promises,
+				validBlindingFactors,
+				validSecrets,
+				keys,
+				requireDleq ?? false
+			)
 		};
 	}
 
@@ -672,6 +689,7 @@ class CashuWallet {
 	 * @param options.outputAmounts? optionally specify the output's amounts to keep and to send.
 	 * @param options.counter? optionally set counter to derive secret deterministically. CashuWallet class must be initialized with seed phrase to take effect
 	 * @param options.pubkey? optionally locks ecash to pubkey. Will not be deterministic, even if counter is set!
+	 * @param options.requireDLEQ? optionally require a DLEQ proof.
 	 * @returns proofs
 	 */
 	async mintProofs(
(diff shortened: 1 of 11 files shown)


src/CashuWallet.ts
```diff
@@ -987,23 +1022,53 @@ class CashuWallet {
...
+			const A = pointFromHex(keyset.keys[p.amount]);
+			const proof = constructProofFromPromise(blindSignature, r, secret, A);
+			if (verifyDLEQ) {
+				if (dleq == undefined) {
+					throw new Error('DLEQ verification required, but none found');
+				}
+				if (!verifyDLEQProof_reblind(secret, dleq, proof.C, A)) {
+					throw new Error('DLEQ verification failed');
+				}
+			}
```
callebtc, line 1061:
```suggestion
			if (verifyDLEQ) {
				if (dleq == undefined) {
					throw new Error('DLEQ verification required, but none found');
				}
				if (!verifyDLEQProof_reblind(secret, dleq, proof.C, A)) {
					throw new Error('DLEQ verification failed');
				}
			}
```

It might be dangerous to throw an error here, we might still want to keep the proofs, even if the verification failed. I have no good solution for this case though.

gudnuf: > It might be dangerous to throw an error here

Agreed.

What if cashu-ts just constructs the dleq, then its up to the implementer to verify and handle failures?

The other idea I had was have some sort of flag either on the `Proof` or in the return value of `constructProofs` that indicates the verification failed, but that gets gross:

```typescript
/**
 * represents a single Cashu proof.
 */
export type Proof = {
	...

         /**
	 * DLEQ proof
	 */
	dleq?: SerializedDLEQ;
        /**
        * Is the DLEQ proof valid
        */
         dleqValid?: boolan;
};
````

Comment (lollerfirst) [reply src/CashuWallet.ts:1061]: I really like @gudnuf 's idea

Comment (lollerfirst) [reply src/CashuWallet.ts:1061]: What do you think @callebtc

Comment (Egge21M) [reply src/CashuWallet.ts]: nostr-tools uses a Symbol to signal verified events: https://github.com/nbd-wtf/nostr-tools/blob/master/core.ts

**Comment by lollerfirst** [comment]
Unfortunately it seems we need to fix cashu-crypto-ts's package.json again...

File: src/CashuWallet.ts (hunk 1 above)
Comment (lollerfirst) [reply src/CashuWallet.ts]: @Egge21M What changes from a normal boolean flag?

Comment (Egge21M) [reply src/CashuWallet.ts]: Symbols are a bit "safer" than regular properties. The are not enumerable and have to be accessed very purposefully. Also, because they are not enumerable they will be excluded by default by methods like JSON.stringify

**Comment by lollerfirst** [comment]
So the integration tests with Nutshell work fine. I might try and other ones better cover cases where DLEQ is not verified correctly.

I couldn't create normal tests because the mocked keyset does not provide the corresponding private keys so I can't create the dleq for the mocked proof.

**Comment by Egge21M** [comment]
Great stuff! LMK when this is ready for review

**Comment by lollerfirst** [comment]
FYI I tried using a symbol for `dleqValid` before but got some error.

File: src/CashuWallet.ts [hunk 2]
```diff
@@ -299,8 +314,13 @@ class CashuWallet {
 			keysetId?: string;
 			offline?: boolean;
 			includeFees?: boolean;
+			includeDleq?: boolean;
 		}
 	): Promise<SendResponse> {
+		if (options?.includeDleq ?? false) {
```
Egge21M, line 320: Why use nullish coalescing here?

src/CashuWallet.ts
```diff
@@ -299,8 +314,13 @@ class CashuWallet {
...
+			// only pick the ones with a DLEQ proof
+			proofs = proofs.filter((p: Proof) => p.dleq != undefined);
```
Egge21M, line 322: If `includeDleq` controls wether or not only proofs with a dleq are considered for sending I would prefer a name like `requireDleq`. What do you think?

src/CashuWallet.ts
```diff
@@ -327,15 +347,30 @@ class CashuWallet {
...
 			options?.proofsWeHave?.push(...keepProofsSelect);
 
-			const { keep, send } = await this.swap(amount, sendProofs, options);
-			const keepProofs = keepProofsSelect.concat(keep);
-			return { keep: keepProofs, send };
+			let { keep, send } = await this.swap(amount, sendProofs, options);
+			keep = keepProofsSelect.concat(keep);
+
+			// strip dleq if explicitly told so
+			if (!options?.includeDleq) {
```
Egge21M, line 354: `!options?.includeDleq` also evaluates to `true`i when it is undefined. So this is not explicit `false`

src/CashuWallet.ts
```diff
@@ -327,15 +347,30 @@ class CashuWallet {
 			);
 			options?.proofsWeHave?.push(...keepProofsSelect);
 
-			const { keep, send } = await this.swap(amount, sendProofs, options);
-			const keepProofs = keepProofsSelect.concat(keep);
-			return { keep: keepProofs, send };
+			let { keep, send } = await this.swap(amount, sendProofs, options);
+			keep = keepProofsSelect.concat(keep);
+
+			// strip dleq if explicitly told so
```
Egge21M, line 353: does `!options?.includeDleq` have two meanings in this case? Ignoring proofs without dleqs when `true` and stripping dleqs when `false` ? Maybe we should not overload this option to be more clear.

src/CashuWallet.ts
```diff
@@ -737,6 +771,10 @@ class CashuWallet {
 				options.privkey
 			).map((p: NUT11Proof) => serializeProof(p));
 		}
+		// Strip DLEQs if any
+		proofsToSend = proofsToSend.map((p: Proof) => {
```
Egge21M, line 775: Do dleqs need to be removed or are you optimising for smaller payloads here?

src/model/BlindedSignature.ts
```diff
@@ -1,19 +1,35 @@
...
+import { bytesToHex } from '@noble/hashes/utils.js';
 
 class BlindedSignature {
 	id: string;
 	amount: number;
 	C_: ProjPointType<bigint>;
+	dleq?: DLEQ;
 
-	constructor(id: string, amount: number, C_: ProjPointType<bigint>) {
+	constructor(id: string, amount: number, C_: ProjPointType<bigint>, dleq: DLEQ) {
```
Egge21M, line 12: `dleq` should be an optional parameter in the constructor, no?

src/utils.ts
```diff
@@ -292,7 +321,15 @@ export function handleTokens(token: string): Token {
 					secret: p.s,
 					C: bytesToHex(p.c),
 					amount: p.a,
-					id: bytesToHex(t.i)
+					id: bytesToHex(t.i),
+					dleq:
```
Egge21M, line 325: We can avoid `dleq:undefined` being present on every token without dleqs by doing something like

```ts
proofs.push({
    secret: p.s,
    C: bytesToHex(p.c),
    amount: p.a,
    id: bytesToHex(t.i),
    ...(p.d && {
        dleq: {
	    e: bytesToHex(p.d.e),
	    s: bytesToHex(p.d.s),
	    r: bytesToHex(p.d.r)
	} as SerializedDLEQ
    })
});
```

What do you think?

src/CashuWallet.ts
```diff
@@ -967,15 +1010,71 @@ class CashuWallet {
...
+		});
+	}
+
+	/**
+	 * Checks that each proof in `token` has a valid DLEQ proof according to
+	 * keyset `keys`
+	 * @param token The token subject to the verification
+	 * @param keys The Mint's keyset to be used for verification
+	 */
+	private requireDLEQ(token: Token, keys: MintKeys) {
```
Egge21M, line 1053: I might be missing something here, but it seems like this function does not rely on any of the instance state. Therefore I would like to move it to `utils.ts` and export it to the public API. 

Functions that do not return anything, but instead throw are hard to work with IMHO. Would you not agree that it would be clearer to rename this to `hasValidDleq`, make it take a SINGLE proof, and make it return `true` is all pass and `false` if dleq are either missing or invalid? 

We could then use this util to build filters like `proofs.filter(p => hasValidDleq(p, keys)` or a util to replace `requireDleq` with a better best-case running time like 

```ts
if (proofs.some(p => !hasValidDelq(p, keys))) {
    throw new Error("Proofs contain invalid DLEQ")
}
```

src/CashuWallet.ts
lollerfirst: They need to be removed, otherwise you destroy your anonimity.

src/CashuWallet.ts
lollerfirst: Yea that's what I intended and what made most sense to me.

src/CashuWallet.ts
lollerfirst: You are right.

src/CashuWallet.ts
Egge21M: Okay I see. I think most people will assume that not-passing an optional flag is equal to setting the boolean to false. Diverging from that needs to be documented very clearly I believe. However I would love to get some implementers thoughts on this. @callebtc @gudnuf could you leave your 2 sats on this?

src/CashuWallet.ts
Egge21M: Oh duh! Yeah that makes sense, thank you!

src/CashuWallet.ts
lollerfirst: Doing this right now: should I let `hasValidDleq` throw in case there is no matching key from the keyset `proof.id` for a particular proof amount? Or should I just return false.

Egge21M: I think it makes sense to throw in this case. `false` would indicate an invalid DLEQ, while in this case it's more than just that.

src/utils.ts
lollerfirst: I tried using this elegant solution in the place of other ugly `dleq == undefined ? undefined : ...` check it out.

src/CashuWallet.ts
lollerfirst: added a test for this as well.

src/CashuWallet.ts
Egge21M: Could you maybe explain the usecase to me? WWhy would someone decide to strip DLEQs when sending to another wallet?

src/model/BlindedSignature.ts
```diff
@@ -1,19 +1,35 @@
...
+		this.dleq = dleq;
 	}
 
 	getSerializedBlindedSignature(): SerializedBlindedSignature {
-		return { id: this.id, amount: this.amount, C_: this.C_.toHex(true) };
+		return {
+			id: this.id,
+			amount: this.amount,
+			C_: this.C_.toHex(true),
+			dleq:
```
Egge21M, line 24: Can we use the spread syntax here too?

lollerfirst: Yes. Also I am responding here to your other comment about "why would someone strip the DLEQs when sending to Carol". One might try to get a smaller size cashuB token for example. In general if it's not required by Carol why include it?

Egge21M (approved):
LGTM!

PR #202, opened 2024-11-04
NUT-17: WebSocket updates

Egge21M:
# Fixes: none

## Description

This adds WebSocket management, as well as a couple of methods on CashuWallet for full NUT-17 support

## Changes

- Added WSConnection Class
- Handle automatic connects and reconnects
- Added Message Queue
- Added CashuWallet methods
- Added unit and integration tests

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -863,22 +863,38 @@ class CashuWallet {
 		return states;
 	}
 
-	async onQuotePaid(quoteId: string, callback: (payload: any) => any) {
-		await this.mint.connectWebSocket();
+	async onQuotePaid(
+		quoteId: string,
+		callback: (payload: any) => any,
+		errorCallback: (e: Error) => void
+	) {
+		try {
+			await this.mint.connectWebSocket();
+		} catch (e) {
+			console.log('caught in quote paid');
+			if (e instanceof Error) {
+				return errorCallback(e);
+			} else if (e) {
+				return errorCallback(new Error('Something went wrong'));
+			}
+		}
 		if (!this.mint.webSocketConnection) {
 			throw new Error('failed to establish WebSocket connection.');
 		}
+		const subCallback = (payload: any) => {
+			if (payload.paid) {
+				callback(payload);
+			}
+		};
 		const subId = this.mint.webSocketConnection.createSubscription(
 			{ kind: 'bolt11_mint_quote', filters: [quoteId] },
-			(payload) => {
-				console.log(payload);
-			},
-			(e) => {
-				throw new Error(e.message);
+			subCallback,
+			(e: Error) => {
+				errorCallback(e);
 			}
 		);
 		return () => {
-			this.mint.webSocketConnection?.cancelSubscription(subId, callback);
+			this.mint.webSocketConnection?.cancelSubscription(subId, subCallback);
 		};
 	}
 
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -884,61 +884,22 @@ class CashuWallet {
 		};
 	}
 
-	async onMintQuotePaid(
-		quoteId: string,
-		callback: (payload: MintQuoteResponse) => void,
+	async onMeltQuoteUpdates(
+		quoteIds: Array<string>,
+		callback: (payload: MeltQuoteResponse) => void,
 		errorCallback: (e: Error) => void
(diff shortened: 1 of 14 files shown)


src/CashuWallet.ts
```diff
@@ -863,6 +867,153 @@ class CashuWallet {
...
+				if (p.state === MintQuoteState.PAID) {
+					callback(p);
+				}
+			},
+			errorCallback
+		);
+		return () => {
+			this.mint.webSocketConnection?.cancelSubscription(subId, callback);
+		};
+	}
```
callebtc, line 954: Can this not use `onMintQuoteUpdates`? Seems like a lot of repeated boiler plate.

Egge21M: Very good point! Done

src/WSConnection.ts
```diff
@@ -0,0 +1,184 @@
...
+	JsonRpcReqParams,
+	RpcSubId
+} from './model/types';
+import { OnOpenError, OnOpenSuccess } from './model/types/wallet/websocket';
+import { getWebSocketImpl } from './ws';
+
+export class WSConnection {
+	public readonly url: URL;
+	private readonly _WS: typeof WebSocket;
+	private ws: WebSocket | undefined;
```
alexlwn123, line 15: Should these private fields be prefixed by `_`? like the classes in utils?

src/WSConnection.ts
```diff
@@ -0,0 +1,184 @@
...
+	private connectionPromise: Promise<void> | undefined;
+	private subListeners: { [subId: string]: Array<(payload: any) => any> } = {};
+	private rpcListeners: { [rpsSubId: string]: any } = {};
```
alexlwn123, line 18: What's the purpose of having both subListeners and rpcListeners?

src/WSConnection.ts
```diff
@@ -0,0 +1,184 @@
...
+						return;
+					}
+					if (this.subListeners[subId]?.length > 0) {
+						const notification = parsed as JsonRpcNotification;
+						this.subListeners[subId].forEach((cb) => cb(notification.params.payload));
+					}
+				}
+			}
+		} catch (e) {
+			console.log(e);
```
alexlwn123, line 141: Should this try to call the errorCallback?

src/model/types/index.ts
```diff
@@ -17,3 +17,48 @@ export type InvoiceData = {
...
+type JsonRpcParams = any;
+
+export type JsonRpcReqParams = {
+	kind: RpcSubKinds;
+	filters: Array<string>;
+	subId: string;
+};
+
+type JsonRpcSuccess<T = any> = {
+	jsonrpc: '2.0';
```
alexlwn123, line 34: maybe make jsonrpc `readonly`, to make sure the editor doesn't display the type as `string`

gandlafbtc (commented):
There is a lot to this PR, I might need some more time to go through it in more detail, but I also don't wanna bike-shed it. On the surface it LGTM. So I'd say after fixing these minor issues, we put it in an RC.

src/WSConnection.ts
```diff
@@ -0,0 +1,205 @@
...
+
+	closeSubscription(subId: string) {
+		this.ws?.send(JSON.stringify(['CLOSE', subId]));
+	}
+
+	addSubListener(subId: string, callback: (payload: any) => any) {
+		(this.subListeners[subId] = this.subListeners[subId] || []).push(callback);
+	}
+
+	//TODO: Move to RPCManagerClass
```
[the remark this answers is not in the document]
Egge21M: Yeah, i wanted to modularise this further to help with testing. It's not necessary for functionality, but only there for clean-code purposes. IMO it's fine to leave this here for someone to pick up at some point

src/WSConnection.ts
```diff
@@ -0,0 +1,205 @@
...
+						return;
+					}
+					if (this.subListeners[subId]?.length > 0) {
+						const notification = parsed as JsonRpcNotification;
+						this.subListeners[subId].forEach((cb) => cb(notification.params.payload));
+					}
+				}
+			}
+		} catch (e) {
+			console.log(e);
```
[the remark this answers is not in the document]
Egge21M: Good point, will update. I don't think we should actually log anything to the console, unless the API consumer opts into it. I will open an issue to implement a "verbose" flag or similar

PR #204, opened 2024-11-13
mintProofs returns Proofs instead of object

Egge21M:
## Description

`CashuWallet.mintProofs` used to return an object with a single key `proofs`. This PR adjusts the function to return proofs directly

## Changes

- Make `mintProofs` return promised-wrapped `Array<Proof>` instead of `{proofs: Array<Proof>}`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


gudnuf (approved):
Nice! I was wondering about this the other day. lgtm

PR #207, opened 2024-11-16
raw token converters

gandlafbtc:
# Fixes: exposing functions for raw token format

## Description

Some transport methods (like nfc) require the lowest amount of data overhead possible. This PR addresses that by externalizing the raw token format (CBOR)

## Changes

- Extracted template converter functions
- added utility function to convert token to and from raw format
- expose utility functions on index.ts

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


src/utils.ts
```diff
@@ -521,3 +531,16 @@ export function hasValidDleq(proof: Proof, keyset: MintKeys): boolean {
 
 	return true;
 }
+
+export function tokenToRawToken(token: string | Token): Uint8Array {
```
Egge21M, line 535: This feels like bait :P 
I think we should come up with a proper name for this. We call the the serialised version TokenV4 and the object notation TokenV4Template, so maybe this can be TokenV4Binary?

src/utils.ts
```diff
@@ -521,3 +531,16 @@ export function hasValidDleq(proof: Proof, keyset: MintKeys): boolean {
...
+	if (typeof token === 'string') {
```
Egge21M, line 536: I would prefer this to only accept token and `outsource` conversion to the implementer. I think the readability improves if functions only do one (or close to one) thing.

Egge21M (commented):
Awesome! I don't think this is specced yet, is it? Blazingly fast! I dont mind getting ahead of the spec, however we need to keep an eye out for any changes.

src/utils.ts
gandlafbtc: Yeah that makes sense to me. I will change it accordingly

Egge21M:
@gandlafbtc is there a spec for this? It seems to me like not having any versioning might be a mistake here.

callebtc:
> @gandlafbtc is there a spec for this? It seems to me like not having any versioning might be a mistake here.

No spec for this afaik but it should be mentioned in NUT-00 that the CBOR structure can be exported without base64 when binary transport is possible. We do however need to come up with a magic byte prefix and a version like "cashuB" if we do that.

gandlafbtc:
yeah, the structure is the "version 4" structure. but I agree with what @callebtc said. maybe just a version field in the CBOR structure?

Egge21M:
> yeah, the structure is the "version 4" structure. but I agree with what @callebtc said. maybe just a version field in the CBOR structure?

Take a look at this: https://github.com/cashubtc/nuts/pull/199

prusnak:
> ake a look at this: [cashubtc/nuts#199](https://github.com/cashubtc/nuts/pull/199)

Yes, let's agree on binary format in described https://github.com/cashubtc/nuts/pull/199 first, merge that PR and only afterwards merge this PR

Egge21M:
I have adjusted this to match the newly merged binary token spec. Tests are still missing though

prusnak:
The code only supports V4 Binary tokens. Should we add the V3 support or should we remove V3 binary tokens from the spec? I am in favor for the latter, but no strong opinions here.

Egge21M:
> The code only supports V4 Binary tokens. Should we add the V3 binary tokens support to cashu-ts or should we remove V3 binary tokens from the spec? I am in favor for the latter, but no strong opinions here.

Good point... Adding them would be easy, but adding them to the spec in the first place was probably a mistake.

Egge21M:
I removed the reference to V4 completely now. I also adjusted the tests and exports. This is good to go i'd say

PR #208, opened 2024-11-16
add multi part support for melt

gandlafbtc:
# Fixes: nut-15 support

## Description

Adds simple stateless nut-15 support

## Changes

- Added options to method signature and and type

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -721,7 +721,7 @@ class CashuWallet {
 			request: invoice
 		};
 		if (options?.mpp) {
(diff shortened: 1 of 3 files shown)


Egge21M:
I fixed a small issue with the types and added an integration test. However it seems like Fakenut Nutshell, does not mocking MPP (cc @callebtc)?

gandlafbtc:
only LND backend seems to support it, at least that's the error message I am getting

PR #209, opened 2024-11-16
Added MintInfo Class

Egge21M:
## Description

Adds a MintInfo class that exposes some helper methods to easily check if a mint supports a NUT.

## Changes

- Added MintInfo Class
- Replaced GetInfoResponse with MintInfo

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/MintInfo.ts
+++ b/src/model/MintInfo.ts
@@ -1,10 +1,10 @@
 import { GetInfoResponse, MPPMethod, SwapMethod, WebSocketSupport } from './types';
 
 export class MintInfo {
-	private readonly mintInfo: GetInfoResponse;
+	private readonly _mintInfo: GetInfoResponse;
 
 	constructor(info: GetInfoResponse) {
-		this.mintInfo = info;
+		this._mintInfo = info;
 	}
 
 	isSupported(num: 4 | 5): { disabled: boolean; params: Array<SwapMethod> };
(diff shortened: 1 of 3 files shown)


src/model/MintInfo.ts
```diff
@@ -0,0 +1,90 @@
...
+	}
+	private checkMintMelt(num: 4 | 5) {
+		const mintMeltInfo = this.mintInfo.nuts[num];
+		if (mintMeltInfo && mintMeltInfo.methods.length > 0 && !mintMeltInfo.disabled) {
+			return { disabled: false, params: mintMeltInfo.methods };
+		}
+		return { disabled: true, params: mintMeltInfo.methods };
+	}
+	private checkNut17() {
+		if (this.mintInfo.nuts['17'] && this.mintInfo.nuts[17].supported.length > 0) {
```
gudnuf, line 51: Can all the keys just be numbers instead of nuts['17'] and nuts[17]? Also in `checkNut15`

Egge21M: Yes, I will change all the accessors to numbers. However the type will keep the strings, as property names can not be numbers. Accessing using numbers only works because JS will coerce to strings automatically

Egge21M: Or maybe I am misunderstanding, because I was referring to using `nuts[15]` instead of `nuts['15']`. Can you give an example?

gudnuf: I think you got my point. I was referring to the accessors because in some places you used a string and some you used a number to access the value. I don't care if its string or number, but should be consistent by using one or the other.

src/model/MintInfo.ts
```diff
@@ -0,0 +1,90 @@
+import { GetInfoResponse, MPPMethod, SwapMethod, WebSocketSupport } from './types';
+
+export class MintInfo {
+	private readonly mintInfo: GetInfoResponse;
```
gudnuf, line 4: should we use `_mintInfo` to match other private fields?

gudnuf (approved):
Looks great. Took me a second to realize what you were doing with the switch statement because I've never seen them used that way before, but now that I get it I think its a clean solution

src/model/MintInfo.ts
```diff
@@ -0,0 +1,90 @@
...
+	constructor(info: GetInfoResponse) {
+		this.mintInfo = info;
+	}
+
+	isSupported(num: 4 | 5): { disabled: boolean; params: Array<SwapMethod> };
+	isSupported(num: 7 | 8 | 9 | 10 | 11 | 12 | 14): { supported: boolean };
+	isSupported(num: 17): { supported: boolean; params?: Array<WebSocketSupport> };
+	isSupported(num: 15): { supported: boolean; params?: Array<MPPMethod> };
+	isSupported(num: number) {
+		switch (num) {
```
lollerfirst, line 15: Should the cases be handled exhaustively so that unrecognized NUT numbers always resolve to false?
Or alternatively throw on unrecognized NUT number.

Egge21M: Very good point! I added a default case that throws. Throwing in this case makes more sense to me, because it makes clear that this NUT is unsupported by the client, not the mint

callebtc: In [nutshell](https://github.com/cashubtc/nutshell/blob/901b16713cd962f613367acd0c77ffa46c42ba2d/cashu/wallet/mint_info.py#L25) there is a general function to check if a mint supports a NUT number, maybe this would be also good here.

Egge21M: I have now added a getter for the `nuts` object. That way, people can check whatever they want on the `nuts` object and use the `isSupported` method for all nuts that are implemented in cashu-ts

gandlafbtc (approved):
LGTM, Great addition!

Issue #210, opened 2024-11-17
Create types for `Options`  


gandlafbtc:
Functions on `CashuWallet` can take optional parameters. The list of optional parameters has grown over time, and it would be nice to create types for each of the Options. 

eg.:

```typescript
type SendOptions = {
			outputAmounts?: OutputAmounts;
			proofsWeHave?: Array<Proof>;
			counter?: number;
			pubkey?: string;
			privkey?: string;
			keysetId?: string;
			offline?: boolean;
			includeFees?: boolean;
			includeDleq?: boolean;
}
```

PR #212, opened 2024-11-17
use sendOption type in function argument

mubarak23:
# Fixes: #210 

## Description

Some Functions in `CashuWallet` take optional paramaters, they have grown overtime, this PR move the option paramaters into a type and use the type in the function argument instead of the long object with different type,

this way we can add more optional parameters in the type and still used there ever they are need in a function 
...

## Changes

- Create SendOption Type with different option parameters
- Pass SendOption as option value in function that require it. 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -11,6 +11,7 @@ import { DLEQ, type Proof as NUT11Proof } from '@cashu/crypto/modules/common/ind
 import { bytesToHex, hexToBytes, randomBytes } from '@noble/hashes/utils';
 import { CashuMint } from './CashuMint.js';
 import { BlindedMessage } from './model/BlindedMessage.js';
+import { MintInfo } from './model/MintInfo.js';
 import {
 	BlindingData,
 	GetInfoResponse,
@@ -52,18 +53,6 @@ import {
 	stripDleq,
 	sumProofs
 } from './utils.js';
-import { hashToCurve, pointFromHex } from '@cashu/crypto/modules/common';
-import {
-	blindMessage,
-	constructProofFromPromise,
-	serializeProof
-} from '@cashu/crypto/modules/client';
-import { deriveBlindingFactor, deriveSecret } from '@cashu/crypto/modules/client/NUT09';
-import { createP2PKsecret, getSignedProofs } from '@cashu/crypto/modules/client/NUT11';
-import { type Proof as NUT11Proof, DLEQ } from '@cashu/crypto/modules/common/index';
-import { SubscriptionCanceller } from './model/types/wallet/websocket.js';
-import { verifyDLEQProof_reblind } from '@cashu/crypto/modules/client/NUT12';
-import { MintInfo } from './model/MintInfo.js';
 
 /**
  * The default number of proofs per denomination to keep in a wallet.
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -246,21 +246,24 @@ class CashuWallet {
 	/**
 	 * Receive an encoded or raw Cashu token (only supports single tokens. It will only process the first token in the token array)
 	 * @param {(string|Token)} token - Cashu token, either as string or decoded
-	 * @param options.keysetId? override the keysetId derived from the current mintKeys with a custom one. This should be a keyset that was fetched from the `/keysets` endpoint
-	 * @param options.outputAmounts? optionally specify the output's amounts to keep and to send.
-	 * @param options.proofsWeHave? optionally provide all currently stored proofs of this mint. Cashu-ts will use them to derive the optimal output amounts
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -521,3 +521,21 @@ export function hasValidDleq(proof: Proof, keyset: MintKeys): boolean {
 
 	return true;
 }
+
+type OutputAmounts = {
```
Egge21M, line 525: Type declarations should go into the `types` directory

src/CashuWallet.ts
```diff
@@ -254,18 +255,7 @@ class CashuWallet {
...
-			keysetId?: string;
-			outputAmounts?: OutputAmounts;
-			proofsWeHave?: Array<Proof>;
-			counter?: number;
-			pubkey?: string;
-			privkey?: string;
-			requireDleq?: boolean;
-		}
-	): Promise<Array<Proof>> {
+	async receive(token: string | Token, options?: SendOptions): Promise<Array<Proof>> {
```
Egge21M, line 258: I don't think that we should have one big merged type for the options of different methods. It does avoid repeating code in types, however it is now no longer clear what options are available on a method. For example the type here suggests that `options.includeFees` would be a valid options for the `receive` method, but it is actually not.

src/CashuWallet.ts
```diff
@@ -1,51 +1,52 @@
...
+import { deriveBlindingFactor, deriveSecret } from '@cashu/crypto/modules/client/NUT09';
+import { createP2PKsecret, getSignedProofs } from '@cashu/crypto/modules/client/NUT11';
+import { verifyDLEQProof_reblind } from '@cashu/crypto/modules/client/NUT12';
+import { hashToCurve, pointFromHex } from '@cashu/crypto/modules/common';
+import { DLEQ, type Proof as NUT11Proof } from '@cashu/crypto/modules/common/index';
 import { bytesToHex, hexToBytes, randomBytes } from '@noble/hashes/utils';
 import { CashuMint } from './CashuMint.js';
 import { BlindedMessage } from './model/BlindedMessage.js';
 import {
+	BlindingData,
```
Egge21M, line 15: Good call on sorting imports alphabetically. Is this something done by your prettier? We should consider adding this to the projects prettierrc.

Egge21M (changes requested):
Thanks so much for taking this on! I left a couple of comments :)

src/CashuWallet.ts
mubarak23: so we should have an object destructuring that pull only the field that function need,

Egge21M: No, the problem is that the type signature of `receive` now looks like `receive(token: string | Token, options?: SendOptions): Promise<Array<Proof>>`. This makes someone believe that this function would accept something like `options.includeFees` because `includeFees` is on type `SendOptions`. However this function does not accept that option.

The type for `options` on `receive` should not include any option that is not supported. If you want to solve this with a single type, please use `Pick`, but TBH I believe having separate option types would be better.

mubarak23: so each function should have it's own separate type options.

gandlafbtc: yes i think it's best to have separate types for each function

gandlafbtc (approved):
LGTM

src/CashuWallet.ts
```diff
@@ -1,39 +1,56 @@
+import {
+	blindMessage,
+	constructProofFromPromise,
+	serializeProof
+} from '@cashu/crypto/modules/client';
+import { deriveBlindingFactor, deriveSecret } from '@cashu/crypto/modules/client/NUT09';
+import { createP2PKsecret, getSignedProofs } from '@cashu/crypto/modules/client/NUT11';
+import { verifyDLEQProof_reblind } from '@cashu/crypto/modules/client/NUT12';
+import { hashToCurve, pointFromHex } from '@cashu/crypto/modules/common';
+import { DLEQ, type Proof as NUT11Proof } from '@cashu/crypto/modules/common/index';
```
callebtc, line 10: Why are there so many new imports in this file?

mubarak23: these import already exist, my changes rearrange them in an alphabetical order, due to the prettier extension am using in my vscode

src/CashuWallet.ts
```diff
@@ -1,39 +1,56 @@
+import {
+	blindMessage,
+	constructProofFromPromise,
+	serializeProof
+} from '@cashu/crypto/modules/client';
+import { deriveBlindingFactor, deriveSecret } from '@cashu/crypto/modules/client/NUT09';
```
Egge21M, line 6: Can you please double check the imports? It seems like there is a bunch of duplication going on. For example is this line duplicated in line 61

Egge21M (changes requested):
Seems like there is some duplication going on with the imports.

PR #214, opened 2024-11-27
Use vite for bundling

Egge21M:
## Description

This PR adds `vite` for bundling and transpiling this library. Vite has become an industry standard, using rollup and esbuild under the hood. tsc is still kept, but only used for type checking and declarations. It also adds a standalone build that is made availabel through jsdelivr.

The PR also changes the build targets to ES2020 (from lowest ES5) and changes some of the export structure. While technically a breaking change, this should not affect downstream, as exports / imports are usually handled by the package manager / bundler.

## Changes

- Added vite and config
- Changed build commands to use vite
- Removed es5 target
- Added standalone build

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,6 @@
 
 	"rules": {
 		"@typescript-eslint/no-unused-vars": "error",
-		"@typescript-eslint/consistent-type-definitions": ["warn", "type"],
 		"@typescript-eslint/array-type": ["error", { "default": "generic" }],
 		"require-await": "off",
 		"@typescript-eslint/require-await": "error",
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,14 @@
 /test
 /coverage
 /.github
+.eslintignore
+.eslintrc.json
+.prettierignore
+.prettierrc
+vite.config.ts
+vitest.workspace.ts
 tsconfig.json
 src
+examples
 /lib/cashu-ts.iife.js
 /lib/cashu-ts.iife.js.map
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -13,7 +13,6 @@ import { CashuMint } from './CashuMint.js';
 import { BlindedMessage } from './model/BlindedMessage.js';
 import { MintInfo } from './model/MintInfo.js';
 import {
-	BlindingData,
 	GetInfoResponse,
 	MeltProofOptions,
 	MeltQuoteState,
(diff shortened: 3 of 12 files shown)


gudnuf:
have you looked at [vite library mode](https://vite.dev/guide/build#library-mode). Its probably overkill, but would allow us to create a demo wallet alongside the lib
`/src` would be the cashu-ts library
`/demo` would be a demo wallet that imports directly from `/src`

Egge21M:
> have you looked at [vite library mode](https://vite.dev/guide/build#library-mode). Its probably overkill, but would allow us to create a demo wallet alongside the lib `/src` would be the cashu-ts library `/demo` would be a demo wallet that imports directly from `/src`

Yes, good idea. This PR actually introduces library mode. I think all the is required to make a dev env work, would be to add a index.html file and reference the lib's index file in there.

Building a simple wallet interface or at least some buttons to test the public API against a local test mint is a good idea IMHO. I also do not think its overkill. It also allows implementers to use their browser as a sort of scratch pad for testing without having to write a test file.

However I think this should be introduced in a separate PR. Where you able to include this PRs output in your projects without any bundling issues @gudnuf ?

gudnuf:
> Where you able to include this PRs output in your projects without any bundling issues @gudnuf ?

Yea works for me now!

src/model/BlindedSignature.ts
```diff
@@ -1,7 +1,7 @@
 import { ProjPointType } from '@noble/curves/abstract/weierstrass';
 import { SerializedBlindedSignature } from './types/index.js';
 import { DLEQ } from '@cashu/crypto/modules/common';
-import { bytesToHex } from '@noble/hashes/utils.js';
+import { bytesToHex } from '@noble/hashes/utils';
```
gudnuf, line 4: Does this allow us to remove .js from every import? Not sure if the file extensions were included previously because it was required for the bundling to work, but shouldn't be necessary anymore with vite I think

gudnuf (approved):

src/model/BlindedSignature.ts
gandlafbtc: Would be worth a try

Egge21M: Yes, that is correct. TSC requires file extensions, while vite will resolve imports without them. We can clean imports in a separate PR, also adding path aliases for cleaner imports (eg `@/utils` instead of `../../utils`

PR #218, opened 2024-11-30
Restructured blinding + custom messages + keep vectors 

Egge21M:
## Description

This PR addresses a couple of things that were discussed in the past: Sorted outputs when swapping, as well as custom outputs. I always felt like the old `BlindingData` type was not ergonomic enough, so I restructured some of the blinding types and logic, moved them to their own Class and Interface. These changes are so closely related, that they became a single PR (that got a little out of hand, sorry...).

## Changes

- BlindingData is now called OutputData. It holds all the data required to construct an Output from a BlindedMessage and a Signature
- OutputData (form. BlindingData) is no longer an Array, but now represents a single encapsulated future Proof
- Wallet methods now accept a new option `outputData` that implements the OutputDataLike Interface. Using this consumers of the library can opt out of the automatic creation of outputs and supply their own logic.
- createSwapPayload now sorts the requested outputs by amount. It returns an additional keepVector (Array<boolean>) that can be used to determine which outputs are meant to be kept. It also returns a sortedIndices array that can be used to reorder the sorted outputs in their original order.
- Removed some of the old blinding code. Was internal code, so no breaking changes.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


Egge21M:
Converted back to draft because I came up with the idea to add factories and this needs some refactoring

Egge21M:
Edited the original PR message to reflect new changes. Review required and much appreciated!

src/model/OutputData.ts
```diff
@@ -0,0 +1,169 @@
...
+		counter: number,
+		keyset: MintKeys,
+		customSplit?: Array<number>
+	): Array<OutputData> {
+		const amounts = splitAmount(amount, keyset.keys, customSplit);
+		const data: Array<OutputData> = [];
+		for (let i = 0; i < amounts.length; i++) {
+			data.push(this.createSingleDeterministicData(amount, seed, counter + i, keyset.id));
+		}
+		return data;
```
mjkeaton, line 151:
```suggestion
		return amounts.map((_, i) =>
			this.createSingleDeterministicData(amount, seed, counter + i, keyset.id)
		);
```

or should it be
```
return amounts.map((a, i) =>
	this.createSingleDeterministicData(a, seed, counter + i, keyset.id)
);
```
?

Egge21M: Yes, good point. Fixed

test/integration.test.ts
```diff
@@ -34,6 +38,13 @@ const unit = 'sat';
 
 injectWebSocketImpl(ws);
 
+function expectProofsSecretToEqual(p: Array<Proof>, s: string) {
```
gudnuf, line 41: should this be renamed to be more clear that it is comparing parsedSecret's data? At first I thought this was comparing the raw secrets.
maybe `expectNUT10SecretDataToEqual`

test/integration.test.ts
```diff
@@ -474,3 +484,153 @@ describe('dleq', () => {
...
+		await new Promise((res) => setTimeout(res, 1000));
+		const testOutputAmounts = [8, 4, 8, 2, 8, 2];
+		const testProofs = await wallet.mintProofs(64, mintQuote.quote);
+
+		const fees = wallet.getFeesForProofs(testProofs);
+
+		const { send } = await wallet.send(32, testProofs, {
+			outputAmounts: {
+				sendAmounts: testOutputAmounts,
+				keepAmounts: [16, 8, ...Array(8 - fees).fill(1)]
```
gudnuf, line 631: can you explain the logic behind this keepAmounts. Is this something the users of cashu-ts will have to do whenever there are fees and they want change?

test/integration.test.ts
```diff
@@ -474,3 +484,153 @@ describe('dleq', () => {
...
+		const { keep, send } = await wallet.send(20, proofs, {
+			outputData: { send: [...data1, ...data2] }
+		});
+		const key1Sends = send.slice(0, data1.length);
+		const key2Sends = send.slice(data1.length);
+		expectProofsSecretToEqual(key1Sends, 'key1');
+		expectProofsSecretToEqual(key2Sends, 'key2');
+	});
+});
+describe('Keep Vector and Reordering', () => {
```
gudnuf, line 600: it doesn't seem like you are testing the "Keep Vector" here, just that `send` is properly created when `keepAmounts` is provided

src/model/OutputData.ts
```diff
@@ -0,0 +1,163 @@
...
+			dleq: dleq
+		};
+		const A = pointFromHex(keyset.keys[this.blindedMessage.amount]);
+		const proof = constructProofFromPromise(blindSignature, this.blindingFactor, this.secret, A);
+		const serializedProof = {
+			...serializeProof(proof),
+			...(dleq && {
+				dleqValid: verifyDLEQProof_reblind(this.secret, dleq, proof.C, A)
+			}),
+			...(dleq && {
```
gudnuf, line 69: can `dleqValid` and `dleq` be moved into the same `...(deli && {` or are those purposely separate?

src/model/OutputData.ts
```diff
@@ -0,0 +1,163 @@
...
+				dleq: {
+					s: bytesToHex(dleq.s),
+					e: bytesToHex(dleq.e),
+					r: numberToHexPadded64(dleq.r ?? BigInt(0))
+				} as SerializedDLEQ
+			})
+		} as Proof;
```
gudnuf, line 76: Instead of casting to `Proof` can we say `const serializedProof: Proof = {...`? With the way it is now, I can remove `serializeProof` and not get any type errors

src/CashuWallet.ts
```diff
@@ -98,6 +105,7 @@ class CashuWallet {
 			mintInfo?: GetInfoResponse;
 			bip39seed?: Uint8Array;
 			denominationTarget?: number;
+			keepFactory?: OutputDataFactory;
```
gudnuf, line 108: should add `keepFactory` to the jsdoc

src/CashuWallet.ts
```diff
@@ -750,34 +791,67 @@ class CashuWallet {
...
 		pubkey?: string,
-		privkey?: string
-	): {
-		payload: SwapPayload;
-		blindingData: BlindingData;
-	} {
+		privkey?: string,
+		customOutputData?: {
+			keep?: Array<OutputDataLike> | OutputDataFactory;
+			send?: Array<OutputDataLike> | OutputDataFactory;
```
gudnuf, line 797: should there be a type for `Array<OutputDataLike> | OutputDataFactory;`? Seems to be repeated a lot, but maybe its more clear to write it out each time

src/CashuWallet.ts
```diff
@@ -1006,43 +1066,38 @@ class CashuWallet {
...
-				deterministicR = bytesToNumber(deriveBlindingFactor(this._seed, keysetId, counter + i));
-			} else {
-				secretBytes = randomBytes(32);
-			}
-			if (!pubkey) {
-				const secretHex = bytesToHex(secretBytes);
-				secretBytes = new TextEncoder().encode(secretHex);
+		pubkey?: string,
+		outputAmounts?: Array<number>,
+		p2pk?: { pubkey: string; locktime?: number; refundKeys?: Array<string> },
```
gudnuf, line 1075: are you keeping `pubkey` for backwards compatibility? Can we remove it to replace with `p2pk`?

src/CashuWallet.ts
```diff
@@ -1006,43 +1066,38 @@ class CashuWallet {
...
-			blindedMessages.push(blindedMessage.getSerializedBlindedMessage());
+			outputData = OutputData.createDeterministicData(
+				amount,
+				this._seed,
+				counter,
+				keyset,
+				outputAmounts
+			);
+		} else if (p2pk) {
+			outputData = OutputData.createP2PKData(p2pk, amount, keyset);
```
gudnuf, line 1093: did you mean to not include `outputAmounts`

src/CashuWallet.ts
```diff
@@ -270,23 +282,30 @@ class CashuWallet {
 			}
 		}
 		const amount = sumProofs(token.proofs) - this.getFeesForProofs(token.proofs);
-		const { payload, blindingData } = this.createSwapPayload(
+		let newOutputData: { send: Array<OutputDataLike> | OutputDataFactory } | undefined = undefined;
+		if (outputData) {
+			newOutputData = { send: outputData };
+		} else if (this._keepFactory) {
+			newOutputData = { send: this._keepFactory };
```
gudnuf, line 289: why are you using the keepFactory for `send`?

gudnuf (commented):
Looks great! Just some minor nits and questions. 

I don't fully understand the `keepFactory` and when it used. Can you elaborate a bit more on the logic behind that? My assumption is that this is used as a way to default how the user wants to store their proofs, but I think my confusion is related to my [comment here](https://github.com/cashubtc/cashu-ts/pull/218#discussion_r1899184183)

test/integration.test.ts
Egge21M: When you use `outputAmounts` you are required to specify amounts that match the keyset and also satisfy the amount constraints exactly. So yes, if your mint takes fees and you want to specify exact output amounts, you need to consider fees.

test/integration.test.ts
Egge21M: It does test the vector, but it is hidden behind the scenes. Any outputs in swap-payload ssend to the mint are now ordered by amount, to increase privacy. By testing that the output of `receive` is in the same order as the keepAmounts specified, we proof that the reordering of proofs works as expected. The code that does this can be found in CashuWallet.ts#305 (reordering) CashuWallet.ts#872 (sorting)

src/model/OutputData.ts
Egge21M: Good point. I copied this logic from the old part of the codebase that created proofs. @lollerfirst am I missing anything here or could we merge these two checks like

```ts
const serializedProof = {
    ...serializeProof(proof),
    ...(dleq && {
        dleqValid: verifyDLEQProof_reblind(this.secret, dleq, proof.C, A),
            dleq: {
            s: bytesToHex(dleq.s),
	    e: bytesToHex(dleq.e),
	    r: numberToHexPadded64(dleq.r ?? BigInt(0))
	    } as SerializedDLEQ
	})
    } as Proof;
```

src/CashuWallet.ts
Egge21M: I decided to keep a union here, because IDEs will then show the union instead of the alias type. I think this is beneficial, because these two types are very different things that change the way the function behaves.

src/CashuWallet.ts
Egge21M: Yes exactly. `pubkey` should be removed with a future major update.

src/CashuWallet.ts
Egge21M: Good catch. Unfortunately there is a bug in the current version that requires this: https://github.com/cashubtc/cashu-ts/issues/224

Egge21M:
> Looks great! Just some minor nits and questions.
> 
> I don't fully understand the `keepFactory` and when it used. Can you elaborate a bit more on the logic behind that? My assumption is that this is used as a way to default how the user wants to store their proofs, but I think my confusion is related to my [comment here](https://github.com/cashubtc/cashu-ts/pull/218#discussion_r1899184183)

Thank you for taking the time to go through this!
The idea of the `keepFactory` is to provide a standard "fallback" factory to the CashuWallet instance that should be used whenever we create proofs that are supposed to be "kept" (e.g. mint, melt-change, send-change, receive, etc.).

Using this, you can create a factory that creates some sort of outputs (like deterministic ones) and make the library use is automatically. A simplistic example of a CashuWallet that creates auto-incrementing, deterministic secrets using a keepFactory can be seen here: https://njump.me/nevent1qqsvzxrfqyt5xk0s7f60mm9rlthkdtv70g3qsl8xykrgcvnfdrfpaecpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgqg0waehxw309ahx7um5wghx6mmdqgsdmup6e2z6mcpeue6z6kl08he49hcen5xnrc3tnpvw0mdgtjemh0slwxw3x

src/model/OutputData.ts
```diff
@@ -0,0 +1,163 @@
...
+	static createDeterministicData(
+		amount: number,
+		seed: Uint8Array,
+		counter: number,
+		keyset: MintKeys,
+		customSplit?: Array<number>
+	): Array<OutputData> {
+		const amounts = splitAmount(amount, keyset.keys, customSplit);
+		return amounts.map((_, i) =>
+			this.createSingleDeterministicData(amount, seed, (counter = i), keyset.id)
```
gudnuf, line 144: BUG! You're creating output data with the same amount.
Should be:
```ts
return amounts.map((amount, i) =>
...
```

gudnuf (approved):

src/model/OutputData.ts
lollerfirst: lgtm :+1:

lollerfirst (approved):

src/utils.ts
```diff
@@ -49,7 +49,7 @@ export function splitAmount(
 			);
 		}
 		split.forEach((amt: number) => {
-			if (!hasCorrespondingKey(amt, keyset)) {
+			if (!hasCorrespondingKey(amt, keyset) && amt !== 0) {
```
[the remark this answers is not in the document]
Egge21M: Check out https://github.com/cashubtc/cashu-ts/pull/240
I think this check was redundant. Removed the check in the refactor

PR #219, opened 2024-12-03
handle specific errors cases in request.ts

gudnuf:
# Fixes: error handling

Error handling was not specific enough and did not even get the error `code` from a `400` response

## Description

There are a variety of errors that can occur when making a request to a mint.

- fetch fail -> instanceof NetworkError
- fetch succeeds, but status is neither 2xx nor 400 -> instanceof HttpResponseError
- fetch succeeds and status is 400 -> instanceof MintOperationError with specific Cashu Error code as defined [in the nuts](https://github.com/cashubtc/nuts/blob/main/error_codes.md)

...

## Changes

- added NetworkError and MintOperationError classes
- catch fetch error and throw NetworkError
- If fetch succeeds but `!response.ok` (ie. http error), then throw MintOperationError if applicable otherwise HttpResponseError
- Export these new error classes so they can be used to handle the specific errors

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,6 @@
 import { CashuMint } from './CashuMint.js';
 import { CashuWallet } from './CashuWallet.js';
 import { PaymentRequest } from './model/PaymentRequest.js';
-import { OutputData } from './model/OutputData.js';
 import { setGlobalRequestOptions } from './request.js';
 import {
 	getEncodedToken,
@@ -19,7 +18,6 @@ export {
 	CashuMint,
 	CashuWallet,
 	PaymentRequest,
-	OutputData,
 	getDecodedToken,
 	getEncodedToken,
 	getEncodedTokenV4,
@@ -31,3 +29,5 @@ export {
 };
 
 export { injectWebSocketImpl } from './ws.js';
+
+export { MintOperationError, NetworkError, HttpResponseError } from './model/Errors.js';
--- a/src/model/Errors.ts
+++ b/src/model/Errors.ts
@@ -1,4 +1,4 @@
-/** This error is thrown when a HTTP response is not 2XX or 400. */
+/** This error is thrown when a HTTP response is not 2XX nor a protocol error. */
 export class HttpResponseError extends Error {
 	status: number;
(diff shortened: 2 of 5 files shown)


Egge21M:
Regarding the structure of the classes:
I think having a separate sub-class for each error type is overkill. These classes do not really do anything, and the only differentiator is the error code and detail. Instead we could have a single MintOperationError that consumes the `code` and `detail` and that can be checked against quite easily like:


```ts
catch(e: unknown) {
  if (e instance of MintOperationError) {
    switch(e.code) {
      case x: 
        // do smth.
        break;
      case y:
        // do smth else
        break;
    }
  }
}
```

gudnuf:
@Egge21M I made some changes. I kept the error messages for specific codes, but moved that into the MintOperationError constructor. I think this is useful in case the mint does not return a `detail` and then the client can have consistent error messages. And my tests pass now, thanks!!

Should I just remove the ApiError type? I don't really understand how it is being used. It looks like most of the responses have ApiError tagged on, but all the actual errors just thrown. Seems like we should get rid of ApiError and make the user catch and handle errors on their own.

gandlafbtc (approved):
LGTM,  maybe I'm missing the point, but my only question is if we should maintain a fallback error list inside the lib. These things tend to be outdated, and chances are if the mint follows the protocol, we get the correct messages anyway, and if not they presumably won't provide the correct codes which makes lookup impossible.

gudnuf:
>  my only question is if we should maintain a fallback error list inside the lib.

Agreed, I removed the fallback messages

I also just made it so that the `MintOperationError` now extends the `HttpResponseError` instead of just `Error` which makes it backwards compatible for anyone that is checking for the `HttpResponseError`

Egge21M:
LGTM! Thanks for picking this up again. @gandlafbtc I would be fine adding this as a "patch" to the latest version as per request. wby?

gandlafbtc:
Let's merge this and release a new minor version from development. there is a few other PRs I've merged that I would like to release (#217  #254)

Issue #225, opened 2024-12-12
Add MintInfo method for amountless


Egge21M:
See: https://github.com/cashubtc/nuts/pull/173

mubarak23:
@Egge21M is anybody working on this, if not, i will work on it.

Egge21M:
> @Egge21M is anybody working on this, if not, i will work on it.

Not that I know of, feel free to take this on. Thank you for your initiative!

mubarak23:
if i understand the [https://github.com/cashubtc/nuts/pull/173](https://github.com/cashubtc/nuts/pull/173) correctly, bunch of work to support amountless invoice will be handle by the mint, but here our `createMeltQuote` will accept an optional field, then check if the mint support amountless invoice, if YES, we refactor the `MeltQuotePayload` to  have this 

```
"options": { // Optional
    "amountless": {
      "amount_msat": <int>
    }

```

or better still, have a new method handle amountless melt quote seperate from `createMeltQuote`
@Egge21M am correct here

Egge21M:
> if i understand the [cashubtc/nuts#173](https://github.com/cashubtc/nuts/pull/173) correctly, bunch of work to support amountless invoice will be handle by the mint, but here our `createMeltQuote` will accept an optional field, then check if the mint support amountless invoice, if YES, we refactor the `MeltQuotePayload` to have this
> 
> ```
> "options": { // Optional
>     "amountless": {
>       "amount_msat": <int>
>     }
> ```
> 
> or better still, have a new method that handle amountless melt quote seperate from `createMeltQuote` @Egge21M am correct here

I think it is fine to have it in one method, as it does not require an additional parameter. However the MintInfo class should be extended by a method to check wether this is supported

PR #230, opened 2024-12-23
chore:add support for amountless in mintinfo and createMeltQuote

mubarak23:
# Fixes: #225 

## Description
PR base on [https://github.com/cashubtc/nuts/pull/173](https://github.com/cashubtc/nuts/pull/173) for handling amountless when melting quote

## Changes

- add check if mint support amountless 
- add options field for handling amountless in `MeltQuotePayload`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -662,11 +662,26 @@ class CashuWallet {
 	 * @param invoice LN invoice that needs to get a fee estimate
 	 * @returns the mint will create and return a melt quote for the invoice with an amount and fee reserve
 	 */
-	async createMeltQuote(invoice: string): Promise<MeltQuoteResponse> {
+	async createMeltQuote(invoice: string, amount?: number): Promise<MeltQuoteResponse> {
 		const meltQuotePayload: MeltQuotePayload = {
 			unit: this._unit,
 			request: invoice
 		};
+
+		if(this._mintInfo?.amountless){
+			const meltQuotePayload: MeltQuotePayload = {
+  		unit: this._unit,
+			request: invoice,
+			options: {
+				amountless: {
+					amount_msat: amount ?? 0,
+				},
+			},
+		};
+		const meltQuote = await this.mint.createMeltQuote(meltQuotePayload);
+		return meltQuote;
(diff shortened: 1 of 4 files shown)


src/model/MintInfo.ts
```diff
@@ -94,4 +94,7 @@ export class MintInfo {
 	get motd() {
 		return this._mintInfo.motd;
 	}
+	get amountless() {
+		return this._mintInfo.amountless;
```
Egge21M, line 98: The mint signals support via the MeltMethodOptions, not on the root level of the InfoReponse. Please check the PR again and make sure to adjust this accordingly.

src/CashuWallet.ts
```diff
@@ -662,11 +662,25 @@ class CashuWallet {
...
 	 * @returns the mint will create and return a melt quote for the invoice with an amount and fee reserve
 	 */
-	async createMeltQuote(invoice: string): Promise<MeltQuoteResponse> {
+	async createMeltQuote(invoice: string, amount?: number): Promise<MeltQuoteResponse> {
 		const meltQuotePayload: MeltQuotePayload = {
 			unit: this._unit,
 			request: invoice
 		};
+
+		if (this._mintInfo?.amountless) {
```
Egge21M, line 671: We should only be adding the amount, if the caller explicitly passes amount. Otherwise this implementation defaults to 0 and mints that support this will throw an error if amount != invoiceAmount.

Also amount passed needs to be in msats

Egge21M (changes requested):

src/model/MintInfo.ts
mubarak23: i can't seem to find `MeltMethodOptions` ,  is this something that already exist or we have to create a new one for it @Egge21M

Egge21M: It is there, but it might need some refactoring for this PR. You can find it as `GetInfoResponse.nuts.5.methods`

gandlafbtc:
what is the status on this PR @mubarak23 ?

PR #236, opened 2025-01-18
Fix deterministic secrets

gandlafbtc:
# Fixes: #[issue]

## Description

...

## Changes

- ...
- ...

## PR Tasks

- [x] Open PR
- [ ] run `npm run test` --> no failing unit tests
- [ ] run `npm run format`

PR #238, opened 2025-01-22
Fix deterministic secrets

Egge21M:
## Description

Replaces https://github.com/cashubtc/cashu-ts/pull/236, but is based of `blinding` branch.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -603,7 +603,7 @@ class CashuWallet {
 			throw new Error('CashuWallet must be initialized with a seed to use restore');
 		}
 		// create blank amounts for unknown restore amounts
-		const amounts = Array(count).fill(1);
+		const amounts = Array(count).fill(0);
 		const outputData = OutputData.createDeterministicData(
 			amounts.length,
 			this._seed,
@@ -616,9 +616,26 @@ class CashuWallet {
 			outputs: outputData.map((d) => d.blindedMessage)
 		});
 
-		const validData = outputData.filter((d) => outputs.some((o) => d.blindedMessage.B_ === o.B_));
+		const outputsWithSignatures: {signature: SerializedBlindedSignature, data: OutputData}[] = []
+
+		console.log(outputs.map(o=>o.amount))
+		console.log(promises.map(o=>o.amount))
+
+		for (let i = 0; i < outputs.length; i++) {
+			const data = outputData.find(d=> d.blindedMessage.B_===outputs[i].B_)
+			if (!data) {
+				continue
+			}
+			outputsWithSignatures[i] = {
+				signature: promises[i],
+				data
+			}
+		}
+		outputsWithSignatures.forEach(o=>o.data.blindedMessage.amount = o.signature.amount)
+
+		
 		return {
-			proofs: validData.map((d, i) => d.toProof(promises[i], keys))
+			proofs: outputsWithSignatures.map((d) => d.data.toProof(d.signature, keys))
 		};
 	}
 
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -616,24 +616,23 @@ class CashuWallet {
 			outputs: outputData.map((d) => d.blindedMessage)
 		});
 
-		const outputsWithSignatures: {signature: SerializedBlindedSignature, data: OutputData}[] = []
-
-		console.log(outputs.map(o=>o.amount))
-		console.log(promises.map(o=>o.amount))
+		const outputsWithSignatures: Array<{
+			signature: SerializedBlindedSignature;
+			data: OutputData;
+		}> = [];
 
 		for (let i = 0; i < outputs.length; i++) {
-			const data = outputData.find(d=> d.blindedMessage.B_===outputs[i].B_)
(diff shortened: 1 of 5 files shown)


Egge21M:
Not sure why format pipeline is failing. Running fine locally...

Egge21M:
@gandlafbtc there still was an issue with the encoding of secrets. I added the test vectors from the spec repo and fixes the encoding issue. Now OutputData is created in line with the test vectors and we should be good.

Egge21M:
> how it is now, it will generate gibberish instead of hex secrets. I think it's due to the fact that we encode general secrets as utf-8 (because they could be scripts) but we need to encode deterministic secrets as hex. 

Did you look at the tests in wallet.test? Vectors are passing

gandlafbtc:
Yeah I've seen the test vectors, but the fact is that if you use the lib like this, you will get unusable secrets (feel free to try it yourself). 

The issue is that proofs that have been created deterministically will end up being encoded wrong, probably because they go through the same pipeline as the regular secrets (that are being utf-8 encoded), and then end up being gibberish.

Egge21M:
> The issue is that proofs that have been created deterministically will end up being encoded wrong, probably because they go through the same pipeline as the regular secrets (that are being utf-8 encoded), and then end up being gibberish.

The issue is `CashuWallet.serializeProofs` assumes the secretBytes to be UTF-8 (which is not right for deterministic secrets i believe, but the spec is not explicit about this).

Egge21M:
This should finally be good to go now :D Sorry for the confusion

src/CashuWallet.ts
```diff
@@ -616,9 +616,25 @@ class CashuWallet {
...
-		const validData = outputData.filter((d) => outputs.some((o) => d.blindedMessage.B_ === o.B_));
+		const outputsWithSignatures: Array<{
+			signature: SerializedBlindedSignature;
+			data: OutputData;
+		}> = [];
+
+		for (let i = 0; i < outputs.length; i++) {
+			const data = outputData.find((d) => d.blindedMessage.B_ === outputs[i].B_);
+			if (!data) {
+				continue;
```
gudnuf, line 627: So this means the cashu-ts user gave a range to restore that includes some blindedMessages not yet signed by the mint? I wonder if it's fine to just ignore that?

src/utils.ts
```diff
@@ -49,7 +49,7 @@ export function splitAmount(
 			);
 		}
 		split.forEach((amt: number) => {
-			if (!hasCorrespondingKey(amt, keyset)) {
+			if (!hasCorrespondingKey(amt, keyset) && amt !== 0) {
```
gudnuf, line 52: are there keys that have a zero? Seems like we shouldn't check that `amt !== 0` because either the keyset does not have a zero value key and this will be false anyways or for some reason the keyset has a '0' and that should be allowed.

src/CashuWallet.ts
Egge21M: This can happen, when people increment the counter optimistically. For example if a mint errors, but the wallet does not roll back the counter, leaving a gap. However we could add a second return value here, that lets the caller know which counts did not result in a valid proof. WDYT?

cc @gandlafbtc

src/utils.ts
Egge21M: Yes, you are right, this seems redundant. @gandlafbtc are we missing something here? I think you added this

src/CashuWallet.ts
gandlafbtc: Yeah that makes sense to me. 

What exactly should be returned?

An array of the unsigned blinded messages,
or the index of the last signed proof
...

src/utils.ts
gandlafbtc: Yes, blank outputs are of value 0, but I'm not sure if it's any longer part of the spec

https://github.com/cashubtc/nuts/blob/main/08.md#mint-flow

At least that was the idea behind it, to changing undetermined values to 0, instead of 1. I'm adding a comment to the relevant line.

src/CashuWallet.ts
```diff
@@ -603,7 +603,7 @@ class CashuWallet {
 			throw new Error('CashuWallet must be initialized with a seed to use restore');
 		}
 		// create blank amounts for unknown restore amounts
-		const amounts = Array(count).fill(1);
+		const amounts = Array(count).fill(0);
```
gandlafbtc, line 606: We're using 0 for unknown/blank outputs

src/utils.ts
Egge21M: I don't think the spec is explicit about the amount a blank output should have. In the example of NUT-08 the blank outputs have an amount of 1.

gandlafbtc: You're right! We can change it to 1 then. Sorry about that, I believe the default used to be 0, but seems to no longer be the case.

src/CashuWallet.ts
gandlafbtc, line 606:
```suggestion
		const amounts = Array(count).fill(1);
```

src/utils.ts
gandlafbtc, line 52:
```suggestion
			if (!hasCorrespondingKey(amt, keyset)) {
```

PR #248, opened 2025-02-07
fix: add icon_url to GetInfoResponse

purrgrammer:
# Fixes: add icon_url to GetInfoResponse


gandlafbtc (commented):
LGTM.

I'll change the target to next version

PR #249, opened 2025-02-15
Fix NUT-15 Underspecified Amount

lollerfirst:
partial amount for multipath melt quotes is now specified as millisat (as in a millisat portion of the invoice)


gandlafbtc (approved):
Thanks! I think that makes sense to stay in line with the LN spec. 

LGTM

Issue #205, opened 2024-11-15
All logs should be conditional


Egge21M:
There is a couple of places where cashu-ts will log errors or warnings. This package is usually used a dependency and should not log to the console, without explicitly being told to do so. We should therefore make all logs opt-in, using a `verbose` flag or similar.

PR #253, opened 2025-02-25
Add opt-in logging with verbose flag

vnnkl:
# Fixes: #205 

## Description

This PR refactors the logging functionality in the cashu-ts library to make it opt-in rather than always logging to the console. This improves the developer experience by reducing console noise and giving users control over when logging occurs.

## Changes

- Added a `_verbose` flag to `CashuWallet`, `CashuMint`, and `WSConnection` classes
- Implemented private logging methods (`log`, `warn`, `error`) in these classes that only log when verbose mode is enabled
- Replaced all direct console logging calls with these conditional logging methods
- Added a global verbose flag in `utils.ts` with `setVerbose` function and `logWarning` function
- Added verbose flag to the `request.ts` module with `setRequestVerbose` function and `logError` function
- Updated the `CashuMint` constructor to set the request verbose flag
- Updated the `ConnectionManager` to pass verbose flag to `WSConnection` instances
- Added documentation in the README explaining the new verbose option
- Created comprehensive tests to verify that the verbose flag works correctly across all components

## PR Tasks

- [x] Open PR
- [ ] run `npm run format`


gandlafbtc (approved):
LGTM

Egge21M:
This is great! Thank you very much for the contribution. What do you think about moving the logging methods to an interface and letting the consumer import an implementation of the interface? I think this would be a great addition to your PR, as it:

- Reduces code duplication (no need to redefine the logging methods)
- Keeps the library classes lean (logging logic is encapsulated in its own class)
- Lets consumers inject their own logging and does not restrict them to the builtin console

Here is a quick example:

```ts
interface Logger {
  debug: (...data: any[]) => void;
  info: (...data: any[]) => void;
  // add more methods here
}

class CustomLogger implements Logger {
  debug(...data: any[]) { console.debug(...data); }
  info(...data: any[]) { console.info(...data); }
}

// CustomLogger can then be passed to CashuMint's contstructor

const mint = new CashuMint(url, undefined, new CustomLogger());

// CashuMint's methods could then call Logger methods

async doSomething() {
  if (someThing) {
    this._logger?.debug("Something happened");
}
```

PR #254, opened 2025-02-26
remove dleqValid from proof and expose verification api

gandlafbtc:
# Fixes: DLEQ validity should not be stored on proof

## Description

The `dleqValid` flag on the proof doesn't make sense, since it has to be verified by the receiver independently. So IMO the better approach is to just encode the dleq params, and instead expose the verification API so that the receiver can verify the proofs on receive

## Changes

- remove dleqValid field from proof
- expose API

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,7 +10,8 @@ import {
 	deriveKeysetId,
 	decodePaymentRequest,
 	getDecodedTokenBinary,
-	getEncodedTokenBinary
+	getEncodedTokenBinary,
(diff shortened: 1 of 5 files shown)


lollerfirst:
I thought this would be useful for wallet implementors since they could just iterate over all of the proofs and check the boolean flag instead of having to understand how the verification functions works. It's ACK for me anyway.

PR #263, opened 2025-03-09
use undefined for p2pk params on swap payload

gandlafbtc:
# Fixes: p2pk params broken

## Description

P2pk params are being set on the creation of the `keep` proofs, leading to unspendable proofs in the sender wallet, if a split occurs

## Changes

- changed p2pk params to undefined in `keep` output creation

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -943,9 +943,9 @@ class CashuWallet {
 				keepAmount,
 				keyset,
 				counter,
-				pubkey,
+				undefined,
 				outputAmounts?.keepAmounts,
-				p2pk,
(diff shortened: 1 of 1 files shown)


Egge21M (approved):
Untested, ACK

Good catch! As this is critical we should merge and release a patch asap.

Egge21M:
Could you please add a test for this too? Seems like this case was not covered.

Egge21M:
Solved as per https://github.com/cashubtc/cashu-ts/commit/18653e86478ac8b3c69cee59cf644851217c7a6f

PR #265, opened 2025-03-11
Add method for automated batch restoration

Egge21M:
## Description

This adds a new method to CashuWallet that automates the proofs restoration process according to the specifications recommendations.

## Changes

- Adds CashuWallet.batchRestore

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -602,25 +602,38 @@ class CashuWallet {
 	}
 
 	async batchRestore(
-		counter = 0,
-		currentGap = 0,
-		accProofs: Array<Proof> = [],
-		lastFoundCounter = 0
+		gapLimit = 300,
+		batchSize = 100,
+		initialCounter = 0
 	): Promise<{ proofs: Array<Proof>; lastCounter: number }> {
-		if (currentGap >= 300) {
-			return { proofs: accProofs, lastCounter: lastFoundCounter };
-		}
+		const recursiveRestore = async (
+			counter: number,
+			currentGap = 0,
+			proofs: Array<Proof> = [],
+			lastFoundCounter = 0
+		) => {
+			if (currentGap >= gapLimit) {
+				return { proofs, lastCounter: lastFoundCounter };
+			}
 
-		const { proofs } = await this.restore(counter, 100);
-		if (proofs.length === 0) {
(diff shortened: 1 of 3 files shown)


callebtc:
Nice one! I think this is so important that a nutshell integration test might be warranted.

Egge21M:
> Nice one! I think this is so important that a nutshell integration test might be warranted.

You are right. I added the test and discovered an issue with the old `lastFoundCounter`implementation. I changed `CashuWallet.restore` to also return the last found counter and refactored it for readability and time complexity. 

This should be good to go now.

lollerfirst (commented):
Personally not a fan of recursions. Is there any reason why you chose recursion as opposed to a simple loop like for [example](https://github.com/cashubtc/cdk/blob/4c447cf046e1c8107ffa40d09f2f61ac7823f46b/crates/cdk/src/wallet/mod.rs#L308) CDK?

src/CashuWallet.ts
```diff
@@ -601,6 +601,44 @@ class CashuWallet {
...
+	async batchRestore(
+		gapLimit = 300,
+		batchSize = 100,
+		initialCounter = 0
+	): Promise<{ proofs: Array<Proof>; lastFoundCounter?: number }> {
+		const recursiveRestore = async (
+			counter: number,
+			currentGap = 0,
+			proofs: Array<Proof> = [],
+			lastFoundCounter = 0
```
lollerfirst, line 613: With a recursion, you're passing down `proofs` many times. I don't know how javascript handles arrays but if this is a deep copy then for a large number of proofs and/or many batch iterations it could be a problem.

Egge21M: That is a very valid concern. It is indeed a deep copy and could potentially eat up memory for large restores. I will refactor to an interative approach.

Egge21M: Correction: the spread operator created shallow objects. However I agree that an iterative approach might increase readability

Egge21M:
Replaced the recursive function with an iterative one and added jsDoc. Now we should be good to go :)

lollerfirst (approved):

gandlafbtc (approved):
LGTM

Issue #267, opened 2025-03-12
Replace enums with const objects


Egge21M:
The concept of enums is great, but in TypeScript they behave in weird and surprising ways. While TS is a structural type language, enums are an exception to this rule.

As an example see this code
<img width="865" alt="Image" src="https://github.com/user-attachments/assets/51711182-3f8d-4a59-841c-dfea5dada480" />

Even though `"PAID"` satisfies runtime value of the type, TS does not allow us to assign this. This is fundamentally different to any other TypeScript concept where duck typing applies:

<img width="409" alt="Image" src="https://github.com/user-attachments/assets/2d4d5aec-f2ad-46f6-8d56-c4c726f67990" />

This inconsistency in the mental model can easily lead to confusion

Solution:

Replace enums with `const val = {} as const` objects. These allow us to create runtime types like enums, but still fit the mental model of structural typing (and you get proper LSP support too)

<img width="474" alt="Image" src="https://github.com/user-attachments/assets/4ca7837c-d9d6-4a3f-9283-cb32c2068b89" />

PR #270, opened 2025-03-19
Replace Enums with Const Objects

lollerfirst:
# Fixes: #267 

## Description
Replaces
```typescript
enum myEnum { val1 = 'val1',  val2 = 'val2' };
```
with
```typescript
const myEnum = { val1: 'val1', val2: 'val2'} as const;
```
and then declares a type made from the keys of the const object:
```typescript
type MyEnum = typeof MyEnum[keyof typeof MyEnum];
```
I kept the same name for both const object and type, and it doesn't seem to be creating problems nor it should because they are in different namespaces.

### DIFF
--- a/src/model/types/mint/responses.ts
+++ b/src/model/types/mint/responses.ts
@@ -30,11 +30,12 @@ export type ProofState = {
 /**
  * Enum for the state of a proof
  */
-export enum CheckStateEnum {
-	UNSPENT = 'UNSPENT',
-	PENDING = 'PENDING',
-	SPENT = 'SPENT'
-}
+export const CheckStateEnum = {
+	UNSPENT: 'UNSPENT',
+	PENDING: 'PENDING',
+	SPENT: 'SPENT'
+} as const;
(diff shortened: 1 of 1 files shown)


gandlafbtc (approved):
LGTM

src/model/types/mint/responses.ts
```diff
@@ -30,11 +30,12 @@ export type ProofState = {
...
-	UNSPENT = 'UNSPENT',
-	PENDING = 'PENDING',
-	SPENT = 'SPENT'
-}
+export const CheckStateEnum = {
+	UNSPENT: 'UNSPENT',
+	PENDING: 'PENDING',
+	SPENT: 'SPENT'
+} as const;
+export type CheckStateEnum = typeof CheckStateEnum[keyof typeof CheckStateEnum];
```
Egge21M, line 38: nit: as long as key and value are the same string `type MyEnum = keyof typeof MyEnum;` is sufficient

PR #274, opened 2025-03-23
Added getDecodedToken example in the README and an example of firebase API to receive cashu payments

nogringo:
Added `getDecodedToken` example in the README and an example of firebase API to receive cashu payments in the folder /examples/


examples/paymentApi_example.js
```diff
@@ -0,0 +1,122 @@
...
+
+    res.json({
+      success: false,
+      error: "cannot_receive_token",
+      message: `Cannot receive token: ${error.code}`
+    });
+    return;
+  }
+
+  const { keep, send } = await wallet.send(totalAmount, receiveProofs, { pubkey: p2pkLock });
```
gandlafbtc, line 103: you could skip this step by locking to the pubkey on the receive step

gandlafbtc (approved):

examples/paymentApi_example.js
nogringo: Oh thanks for the tips

Issue #266, opened 2025-03-11
Implement new fields in NUT-04/05 change


callebtc:
Implement new optional fields introduced in https://github.com/cashubtc/nuts/pull/235

PR #275, opened 2025-03-25
update response types

gandlafbtc:
# Fixes: #266 

## Description

adds new optional fields

## Changes

- add new optional fields on the mint and melt responses

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/types/mint/responses.ts
+++ b/src/model/types/mint/responses.ts
@@ -145,6 +145,14 @@ export type MeltQuoteResponse = {
 	 * Return/Change from overpaid fees. This happens due to Lighting fee estimation being inaccurate
 	 */
 	change?: Array<SerializedBlindedSignature>;
+	/**
+	 *  Payment request for the melt quote.
+	 */
+	request?: string;
+	/**
+	 *  Unit of the melt quote.
+	 */
+	unit?: string;
 } & ApiError;
 
 export const MeltQuoteState = {
@@ -190,6 +198,14 @@ export type MintQuoteResponse = {
 	 * Public key the quote is locked to
 	 */
 	pubkey?: string;
+	/**
+	 * Unit of the quote
+	 */
+	unit?: string;
+	/**
+	 * Amount requested for mint quote
+	 */
+	amount?: number;
 } & ApiError;
 
 export type LockedMintQuoteResponse = MintQuoteResponse & { pubkey: string };
--- a/src/model/types/mint/responses.ts
+++ b/src/model/types/mint/responses.ts
@@ -177,7 +177,7 @@ export type MintQuoteState = (typeof MintQuoteState)[keyof typeof MintQuoteState
 /**
  * Response from the mint after requesting a mint
  */
-export type MintQuoteResponse = {
+export type PartialMintQuoteResponse = {
 	/**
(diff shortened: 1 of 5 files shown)


src/model/types/mint/responses.ts
```diff
@@ -190,6 +198,14 @@ export type MintQuoteResponse = {
...
 	 */
 	pubkey?: string;
+	/**
+	 * Unit of the quote
+	 */
+	unit?: string;
+	/**
+	 * Amount requested for mint quote
+	 */
+	amount?: number;
```
Egge21M, line 208: The change to NUT-04/05 is mandatory. I wonder if we can avoid the optional fields here. Of course not all mints might support this straight away, but we could monkey-patch the reponse in `CashuMint.createMintQuote` if the fields are not present.

Egge21M (changes requested):

src/model/types/mint/responses.ts
gandlafbtc: That seems like the better way to do it, according to the specs. Thanks for the input!

Egge21M: I stumbled upon an issue when implementing the monkey-patch. While all information necessary is available for `createMintQuote` it is not for `checkMintQuote`. So we are back to either marking it optional or trusting the mint to return it.

We usually do not validate any mint responses, so we could simply not make this optional (as it is mandatory in the spec) and move on...

gudnuf: Seems like "trusting the mint to return it" does not work because likely not all mints will upgrade at the same time that cashu-ts makes this change. We need some sort of backwards compatibility in the meantime

Egge21M: @gudnuf I agree and I came up with a "fix". I will have it ready by weekly dev call tomorrow so that this can be merged asap.

Egge21M: @gudnuf what do you think about this?
(PartialPick<T,K keyof T> makes K optional in T)

```ts
/**
	 * Gets an existing mint quote from the mint.
	 * @param quote Quote ID
	 * @returns the mint will create and return a Lightning invoice for the specified amount
	 */
	async checkMintQuote<T extends string | MintQuoteResponse>(
		quote: T
	): Promise<
		T extends MintQuoteResponse
			? MintQuoteResponse 
			: PartialPick<MintQuoteResponse, "unit" | "amount"> 
	> {
		const quoteId = typeof quote === 'string' ? quote : quote.quote;
		const baseRes = await this.mint.checkMintQuote(quoteId);
		if (typeof quote === 'string') {
			return baseRes;
		}
		return { ...baseRes, amount: quote.amount, unit: quote.unit };
	}
```

gudnuf: Yea this seems pretty good. Its a bit weird because the mint could return amount and unit even if you just pass quote as a string and then those keys won't be present on the type, but this does seem like a good middle ground

Egge21M: The keys are present, but optional. There is no way to tell at compile time whether those fields will be present if you pass in a string. The only way to guarantee is using the full response object.

Issue #231, opened 2024-12-27
Re-structure package/publish structure as monorepo


gandlafbtc:
For improved DX, we should migrate to using a monorepo and consolidate the cashu-crypto-ts repo.

we could use lerna: https://lerna.js.org/

Egge21M:
Yes, let's use this thread to discuss possible structures. The amount of assets that we are building is quite low, so I personally do not think that we need to introduce any new dependencies to the project. Here is what I had in mind:

- include the crypto-package in the main repo
- replace the npm-hosted dependency with a link of the local one
- use npm workspaces to bundle the crypto-package before bundling the library

That way we can still publish the crypto-package as a separate module if desirable, but make sure that cashu-ts always uses the latests version by default and developers can do all changes in a single repo.

We could also use git submodules to keep git history separated, but I dont think that is a good thing in our case and slows down development. Instead we could come up with a rule that all commits for the crypto lib should be squashed and prepended with `crypto:`

PR #279, opened 2025-03-27
Bringing back cashu-crypto-ts code into cashu-ts

gandlafbtc:
# Fixes: #231 

## Description

Moving crypto back into cashu-ts

## Changes

- moved crypto code
- fixed imports and tests
- updated deps

## Not implemented

- Modularized exports

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


Egge21M:
I think we should create the monorepo structure first and then move things over

gandlafbtc:
i don't even wanna necessarily pull the crypto lib into cashu-ts as a "package". I'd rather build it in a way that lets us make use of modular exports, but all in one lib (crypto + cashu-ts). So in a sense, it has nothing to do with monorepo. I guess the branch name is poorly chosen.

Egge21M:
> i don't even wanna necessarily pull the crypto lib into cashu-ts as a "package". I'd rather build it in a way that lets us make use of modular exports, but all in one lib (crypto + cashu-ts). So in a sense, it has nothing to do with monorepo. I guess the branch name is poorly chosen.

So the crypto package will no longer be released as a separate package and will only live inside Cashu-TS?

gandlafbtc:
yes, and should be exported as a module from cashu-ts

Egge21M:
Maybe I am missing something here, but I don't think this PR added a build step for the crypto package, right? So as of right now all the crypto package is not exposed to the public and the export defined in the package.json should not resolve at all

PR #284, opened 2025-04-10
Added NUT-11 P2PK multisig support and corresponding tests

robwoodgate:
Adds [NUT-11 P2PK multisig support](https://github.com/cashubtc/nuts/blob/main/11.md) and corresponding tests.

This PR is backwards compatible, is formatted and passes unit tests.

## Changes

- Updates the p2pk type signature from:

`p2pk?: { pubkey: string; locktime?: number; refundKeys?: Array<string> }`

to support multisig tags in a backwards compatible way:

```
p2pk?: {
    pubkey: string | Array<string>;
    locktime?: number;
    refundKeys?: Array<string>;
    nsig?: number;
    rsig?: number;
}
```

- Implements multisig handling in `createSingleP2PKData()`, also in a backwards compatible way.

- Added unit tests to confirm functionality works as per NUT-11 spec

- Adds utility functions for P2PK handling

- Exposes P2PK utility functions to public API

- Harmonizes NUT-11 crypto module client side to use the main cashu-ts Proof type definition 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

## Post PR Tasks

- The change in p2pk type signature should be noted in documentation.

### DIFF
--- a/src/crypto/client/NUT11.ts
+++ b/src/crypto/client/NUT11.ts
@@ -62,14 +62,16 @@ export const verifyP2PKSecretSignature = (
  */
 export function getP2PKExpectedKWitnessPubkeys(secret: Secret): Array<string> {
 	try {
-		const now = Math.floor(Date.now() / 1000);
+		// Validate secret format
+		if (secret[0] !== 'P2PK') {
+			throw new Error('Invalid P2PK secret: must start with "P2PK"');
+		}
 		const { data, tags } = secret[1];
+		const now = Math.floor(Date.now() / 1000);
 		const locktime = getP2PKLocktime(secret);
-		const n_sigsTag = tags && tags.find((tag) => tag[0] === 'n_sigs');
-		const n_sigs = n_sigsTag && n_sigsTag.length > 1 ? parseInt(n_sigsTag[1], 10) : null;
 		if (locktime > now) {
 			// Am interpretting NUT-11 as intending pubkeys to be usable for a
-			// 1-of-m multisig if provided, even if n_sigs is not
+			// 1-of-m multisig if provided, even if n_sigs is not set
 			const pubkeysTag = tags && tags.find((tag) => tag[0] === 'pubkeys');
 			const pubkeys = pubkeysTag && pubkeysTag.length > 1 ? pubkeysTag.slice(1) : [];
 			return [data, ...pubkeys];
@@ -80,7 +82,7 @@ export function getP2PKExpectedKWitnessPubkeys(secret: Secret): Array<string> {
 			return refundKeys;
 		}
 	} catch {}
-	return []; // Unlocked or expired with no refund keys
+	return []; // Unlocked, malformed or expired with no refund keys
 }
 
 /**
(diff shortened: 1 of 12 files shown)


lescuer97:
I believe this could be a problem:
https://github.com/cashubtc/cashu-crypto-ts/blob/master/src/client/NUT11.ts

getSignedProof makes a new array every time the function is called. So previous signatures are deleted. 

This is my understanding of the code at least.

robwoodgate:
Thank you for highlighting this. There is definitely an issue with the multisig signing in `getSigProof()`, but IMO, it doesn't invalidate the PR... as this is concerned with CREATING multisig proofs.

It is, however, something we should fix to allow multisig NUT-11 handling on RECEIVE.

A fix might look something like:

```
export const getSignedProof = (proof: Proof, privateKey: PrivKey): Proof => {
  const signature = bytesToHex(signP2PKsecret(proof.secret, privateKey));
  if (!proof.witness) {
    proof.witness = { signatures: [signature] };
  } else {
    proof.witness.signatures = [...(proof.witness.signatures || []), signature];
  }
  return proof;
};
```

Happy to add it to the PR or it can be submitted separately as part of improving receive side of multisig.

robwoodgate:
> Thank you for the work!
> 
> I believe this could be a problem: https://github.com/cashubtc/cashu-crypto-ts/blob/master/src/client/NUT11.ts
> 
> getSignedProof makes a new array in the signatures field every time the function is called.
> 
> So previous signatures would be deleted.
> 
> This is my understanding of the code at least.

I've added a PR to cashu-crypto-ts (see ref above).

robwoodgate:
I saw, be great to have harmonised type definitions for Proof etc at that point

gandlafbtc:
Yeah that's a good point, I opened an issue for that #289

We merged the #279 PR, so the crypto stuff is back, so if you don't mind opening a PR here, we can get it in more quickly!

Cashu-crypto-ts will be archived

robwoodgate:
Ok no problem, will look to do it Tuesday when back from Easter break 🫡

robwoodgate:
I've ported over the multisig changes to the old cashu-crypto-ts library, and enhanced support for NUT-11 including the new `n_sigs_refund` tag.

The integration tests with Nutshell are failing most likely because of the stricter criteria for signatures. 

Signatures are only added if:
- the proof P2PK conditions say the public key should sign
- the public key has not already signed

It's likely the tests need updating in nutshell.

robwoodgate:
I've ported over the multisig changes to the old cashu-crypto-ts library, and enhanced support for NUT-11 including the new n_sigs_refund tag. Integration test required tweaking as this update no longer allows ineligible witnesses to sign a proof.

robwoodgate:
Final update this morning - have extended the p2pk type signature to include an `rsigs` option to support `n_sigs_refund` tag. Unit tests added too, so we should be in good shape to support calle's [upcoming changes to NUT-11](https://github.com/cashubtc/nuts/pull/149) too.

**Note:** I have assumed my suggested clarifications of NUT-11 are correct. See the PR review comments.

robwoodgate:
Have expanded test coverage, refactored/simplified a couple of the utility functions and squashed a bug.

src/crypto/client/NUT11.ts
```diff
@@ -30,42 +31,195 @@ export const signBlindedMessage = (B_: string, privateKey: PrivKey): Uint8Array
...
+	const { tags } = secret[1];
+	const sigFlagTag = tags && tags.find((tag) => tag[0] === 'sigflag');
+	return sigFlagTag && sigFlagTag.length > 1 ? sigFlagTag[1] : 'SIG_INPUTS';
+}
+
+/**
+ * Gets witness signatures as an array
+ * @type {array} of signatures
+ */
+export const getSignatures = (witness: string | P2PKWitness | undefined): Array<string> => {
```
lollerfirst, line 150:
```suggestion
export const getWitnessSignatures = (witness: string | P2PKWitness | undefined): Array<string> => {
```

src/crypto/client/NUT11.ts
```diff
@@ -30,42 +31,195 @@ export const signBlindedMessage = (B_: string, privateKey: PrivKey): Uint8Array
...
+export const getSignedProof = (proof: Proof, privateKey: string): Proof => {
+	// Check secret is P2PK
+	const parsed: Secret = parseSecret(proof.secret);
+	if (parsed[0] !== 'P2PK') {
+		throw new Error('not a P2PK secret');
+	}
+	// Check this pubkey is required to sign
+	const pubkey = bytesToHex(secp256k1.getPublicKey(privateKey)); // for Cashu
+	const witnesses = getP2PKExpectedKWitnessPubkeys(parsed);
+	if (!witnesses.length || !witnesses.includes(pubkey)) {
```
lollerfirst, line 202: Should we throw in this case?

test/wallet.test.ts
```diff
@@ -897,6 +897,177 @@ describe('P2PK BlindingData', () => {
...
+			expect(s[1].data).toBe('thisisatest');
+			expect(s[1].tags).toContainEqual(['pubkeys', 'asecondpk', 'athirdpk']);
+			expect(s[1].tags).not.toContainEqual(['n_sigs', 1]);
+		});
+	});
+	test('Create BlindingData locked to multiple pks with 2-of-3 nsig', async () => {
+		const wallet = new CashuWallet(mint);
+		const keys = await wallet.getKeys();
+		const data = OutputData.createP2PKData(
+			{ pubkey: ['thisisatest', 'asecondpk', 'athirdpk'], nsig: 2 },
```
lollerfirst, line 933: Should we make it so that we only allow 65 char hex in the `pubkey` argument?

lollerfirst (approved):

src/crypto/client/NUT11.ts
robwoodgate: We could... as getSignedProofs() would catch and return the `proof` in any case. But not sure it's necessary, as attempting to sign a `proof` that doesn't need signing is not really a fatal error, more a soft one.

src/crypto/client/NUT11.ts
robwoodgate: Good idea - I've pushed a commit to change this in all instances

robwoodgate:
Crypto client side module now uses the main cashu-ts `Proof` type definition, and the P2PK utility functions have been exposed to the public API as they are generally useful for apps (eg  Cashu Witness)

robwoodgate:
I widened the P2PK utility functions to allow string input, as this is what a public proof.secret will be.
I don't know how to resolve the API change warning, it's an intentional change (and only affects this development branch).

robwoodgate:
This turned into a much wider set of changes than originally planned... I now have cashu witness and cashu redeem running off this branch now to battle test the utility functions. I'd really like to freeze this PR for merge before it gets any bigger, so wondering if I should close this PR and open the changes as a feature branch of development?

PR #297, opened 2025-05-01
Modular exports

gandlafbtc:
# Fixes: #290 

## Description

Since we're integrating cashu-crypto into cashu-ts , we want to export its components in a modularized fashion

## Changes

- changed vite config to support modules
- export modules in package.json

## PR Tasks

- [x] Open PR
- [x] run `npm run test` --> no failing unit tests
- [x] run `npm run format`

PR #299, opened 2025-05-08
Added NUT-11 P2PK multisig support and corresponding tests

robwoodgate:
# Replaces: https://github.com/cashubtc/cashu-ts/pull/284

Adds [NUT-11 P2PK multisig support](https://github.com/cashubtc/nuts/blob/main/11.md) and corresponding tests.

This PR is backwards compatible, is formatted and passes unit tests.
**NOTE:** it expands the public API to include the utility functions. (EDIT: via #297)

## Changes

- Updates the p2pk type signature from:

`p2pk?: { pubkey: string; locktime?: number; refundKeys?: Array<string> }`

to support multisig tags in a backwards compatible way:

```
p2pk?: {
    pubkey: string | Array<string>;
    locktime?: number;
    refundKeys?: Array<string>;
    requiredSignatures?: number;
    requiredRefundSignatures?: number;
}
```

- Implements multisig handling in `createSingleP2PKData()`, also in a backwards compatible way.

- Added unit tests to confirm functionality works as per NUT-11 spec

- Adds utility functions for P2PK handling

- Exposes P2PK utility functions to public API

- Harmonizes NUT-11 crypto module client side to use the main cashu-ts Proof type definition 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

## Post PR Tasks

- The change in p2pk type signature should be noted in documentation.


robwoodgate:
I ran `npm run api:update` and `npm run api:check` but cannot seem to clear the API check error.

src/crypto/client/NUT11.ts
```diff
@@ -1,75 +1,261 @@
...
 
-export const signBlindedMessage = (B_: string, privateKey: PrivKey): Uint8Array => {
+export const signBlindedMessage = (B_: string, privateKey: PrivKey): string => {
 	const msgHash = sha256(B_);
 	const sig = schnorr.sign(msgHash, privateKey);
-	return sig;
+	return bytesToHex(sig);
 };
 
-export const getSignedProofs = (
```
robwoodgate, line 33: **Note for reviewers:** Git has made the following section look way more complex than it really is...
The old `getSignedProofs` and `getSignedProof` functions are renamed and rebuilt as `signP2PKProofs` and `signP2PKProof`. The rest is a bunch of new P2PK utility functions.

src/model/OutputData.ts
```diff
@@ -80,23 +86,48 @@ export class OutputData implements OutputDataLike {
...
+		if (pubkeys.length > 1) {
+			newSecret[1].tags.push(['pubkeys', ...pubkeys.slice(1)]); // Additional keys
+			if (n_sigs > 1) {
+				// 1 is the default, so we can save space if not multisig
+				newSecret[1].tags.push(['n_sigs', String(n_sigs)]); // NUT-10 string
+			}
 		}
 		if (p2pk.refundKeys) {
 			newSecret[1].tags.push(['refund', ...p2pk.refundKeys]);
+			if (n_sigs_refund > 1) {
```
robwoodgate, line 127: Anticipates the NUT-11 change here: https://github.com/cashubtc/nuts/pull/149

src/CashuWallet.ts
```diff
@@ -910,17 +910,9 @@ class CashuWallet {
...
-					return {
-						amount: p.amount,
-						C: pointFromHex(p.C),
-						id: p.id,
-						secret: new TextEncoder().encode(p.secret)
-					};
-				}),
-				privkey
-			).map((p: NUT11Proof) => serializeProof(p));
+			proofsToSend = signP2PKProofs(proofsToSend, privkey).map((p: Proof) => {
```
robwoodgate, line 913: Simpler now that crypto/client uses the main Proof type definition.

src/CashuWallet.ts
```diff
@@ -1016,17 +1014,9 @@ class CashuWallet {
...
```
robwoodgate, line 1017: Simpler now that crypto/client uses the main Proof type definition.

test/wallet.test.ts
```diff
@@ -897,6 +897,177 @@ describe('P2PK BlindingData', () => {
 			expect(s[1].tags).toEqual([]);
 		});
 	});
+	test('Create BlindingData locked to single pk with unexpected nsig', async () => {
```
robwoodgate, line 900: Lots of tests...

robwoodgate (commented):
Added some notes to help reviewers navigate this PR.

src/crypto/client/NUT11.ts
```diff
@@ -1,75 +1,261 @@
...
 		}
 	});
+	if (alreadySigned) {
+		console.warn(`Proof already signed by [02|03]${pubkey}`);
+		return proof; // Skip signing if pubkey has a valid signature
+	}
+	// Add new signature
+	const signature = signP2PKSecret(proof.secret, privateKey);
+	signatures.push(signature);
+	return { ...proof, witness: { signatures } };
```
robwoodgate, line 252: I left the witness in `P2PKWitness` type format for now, as that was the original output of the function, but we should consider whether the JSON.stringified() format is more appropriate going forward.

At the moment, CashuWallet needs to stringify witness to use it (see my 2 x `Simpler now` comments for refs), and aside from the test suite, is the only internal consumer. So internally, a stringified witness would be optimal.

The wrinkle is that, as a public as a utility function, there is a conflict of desires for witness output type due to the way tokens are encoded/decoded.

1)  For proofs being sent in for melt/swap/receive etc, the witness is expected to be a string as per NUT-11
2) For encoding signed proofs back into a token, the witness is expected to be a `P2PKWitness`.

My gut feeling is that a decoded token proof SHOULD always look like a NUT-11 proof - ie witness is JSON stringified in public, and only parsed internally as needed.

This would require `templateFromToken` to treat `proof.witness` as already stringified to bring it into line, and all tests updated to expect a stringified witness return.

src/index.ts
```diff
@@ -14,6 +14,16 @@ import {
 	hasValidDleq
 } from './utils.js';
 import { CashuAuthMint, CashuAuthWallet, getBlindedAuthToken, getEncodedAuthToken } from './auth';
```
robwoodgate, line 16: Assuming the [modularized exports PR](https://github.com/cashubtc/cashu-ts/pull/297) is added first, we should not need to expose the NUT-11 utility functions to the public API here.

src/crypto/client/NUT11.ts
```diff
@@ -1,75 +1,261 @@
...
-		pk = privateKey;
+	});
+};
+
+/**
+ * Returns the expected witness public keys from a NUT-11 P2PK secret
+ * @param secretStr - The NUT-11 P2PK secret.
+ * @returns {array} with the public keys or empty array
+ */
+export function getP2PKExpectedKWitnessPubkeys(secretStr: string | Secret): Array<string> {
```
Egge21M, line 85: I am unsure if this method does not do too much. I see the ergonomics of having a single method to get either regular pubkeys or refund pubkeys, but calling this method does hide the fact of whether a lock is actually expired.

What do you think about splitting this up in
- getP2PLLockTime
- getP2PKExpectedWitnessPubkeys
- getP2PKExpectedWitnessRefundkeys

We could then add a helper thats called something like `getP2PKExpectedSignatories` that does all the above in one sweep

src/crypto/client/NUT11.ts
```diff
@@ -1,75 +1,261 @@
...
+	const parsed: Secret = parseP2PKSecret(proof.secret);
+	if (parsed[0] !== 'P2PK') {
+		throw new Error('not a P2PK secret');
+	}
+	// Check if the private key is required to sign by checking its
+	// X-only pubkey (no 02/03 prefix) against the expected witness pubkeys
+	// NB: Nostr pubkeys prepend 02 by convention, ignoring actual Y-parity
+	const pubkey = bytesToHex(schnorr.getPublicKey(privateKey)); // x-only
+	const witnesses = getP2PKExpectedKWitnessPubkeys(parsed);
+	if (!witnesses.length || !witnesses.some((w) => w.includes(pubkey))) {
```
Egge21M, line 232: I have the feeling that if you try to sign a proof that does not require a singature of this key, this function should throw. I might be wrong though. Whats your rationale on this?

src/crypto/client/NUT11.ts
Egge21M: I think serialization should always happen at the edges (e.g. when encoding a token, or sending proofs to the mint). If we pass around serialized data around internally, we might spend a lot of time parsing and stringifying in each method that needs to access the witness data.

robwoodgate: I generally agree, the problem is the edges are a little fuzzy in Wallet… 

But if we agree that’s the goal, then this function is ok and we can work towards making the edges clearer going forward

src/crypto/client/NUT11.ts
robwoodgate: The questions apps need to ask/answer for P2PK proofs are:

- Is the Proof locked? If so:
- Who is expected to witness it?
- How many witnesses must sign?
- Will the witness requirements change?

The `getP2PKExpectedKWitnessPubkeys()` function answers the first two questions by returning the eligible / expected witnesses if the proof is locked, or an empty array if the proof is unlocked.

The `getP2PKNSigs` function answers the question of how many of the expected witnesses must sign.

The `getP2PKLocktime` function answers the question of when the witness requirements will (or did) change

There may be some marginal value in abstracting the raw lookup of witness/refund witness to helpers, as this would allow apps (and this function) to save 2 lines of code filtering each for themselves if they want that info.
Eg: `getP2PKWitnessPubkeys` / `getP2PKWitnessRefundkeys`

But to get **_Expected_** witness keys requires the full P2PK logic to be invoked... That's because each piece of witness information is only useful in the context of the Locktime... and you only get the full picture when you consider both lock witness and refund witnesses together.

src/crypto/client/NUT11.ts
robwoodgate: My rationale for a soft fail with `console.warn()` is that signing is a threshold operation (n-of-m), not a fail safe one - a mint would simply ignore duplicate, irrelevant, or invalid signatures.

For example: Alice and Carol are 2-of-2 locktime witnesses, the refund witness is Carol only. Alice signs in time and sends the proof to Carol, who signs and redeems after locktime has expired. The mint only cares that Carol has signed, meeting the threshold. Alice's signature is now irrelevant and is ignored.

So we ony really care about enforcing eligibility from a performance / token size perspective. Adding a throw here would impose a stricter constraints than necessary, limiting flexibility and adding performance overhead.

Example: A Nostr user may have multiple NIP-60 private keys, as well as their main NSEC. The soft fail allows all keys to be tried - a hard throw would require an app to derive public keys and check they are expected before sending them in for signing... the function then derives and checks them all over again.

If we think a throw may be useful here, then perhaps we can adjust `signP2PKProofs` to catch and ignore these errors to continue processing other keys (which is partially handled). We could also add a boolean `strict` flag to give callers more control over whether they want a stricter throw or not?

robwoodgate:
Have split out `getP2PKWitnessPubkeys` / `getP2PKWitnessRefundkeys`, added throw to `getSignedProof` and `beStrict` option to `getSignedProofs`. Still can't get API check to pass for some reason, have run check and update

src/crypto/client/NUT11.ts
robwoodgate: I agree there is value in throwing for `getSignedProof`, as it's one key, one proof, so have implemented, defaulting `getSignedProofs` to soft fail as it's multi-key, multi-proof.

src/crypto/client/NUT11.ts
robwoodgate: Have split out `getP2PKWitnessPubkeys` / `getP2PKWitnessRefundkeys` into their own utility functions as it may be useful for some apps.

robwoodgate:
Rebased to development branch... have also added filters to getP2PKWitnessPubkeys / getP2PKWitnessRefundkeys to handle NUT-10 non-compliant P2PK secrets.

src/index.ts
robwoodgate: Have removed this section now as modular exports have been merged.

Egge21M:
Awesome, thank you @robwoodgate. I am still travelling but will be able to confirm and merge coming Thursday. @gandlafbtc please base your work of this or review and merge this if you find time.

src/CashuWallet.ts
```diff
@@ -952,7 +949,13 @@ class CashuWallet {
 			keep?: Array<OutputDataLike> | OutputDataFactory;
 			send?: Array<OutputDataLike> | OutputDataFactory;
 		},
-		p2pk?: { pubkey: string; locktime?: number; refundKeys?: Array<string> }
+		p2pk?: {
+			pubkey: string | Array<string>;
+			locktime?: number;
+			refundKeys?: Array<string>;
+			nsig?: number;
+			rsig?: number;
```
gandlafbtc: are  rsig refund sigs? to me it is kinda unlcear, just from looking at the code. I think we should consider renaming it or adding some docs

gandlafbtc: change the nsig and rsig to:
`requiredSignatures`
`requiredRefundSignatures`

robwoodgate: Implemented, and TSDOCS updated as appropriate. Thanks @gandlafbtc !

PR #300, opened 2025-05-08
Ensure active keyset chosen has the wallet unit

robwoodgate:
## Description

This makes CashuWallet more robust by avoiding the need for apps to filter persisted keysets by unit before instantiating a CashuWallet instance.

This makes the persistence of mint keys/keysets and instantiation simpler, and prevents input/output keyset unit mismatch MintOperation errors if the app does not properly filter keysets when loading them.

## Changes

- Adds a unit filter to `getActiveKeyset()`.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -192,7 +192,7 @@ class CashuWallet {
 	 * @returns active keyset
 	 */
 	getActiveKeyset(keysets: Array<MintKeyset>): MintKeyset {
-		let activeKeysets = keysets.filter((k: MintKeyset) => k.active);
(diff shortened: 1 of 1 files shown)


robwoodgate:
I have not done so, but having added unit filtering to `getActiveKeyset()`, we could remove it from `getKeySets()`. 
This would allow a wallet to get any keyset without calling the mint API, while wallet operations use the `getKeys() -> getActiveKeyset()` pathway to get the working keyset in correct unit.

callebtc (approved):
utACK

PR #309, opened 2025-05-23
Remove `stripDleq` from `CashuWallet.send`

Egge21M:
# Fixes: #[308]

## Description
This PR makes sure that DLEQs are not removed from proofs when calling 'CashuWallet.send'. The 'includeDleq' option was overloaded and not easy to understand, so removing this seems to be the safest option.

## Changes

- Remove 'stripDleq' from 'CashuWallet.send'
- Add 'removeDleq' to 'utils.getEncodedToken'

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/test/integration.test.ts
+++ b/test/integration.test.ts
@@ -18,6 +18,7 @@ import { MintOperationError } from '../src/model/Errors.js';
 import ws from 'ws';
 import { injectWebSocketImpl } from '../src/ws.js';
 import {
+	getDecodedToken,
 	getEncodedToken,
 	getEncodedTokenV4,
 	hexToNumber,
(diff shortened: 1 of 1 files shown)


gandlafbtc:
@Egge21M there are some pipelines failing, is this ready to be reviewed? Or will you take another look at it first?

Egge21M:
@gandlafbtc, sorry forgot about the integration test. Updated and all green now

lollerfirst:
> FYI in theory this is a breaking change, however the old behaviour was not documented so it's more like removing an unwanted side effect. But as hyrum's law suggest this could still affect some consumers

ACK. Worst that can happen is DLEQ gets included when the user is asking to strip it. Can we merge this?

PR #313, opened 2025-06-06
[FEAT] GCS Filter Support

lollerfirst:
## Description

Add support for getting spent ecash filters

## Changes

- `CashuMint.getSpentFilter(keysetId: str)`
- `CashuWallet.getSpentFilter(keysetId: str)`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -1,6 +1,6 @@
 import { serializeProof } from './crypto/client/index.js';
 import { getSignedProofs } from './crypto/client/NUT11.js';
-import { BlindSignature, hashToCurve, pointFromHex, type Proof as NUT11Proof } from './crypto/common/index.js';
+import { hashToCurve, pointFromHex, type Proof as NUT11Proof } from './crypto/common/index.js';
 import { CashuMint } from './CashuMint.js';
 import { MintInfo } from './model/MintInfo.js';
 import {
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -1,6 +1,6 @@
 import { serializeProof } from './crypto/client/index.js';
 import { getSignedProofs } from './crypto/client/NUT11.js';
-import { hashToCurve, pointFromHex, type Proof as NUT11Proof } from './crypto/common/index.js';
+import { BlindSignature, hashToCurve, pointFromHex, type Proof as NUT11Proof } from './crypto/common/index.js';
 import { CashuMint } from './CashuMint.js';
 import { MintInfo } from './model/MintInfo.js';
 import {
@@ -1213,6 +1213,15 @@ class CashuWallet {
 		return () => {
 			this.mint.webSocketConnection?.cancelSubscription(subId, callback);
 		};
+	}
+
+	/**
+	 * Get the spent ecash filter for a specific keyset
+	 * @param The keyset ID
+	 * @returns `GetFilterResponse`
+	 */
+	async getSpentFilter(keysetId: string) {
+		return this.mint.getSpentFilter(keysetId);
 	}
 
 	/**
--- a/src/model/types/mint/responses.ts
+++ b/src/model/types/mint/responses.ts
@@ -309,3 +309,14 @@ export type WebSocketSupport = {
 export type BlindAuthMintResponse = {
 	signatures: Array<SerializedBlindedSignature>;
 } & ApiError;
+
+/**
+ * Response to a get spent filter request
+ */
+export type GetFilterResponse = {
+	n: number,
+	p?: number,
+	m?: number,
+	content: string,
+	timestamp: number,
+} & ApiError;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@cashu/cashu-ts",
-	"version": "2.5.0",
+	"version": "2.5.1",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@cashu/cashu-ts",
-			"version": "2.5.0",
+			"version": "2.5.1",
 			"license": "MIT",
 			"dependencies": {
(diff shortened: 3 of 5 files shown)


gandlafbtc:
Nice! ready for review @lollerfirst ?

gandlafbtc:
Awesome, great work! did you already run the integration tests against a local mint instance that supports filters?

lollerfirst:
> Awesome, great work! did you already run the integration tests against a local mint instance that supports filters?

Yes I did. The Mint would also have to be configured to recompute the filters within 5 seconds.

src/CashuWallet.ts
```diff
@@ -594,8 +598,18 @@ class CashuWallet {
...
 
+		if (!keysetId) {
+			keysetId = this.keysetId;
+		}
+
+		const mintInfo = await this.lazyGetMintInfo();
+		const issuedFilter = mintInfo.isSupported(25).supported
+			? await this.getIssuedFilter(keysetId)
+			: undefined;
+		console.debug(`issuedFilter.numItems: ${issuedFilter?.numItems}`);
```
robwoodgate:
```suggestion
		this._logger.debug(`issuedFilter.numItems: ${issuedFilter?.numItems}`);
```

src/model/types/mint/responses.ts
```diff
@@ -309,3 +313,14 @@ export type WebSocketSupport = {
...
+/**
+ * Response to a get spent filter request
+ */
+export type GetFilterResponse = {
+	n: number;
+	p?: number;
+	m?: number;
+	content: string;
+	timestamp: number;
+} & ApiError;
```
robwoodgate: We should TSDOC the `n`, `p`, `m` etc

src/gcs.ts
```diff
@@ -0,0 +1,191 @@
...
+
+	const x = (q << P) | r;
+	return [x, offset + Number(P)];
+}
+
+export class GCSFilter {
+	numItems: number;
+	invFpr: number;
+	remBitlength: number;
+	content: Buffer;
```
robwoodgate, line 112: We should TSDOC these params

src/CashuWallet.ts
```diff
@@ -51,6 +52,9 @@ import {
 	OutputDataLike,
 	isOutputDataFactory
 } from './model/OutputData.js';
+import { GCSFilter } from './gcs.js';
+import { BlindedMessage } from './model/BlindedMessage.js';
+import { BlindedSignature } from './model/BlindedSignature.js';
```
robwoodgate: More generally, do we need the .js extensions? #332

robwoodgate (approved):
utACK - great work @lollerfirst

lollerfirst:
@robwoodgate @gandlafbtc @Egge21M Do you think GCS filters should be received through streaming HTTP responses?

robwoodgate:
> @robwoodgate @gandlafbtc @Egge21M Do you think GCS filters should be received through streaming HTTP responses?

I'm not sure that's useful... as I understand it, GCS filters are atomic blobs that need to be computed periodically. In BIP158 (compact block filters), for example, a new GCS filter is generated per block.

We are currently computing one per keyset, which becomes intensive over time, and so the update period is going to be minutes to an hour, making streaming HTTP less efficient than simple periodic fetches.

lollerfirst:
> > @robwoodgate @gandlafbtc @Egge21M Do you think GCS filters should be received through streaming HTTP responses?
> 
> I'm not sure that's useful... as I understand it, GCS filters are atomic blobs that need to be computed periodically. In BIP158 (compact block filters), for example, a new GCS filter is generated per block.
> 
> We are currently computing one per keyset, which becomes intensive over time, and so the update period is going to be minutes / hours / days, making streaming HTTP less efficient than simple periodic fetches.

These GCS filters can be processed in segments. Imagine this: you're trying to restore, and you need to download a filter that is 50 megabytes in size. Instead of waiting for and loading the entire 50 megabytes you process through chunks with your `Y` and `B_` values. 

The only problem is the BATCHES. Since we are doing recovery with batches, then each batch would have to go through the whole filter again (which prohibits us from throwing away chunks of the filter we already checked).

lollerfirst:
> **EDIT:** Even if, say, the mint calculated the full filter hourly... and a delta filter every 5 minutes (for transactions in that 5 minute period), I suspect it would still be more resource efficient for mint to deliver the last hourly and 5m deltas in one JSON array on poll than to stream them.
> 
> Streaming might make sense if deltas are, say, every minute, delivered in real time, but that might be computationally and resource expensive for the mint.

I think there is a misunderstanding: the Mint wouldn't change the way it computes the filter. It would still do it hourly or so, but the way it's returned to the requesting clients would be "streamed" through chunks:

1. Client requests filter
2. Mint pulls it from database
3. Mint returns the parameters in the response headers, like:
```
X-Filter-N: 10000000
X-Filter-P: 19
X-Filter-M: 784931
X-Filter-timestamp: <timestamp>
Content-Length: <num_bytes>

<raw_filter_content_bytes>
```

robwoodgate:
> These GCS filters can be processed in segments. Imagine this: you're trying to restore, and you need to download a filter that is 50 megabytes in size. Instead of waiting for and loading the entire 50 megabytes into memory, you could test through chunks of it with your `Y` and `B_` values. You could even stop downloading the filter early when you find that the last processed value is greater than the greatest value in your `Y` or `B_`.

Ok, I was (as per my edit above) thinking in terms of "real time updates". Streaming / chunking could be useful in the use case you mention, if the filters are large. Being able to stop when found / resume in case of connection issue etc is a benefit, potentially saving bandwidth, memory etc too.

gandlafbtc:
Streaming sounds like a good way to do it. But maybe we can support both streaming and downloading the whole thing in one go. Seems like the latter is way lower complexity (if the filter size allows for it)

Issue #310, opened 2025-05-25
selectProofsToSend does not always select the correct amounts when including fees.


gudnuf:
# The bug

When including fees to send, there's an edge case where the `selectProofsToSend` method does not select enough proofs to send.

Here I show the denominations of all my proofs, then the denominations that were selected (which sum to 5233)

<img width="696" alt="Image" src="https://github.com/user-attachments/assets/cbed23c4-9581-4aa8-a3c1-4d85acbc127b" />

You'll see that there are 13 selected proofs, and the keyset that I am using has a 100 ppk fee which results in a 2 sat fee.

## Expected behavior

This method should return proofs that sum to the requested amount + fee. So in this case I would expect the selected proofs to sum to at least 5234

lollerfirst:
`src/CashuWallet.ts:393-395`
```typescript
const smallerProofs = sortedProofs
	.filter((p: Proof) => p.amount <= amountToSend)
	.sort((a: Proof, b: Proof) => b.amount - a.amount);
```

when `amountToSend` is a float between (0, 1] then `smallerProofs = []`.
Then you have this:
```typescript
if (!smallerProofs.length && !nextBigger) {
	return { keep: proofs, send: [] };
}
```

jbojcic1:
@lollerfirst `selectProofsToSend` is still pretty confusing. It would be nice to at least add some tsdoc description for it.  Also imo if you don't have enough proofs for the amount would be nicer if the method would return `{ keep: proofs, send: [] }`. Not sure it makes sense to return send with something that won't work.

Also I still wonder how can this part just ignore the includeFees flag:
```
if (!smallerProofs.length && nextBigger) {
	return {
		keep: proofs.filter((p: Proof) => p.secret !== nextBigger.secret),
		send: [nextBigger]
	};
}
```

couldn't it happen that there is bigger and no smaller proofs but amount to send + keyset fee is still bigger than the amount to send?

lollerfirst:
> This function is still failing in some circumstances... for example, when `remainder > 0` but no proofs are `<= remainder`.

This function isn't perfect. There are some cases where it doesn't work as it should. But in this case I get a different result from yours
```
stdout | test/wallet.test.ts > Test coinselection > offline coinselection with input fee 600 ppk
smallerProofs: 16,16,8,2,1,1
biggerProofs: 

smallerProofs: 8,2,1,1
biggerProofs: 16

smallerProofs: 2,1,1
biggerProofs: 

[...]

PR #314, opened 2025-06-12
Implements RGLI algorithm for selectProofsToSend(), plus tests

robwoodgate:
# Fixes: #310 

## Description

Thie PR adds an adapted RGLI (Randomized Greedy with Local Improvement) algorithm to the `selectProofsToSend()` method.

This allows proofs to be selected based on amount, fee inclusion, and exact match requirements.

The tests include aggressive selection with 10,000 proofs as well as combinations of exact and close match.

## Changes

- `selectProofsToSend()` now has an optional `exactMatch` param, which defaults to false, ensuring it is backwards compatible. **EDIT: This param has been removed as per dev meeting.**

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


src/CashuWallet.ts
```diff
@@ -383,54 +385,218 @@ class CashuWallet {
...
+			}
+
+			// Update best solution
+			const cost = calculateCost(amount, feePPK, S.length);
+			if (cost < bestCost) {
+				bestSubset = [...S];
+				bestCost = cost;
+			}
+
+			// If not minimizing costs (!includeFees) or proof set is large (>MAX_PROOFS)
```
a1denvalu3: I don't really understand this deviation from the paper.

src/CashuWallet.ts
```diff
@@ -383,54 +385,218 @@ class CashuWallet {
...
+		/**
+		 * Helper functions
+		 */
+		const amountExFee = (p: Proof): number => {
+			return includeFees ? p.amount - (proofToFeePPK.get(p) ?? 0) / 1000 : p.amount;
+		};
+		const sumExFees = (amount: number, feePPK: number): number => {
+			return amount - (includeFees ? Math.ceil(feePPK / 1000) : 0);
+		};
+		// "Cost" is the excess over target, plus fees, plus a PPK "tiebreaker"
```
a1denvalu3: Can't the cost just be `selectionSum - amountToSend`? This would already incorporate fees as a penalty. Since we know from greedy selection that the selection is at least >= amountToSend (for the `!exactMatch` case)

src/CashuWallet.ts
```diff
@@ -383,54 +385,218 @@ class CashuWallet {
...
+					amount = tempAmount + q.amount;
+					feePPK = tempFeePPK + (proofToFeePPK.get(q) ?? 0);
+					others.splice(qIndex, 1);
+					insertSorted(others, p);
+				}
+			}
+
+			// Update best solution
+			const cost = calculateCost(amount, feePPK, S.length);
+			if (cost < bestCost) {
```
a1denvalu3: in the `exactMatch` case, don't we want to consider the "error" (`amountToSend - selectionSumExFees`) rather than the cost for the update here?

src/CashuWallet.ts
robwoodgate: For exact match, it is not guaranteed to be over... in fact, it will usually be under as pure RGLI seeks to "get to" the amount - as expressed by  `if (exactMatch && netSum > amountToSend) break;` (Line 522 of code, which is line 6 of the algorithm in the paper). This is then improved upon in phase 2, but there is no guarantee that the "best" solution will get us to the exact match.

We might exhaust all trials and end up with a "best" solution that still doesn't meet our criteria.

So we need to handle "solutions" that undershoot (`netSum < amountToSend`) and we penalise them with `Infinity` cost. Basically, any exact match  that doesn't hit target.

For close match, we are guaranteed to be over, so that's where....

The `excess` calculation tells us how far from target we are.
The `feeCost` helps us tiebreak lower fee solutions for same `excess` solutions.
The `feePPK / 1000 ` is a fine grained tiebreaker to prioritize low fee keysets (eg: minimal fee selection test).

The latter two tests help us lower costs between exact match solutions.
The calculation is only performed once per trial, so is at most O(MAX_TRIALS). It's not a bottleneck as far as I can tell.

src/CashuWallet.ts
robwoodgate: I may be wrong, but the way I see it, calculateCost considers the total amount of units (eg sats) sent.... first by `excess` (`selectionSumExFees - amountToSend`), with finer grained distinctions for `feeCost` and `feePPK`.So `amountToSend - selectionSumExFees` is covered, with finer distinctions.

In other words, the algo seeks to minimize the total number of units sent, considering costs. If it makes sense to send more units but pay fewer fees, the algo will prioritize that... even if it means more units are sent.

In the exact match case, `excess` is zero, so if a solution is found with lower `feeCost` or `ppkFee` then that will lower the total amount (inc fees) sent.

src/CashuWallet.ts
robwoodgate: The idea here is to return pure RGLI... ie an exact match or best match within the  `MAX_OVRPCT` / `MAX_OVRAMT` params, as soon as possible if fees are not an issue or the proof set `S` is larger than `MAX_PROOFS`.

This aligns with lines 28-30 of the algo in the paper, which returns immediately if the exact match delta is zero.

In other words, in our modified RGLI, we will continue trying to find a lower "cost" solution if we have a smaller proof set, but will return immediately (as per RGLI) if we have found a solution in a fee agnostic or high proof set case. Does that make sense?

IE, we can afford to run more trials to lower costs if that's important (`includeFees`) and the subset is small enough that user will not notice

robwoodgate:
> concept-ACK. Still some things that are not clear to me.

Appreciate you helping me get this polished, your input has already made this at least 10x faster than my initial concept code 💪🙌

src/CashuWallet.ts
```diff
@@ -383,54 +385,218 @@ class CashuWallet {
...
-			const { keep, send } = this.selectProofsToSend(
-				smallerProofs.slice(1),
-				remainder,
-				includeFees
-			);
-			selectedProofs.push(...send);
-			returnedProofs.push(...keep);
-		}
+		// Precompute max acceptable amount for non-exact matches
+		const maxOverAmount = Math.min(
```
robwoodgate, line 525: This is a bug... should be:

```js
const maxOverAmount = Math.min(
    Math.ceil(amountToSend * (1 + MAX_OVRPCT / 100)),
    amountToSend + MAX_OVRAMT,
    sumExFees(totalAmount, totalFeePPK)
);
```

src/CashuWallet.ts
robwoodgate: Ahhh, I think I see now what you are saying... we could simplify to:

```js
// "Cost" is the excess over target, plus fees
// with tiebreaker to favour lower PPK keysets
const calculateCost = (amount: number, feePPK: number, length: number): number => {
    const netSum = sumExFees(amount, feePPK);
    if (netSum < amountToSend) return Infinity; // no good
    return amount - amountToSend + feePPK / 1000;
};
```

As you rightly say, amount includes fees... so we just need the tiebreaker for low fee keysets (minimal fee selection test).

Is that what you meant?

src/CashuWallet.ts
robwoodgate: Fixed

src/CashuWallet.ts
robwoodgate: Have implemented change

src/CashuWallet.ts
robwoodgate: For example, in the `optimal offline coinselection with 600 ppk input fees` test, the algo can find the optimal `[16,16,1]` if allowed to iterate on costs, but will often return `[16,16,1,1]` if it is in strict RGLI.

Both solutions achieve `amountToSend = 31`.

`16, 16, 1 + 3*600ppk = 33 - 2 = 31 net`
`16, 16, 1, 1 + 4*600ppk = 34 - 3 = 31 net`

robwoodgate:
Hey @lollerfirst - I've squashed a couple of bugs, and the algo gets pretty close to DP now. I've pushed a dev branch with the comparison algo [here](https://github.com/robwoodgate/cashu-ts/blob/select-proofs-dev/src/CashuWallet.ts#L397) so you can check out the benchmarks.

```
totalAmount: 7643591
N Proofs: 30
target amount to send: 6739242
selectProofs-DP: send.length = 24
selectProofs-DP: amountSend = 6739257
selectProofs-RGLI: send.length = 24
selectProofs-RGLI: amountSend = 6739321

amountToSend-RGLI relative error: 9.49650566e-6


totalAmount: 8581193
N Proofs: 30
target amount to send: 6358740
selectProofs-DP: send.length = 21
selectProofs-DP: amountSend = 6358753
selectProofs-RGLI: send.length = 21
selectProofs-RGLI: amountSend = 6358795

amountToSend-RGLI relative error: 6.60502501e-6

totalAmount: 8554011
N Proofs: 30
target amount to send: 4716300
selectProofs-DP: send.length = 14
selectProofs-DP: amountSend = 4716309
selectProofs-RGLI: send.length = 16
selectProofs-RGLI: amountSend = 4716377

amountToSend-RGLI relative error: 1.44178466e-5
```

Run times are also still excellent:

```
totalAmount: 2629197552
N Proofs: 10000
selectProofs-fees-closest: 67.034ms
selectProofs-fees-exact: 77.331ms

totalAmount: 2657471751
N Proofs: 10000
selectProofs-fees-closest: 71.717ms
selectProofs-fees-exact: 32.698ms

totalAmount: 2625468238
N Proofs: 10000
selectProofs-fees-closest: 38.566ms
selectProofs-fees-exact: 34.831ms
```

src/CashuWallet.ts
robwoodgate: [This comment](https://github.com/cashubtc/cashu-ts/pull/314/files/67b9328e279e774ffe4c8d8d98b1c5ef606c8b00#r2145426395) also applies here. The cost factor helps the algo pick the "right" option.

src/CashuWallet.ts
```diff
@@ -383,54 +385,232 @@ class CashuWallet {
...
-		if (!smallerProofs.length && nextBigger) {
-			return {
-				keep: proofs.filter((p: Proof) => p.secret !== nextBigger.secret),
-				send: [nextBigger]
-			};
+		// Init vars
+		const MAX_TRIALS = 60; // 40-80 is optimal (per RGLI paper)
+		const MAX_PROOFS = 100; // Strict RGLI will apply over this amount
+		const MAX_OVRPCT = 1; // Acceptable close match overage (percent)
+		const MAX_OVRAMT = 1024; // Acceptable close match overage (absolute)
```
robwoodgate: Do we want to expose the configuration options, or keep them private for now?
If expose, perhaps as some kind of RGLI options type that can be passed in as a single param?

Pro: Gives wallets a lot of freedom to tweak
Con: More API to maintain

src/CashuWallet.ts
robwoodgate: Maybe, instead of iterating again to hit a lower overall cost, we can do a quick check to see if removing the lowest contribution proof gives a better calculateDelta. I'll test it.

robwoodgate: Removed MAX_PROOFS in favour of a fee minimization check.

src/CashuWallet.ts
robwoodgate: Function is now called `calculateDelta()` - it's job is to signal the strength of the _solution_. It is not used to home in on one. That's the job of the local improvement phase (Phase 2).

For exactMatch, there is no compromise - per conversations with @Egge21M - we should get an exact match or fail. So there is no negative delta ("error")... if it falls short, it's an "Infinity" delta solution (ie no good).

robwoodgate:
Thanks again for flagging those areas @lollerfirst. As usual, your spidey-senses were spot on. Have now removed the MAX_PROOFS constant, so the only deviation from the paper is now a final fee check to make sure we haven't overpaid.

I think all the kinks are now ironed out, but if you find anything else, please shout.

lollerfirst:
Yes I'll take another look at it!

lollerfirst:
Very nice work! I don't think I have anything else to add at this point.

src/CashuWallet.ts
a1denvalu3: I think keep them interal.

robwoodgate: I agree

robwoodgate:
Rebased to the [logger PR](#324), as this PR now uses logger functionality.

Please let me know how you’d like to proceed:
- If you merge the logger PR (#324) first, I will rebase this PR back onto the `development` branch
- Alternatively, merging this PR will include the logger changes, allowing us to close #324.

robwoodgate:
Rebased to logger branch again following @Egge21M POJO tweak and my ConsoleLogger tweak. 

Can proceed to merge [as before](https://github.com/cashubtc/cashu-ts/pull/314#issuecomment-3033638430) or any other way you wish.

robwoodgate:
Rebased back to development as #324 was merged first.

Egge21M:
> @robwoodgate We should modify the random shuffle to be based on a `seed`, and the `seed` to be derived from the given proofs.
> 
> Reason: since [NUT-19](https://github.com/cashubtc/nuts/blob/main/19.md) gives idempotency for the `swap` `mint` and `melt` endpoints so if any network related error interrupt the request, you can easily retrieve the output ecash by firing an **IDENTICAL** one.
> 
> This isn't possible if the shuffle is seeded randomly.

I think the idea of "deterministic" randomness is great regardless of nutand could come in useful.

robwoodgate:
> @robwoodgate We should modify the random shuffle to be based on a `seed`, and the `seed` to be derived from the given proofs.
> 
> Reason: since [NUT-19](https://github.com/cashubtc/nuts/blob/main/19.md) gives idempotency for the `swap` `mint` and `melt` endpoints so if any network related error interrupt the request, you can easily retrieve the output ecash by firing an **IDENTICAL** one.
> 
> This isn't possible if the shuffle is seeded randomly.

I see the concern, but wouldn't it be the job of the `swap` / `send` methods to capture the `sendProofs`/ `sendProofsOnline` OUTPUT of `selectProofsToSend` in some way to protect against this issue?

They are closer to the mint operation edge, the `selectProofsToSend` function should be agnostic to NUT-19, mints and anything other than doing it's job of selecting proofs.

Am also not 100% sure that making the shuffle deterministic will help the algorithm... as it removes the `R` in `RGLI`. That said, I do see a logic to it if `selectProofsToSend` may be used in other scenarios.

Egge21M:
I generally agree with @robwoodgate here. I see your point @lollerfirst, but I believe the caching should be done on a different level (like the request instance of CashuMint), as this is the only way to ensure idempotency.

E.g. if we talk about a `swap` having "deterministic" randomly proof selection means inputs are reproducible, but outputs would still differ (as secret generation is not reproducible)

robwoodgate:
> The randomness is still there, only if you repeat the algorithm with the same inputs you get the same randomness.

Understood - on the upside it makes benchmark tests more reliable and predictable.

Adding a PRNG seems to add about 80-100ms on selection in a 50k proof set.... about 30% overhead.

Here's a [demo branch with console timings](https://github.com/robwoodgate/cashu-ts/blob/select-proofs-seeded/src/CashuWallet.ts#L502).

If we think it's a flyer, I'll push a commit to this PR branch too.

Issue #318, opened 2025-06-18
Ensure receiver's fee when an online-swap is necessary


lollerfirst:
When we want to `send` some ecash and want the receiver to be able to receive it whole, we want to include the swap fees in the selected amount.

This is trivial in the case we have the exact notes for an offline swap: `selectProofsToSend` with the `includeFees` flag is enough to get the job done.

However, in the case it doesn't find an exact notes selection, the wallet needs to perform an **online swap** with the current following logic:

1. `selectProofsToSend` for `amountToSend` with `includeFees` which results in `{ sendProofs, keepProofs }` (imprecise selection)
2. Given `sendProofs`, create an output amount split `{ keepAmounts, sendAmounts }` for which `sum(sendAmounts) >= amountToSend +  receiverFee(sendAmounts)`.
3. *(Paraphrased)* `inputs = sendProofs`, `outputs = {keepAmounts, sendAmounts}`.
4. Create swap payload and swap.

Now step 2 is not guaranteed to succeed because `sum(inputs) - senderFee(inputs)` might not be enough to satisfy `sum(sendAmounts) >= amountToSend + receiverFee(sendAmounts)`.

I propose a new logic:
1. Given `amountToSend` create the output amount split `sendAmounts` for which `sum(sendAmounts) >= amountToSend + receiverFee(sendAmounts)`.
2. `selectProofsToSend` for `sum(sendAmounts)`  with `includeFees` which results in `{sendProofs, keepProofs}`
4. Given `sendProofs` and `sendAmounts`, it is strictly the case that `sum(sendAmounts) <= sum(sendProofs) - senderFee(sendProofs)`. So, create the output split `keepAmounts` such that `sum(sendAmounts) + sum(keepAmounts) == sum(sendProofs) - sendersFee(sendProofs)`. 
5. *(Paraphrased)* `inputs = sendProofs`, `outputs = {keepAmounts, sendAmounts}`.
6. Create swap payload and swap.

gudnuf:
Yea this is needed. Your logic seems sound.

For reference, this is what I am currently doing when sending ecash with the receiver's fee included: [`prepareProofsAndFee`](https://github.com/MakePrisms/boardwalkcash/blob/master/app/features/send/cashu-send-swap-service.ts#L325-L425).

Seems you already account for this in your logic, but I will point it out anyways: 
[...]

PR #319, opened 2025-06-20
New online swap order

a1denvalu3:
# Fixes: #318 

## Description

Change the order by which we `selectProofsToSend` and select the amounts in the request output such that the input selection guarantees we can always select outputs that include the receiver's fee.

## Changes

- `send` - passes the entire coin-set to `swap` in case an online-send is in order, without doing coinselection first.
- `swap` - Now does coinselection. It does so after having decided the output amount split, which includes the receiver's fee.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/test/wallet.test.ts
+++ b/test/wallet.test.ts
@@ -676,6 +676,75 @@ describe('send', () => {
 
 		expect(result).toEqual(new Error('bad response'));
 	});
+	test('test send preference with fees included', async () => {
+		server.use(
+			http.get(mintUrl + '/v1/keysets', () => {
+				return HttpResponse.json({
+					keysets: [{ id: '009a1f293253e41e', unit: 'sat', active: true, input_fee_ppk: 600 }]
+				});
+			})
+		);
+		server.use(
+			http.post(mintUrl + '/v1/swap', () => {
+				return HttpResponse.json({
+					signatures: [
+						{
+							id: '009a1f293253e41e',
+							amount: 1,
+							C_: '021179b095a67380ab3285424b563b7aab9818bd38068e1930641b3dceb364d422'
+						},
+						{
+							id: '009a1f293253e41e',
+							amount: 2,
+							C_: '021179b095a67380ab3285424b563b7aab9818bd38068e1930641b3dceb364d422'
+						},
+						{
+							id: '009a1f293253e41e',
+							amount: 2,
+							C_: '021179b095a67380ab3285424b563b7aab9818bd38068e1930641b3dceb364d422'
+						},
+						{
+							id: '009a1f293253e41e',
+							amount: 2,
+							C_: '021179b095a67380ab3285424b563b7aab9818bd38068e1930641b3dceb364d422'
+						}
+					]
+				});
+			})
+		);
+		const wallet = new CashuWallet(mint, { unit });
+		await wallet.getKeys();
+
+		const overpayProofs = [
+			{
+				id: '009a1f293253e41e',
+				amount: 1,
+				secret: '1f98e6837a434644c9411825d7c6d6e13974b931f8f0652217cea29010674a13',
+				C: '034268c0bd30b945adf578aca2dc0d1e26ef089869aaf9a08ba3a6da40fda1d8be'
+			},
+			{
+				id: '009a1f293253e41e',
+				amount: 8,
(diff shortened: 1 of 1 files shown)


src/CashuWallet.ts
```diff
@@ -515,12 +501,27 @@ class CashuWallet {
...
+
+		// include the fees to spend the the outputs of the swap
+		// input selection, needs fees because of the swap
+		const { keep: keepProofs, send: sendProofs } = this.selectProofsToSend(
+			proofs,
+			amountToSend,
+			true
+		);
+
+		const amountToKeep = sumProofs(sendProofs) - this.getFeesForProofs(sendProofs) - amountToSend;
```
gudnuf, line 514: Does `sumProofs(sendProofs)` not already include `amountToSend`.

amountToSend = amount + receiverFee
sumProofs(sendProofs) = amountToSend + getFees(proofs)
this.getFeesForProofs(sendProofs)  = getFees(sendProofs)
amountToSend = amount + getFees(sendProofs)

gudnuf, line 514: Does sumProofs(sendProofs) not already include amountToSend.

amountToKeep = sum(proofs) - amount - sendSwapFee - receiveSwapFee

amountToSend = amount + receiveSwapFee

amountToKeep = sum(proofs) - amountToSend - sendSwapFee

sumProofs(sendProofs) = amountToSend + getFees(proofs)
this.getFeesForProofs(sendProofs) = getFees(sendProofs)
amountToSend = amount + getFees(sendProofs)

gudnuf, line 499: should `sumProofs(sendProofs)` be changed to `sumProofs(proofs)`?

Amount to keep should be the total fees (send and receive) and the requested amount subtracted from the total input amount. ie. `amountToKeep = sum(inputProofs) - (amount + receiveSwapFee) - sendSwapFee`

a1denvalu3: Let's break it down:
- We first select the output amounts `sendAmounts`, which might include receiver's fees. So `sum(sendAmounts) == amountToSend + receiverFees(sendAmounts)`.
- We then do coinselection with the **all** of the `proofs` for a target amount of `targetAmount == sum(sendAmounts)`  and **include** the sender's fee in the coinselection with the `includeFees` flag.

So now we have `{sendProofs, keepProofs}`, of which the only inputs to the swap are `sendProofs`.
`sendProofs` are necessarily greater or equal to `targetAmount`, which is the amount we did coinselection for. In fact, if the coinselection is successful, it should contain proofs whose amounts sum to: `amountToSend + receiverFees(sendAmounts) + senderFees(sendProofs) + excess`.

If `excess > 0` then we would like to capture that into `amountToKeep`, which is the amount we want to keep from the online swap. So then we use `amountToKeep` to decide the `keepAmounts` part of the outputs in the swap payload.

Now let's analyze the line:
```typescript
const amountToKeep = sumProofs(sendProofs) - this.getFeesForProofs(sendProofs) - amountToSend; 
```
I think the confusing part here is that the code updates `amountToSend` instead of defining `targetAmount` like I did in this breakdown. So I'll prefix the original `amountToSend` with (original):

1. `amountToSend ` $\rightarrow$ `targetAmount == sum(sendAmounts) == (original) amountToSend + receiverFees(sendAmounts)`

2. `sumProofs(sendProofs)` $\rightarrow$ `sum(sendProofs) == (original) amountToSend + receiverFees(sendAmounts) + senderFees(sendProofs) + excess`

3. `this.getFeesForProofs(sendProofs)` $\rightarrow$ `senderFees(sendProofs)`.

If you subtract (1) and (3) from (2) you'll see how all you're left with is `excess`.

robwoodgate (approved):
ACK - have suggested a specific unit test for this change, which it passes

robwoodgate:
One minor thing... should we not keep the 'proofsWeHave' option in `wallet.send()` and just pass it through to `wallet.swap()`?

gudnuf:
> @robwoodgate would you like to open a PR to this branch with the unit test in question?

Looks like he did https://github.com/lollerfirst/cashu-ts/pull/2

gudnuf (approved):

lollerfirst:
@robwoodgate re-added

PR #324, opened 2025-06-26
Adds Logger Interface and ConsoleLogger class to make logging opt-in and flexible

robwoodgate:
# Fixes: #205, Replaces: #253 

## Description

This PR refactors the logging functionality in the cashu-ts library to make it opt-in rather than always logging to the console. 

It abstracts logging to a new `Logger` interface, and implements a default `NULL_LOGGER` and optional `ConsoleLogger` class for convenience.

It offers multi-level logging, with the ability to set minimum logging levels. It also offers interpolation of log messages.

Example:
```
const logger = new ConsoleLogger(LogLevel.DEBUG);
logger.info('User {username} logged in', { username: 'alice', ip: '127.0.0.1' });
// Output: [INFO] User alice logged in { ip: "127.0.0.1" }
```

This allows the library to implement / retain logging calls, which can then be turned on as required by developers. Developers can use the provided `ConsoleLogger` for convenience, or wrap their preferred logger in an adapter class to conform to the `Logger` interface.

At this stage, I've just replaced all existing `console` functions, but going forward, the library can easily add other logging statements that will be useful to developers.

## Changes

- Added `Logger` interface, `ConsoleLogger` class and default `NULL_LOGGER`
- Added logger option to CashuWallet, CashuMint, and supporting classes
- Replaced all direct console logging calls with logger interface aware calls
- Added `setRequestLogger` function to `request.ts` module to allow requests logging
- Added logging documentation in the README
- Adds unit test suite for the provided `ConsoleLogger` and `NULL_LOGGER`

**NOTE:** The addition of the logger class extends the public API, but in a backwards compatible way.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


robwoodgate:
> Awesome! This supersedes #253 Will look at it asap.
> 
> Edit: oops, just saw that you mentioned that in the heading :D

Yes, I saw your reply to #253, and agreed with your reasoning for a Logger interface. So I wrote this one... which gives a bit of a nod to the PSR-3 / RFC5424 spec, but is tailored to the TS/JS environment (fewer levels).

Should be fairly easy for developers to wrap loggers like Winston etc too if they prefer it.

robwoodgate:
Finally got there... LGTM!

Egge21M:
LGTM. I added a tiny commit, replacing the enum with a plain javascript object + type, which is in line with the rest of cashu-ts. This allows for duck-typing (using LogLevel.FATAL or just "FATAL"), which is more in line with the general typescript style.

PTAL @robwoodgate and we should be good to merge this. Thank you very much for taking care of this.

robwoodgate:
@Egge21M - nice addition, it makes sense. I've just pushed a tweak to ConsoleLogger to work with the new POJO+Type as the enum provided number mapping before. Happy to roll with this now.

PR #325, opened 2025-07-01
Feat/improve auth support

gudnuf:
## Description

All endpoints that requires blind auth may also require clear auth.  This PR makes it so a client can properly handle mints that have protected endpoints.

## Changes

- added NUT-21 clear auth to `MintInfo`
- exposed NUT-22 blind auth in `MintInfo` so  we can know if the mint requires BATs and how many we can mint
- modified `CashuMint` to include a `clearAuthGetter` to add the CAT to any requests to endpoints protected by clear auth instead of just including BATs. To do this I refactored a bit by making a private `getAuthHeaders` method

## PR Tasks

- [x] Open PR
- [ ] run `npm run format`


etc/cashu-ts.api.md
gudnuf, line 1: I ran `npm run api:check` then `cp temp/cashu-ts.api.md etc/cashu-ts.api.md`.  @Egge21M what else is needed for the api extractor test to pass? The error is not very indicative and I can't find any documentation here
image

Egge21M: I will add a guide to the contribution section today.
If you changed the public API you want to compile the lib and then commit the updated `etc/cashu-ts.api.md` file with your PR. This ensures that reviewers can spot an API change at the first glance. The CI will generate its own API docs based on the compiled lib and throw if it does not match the one present in `etc/cashu-ts.api.md`.
The flow is:

```sh
npm run compile
npm run api:check
# if check let's you know that the API has changed, run the following
npm run api:update
git add ./etc/cashu-ts.api.md
git commit -m "chore: update API docs via API Extractor"
```

Egge21M: Update: Please rebase this to the latest start of `development` and read `CONTRIBUTING.md` for a short guide on how to use api-extractor

gudnuf: cool thanks, it works now.

src/CashuMint.ts
```diff
@@ -570,18 +592,27 @@ class CashuMint {
...
+		}
+
 		if (info.requiresBlindAuthToken(path)) {
-			if (!this._authTokenGetter) {
-				throw new Error('Can not call a protected endpoint without authProofGetter');
+			if (!this._blindAuthTokenGetter) {
+				throw new Error('Can not call a protected endpoint without blindAuthTokenGetter');
 			}
-			return this._authTokenGetter();
+			const blindAuthToken = await this._blindAuthTokenGetter();
```
gandlafbtc, line 644: not awaiting this was a bug, right?

gudnuf: apparently so, the original type of authTokenGetter also returned a promise

PR #326, opened 2025-07-02
Streamline API-Extractor flow

Egge21M:
# Fixes: https://github.com/cashubtc/cashu-ts/issues/321

## Description

This PR streamlines the API Extractor flow for CI/CD.

## Changes

- Adds local reports to `.gitignore` so they are not commited by accident.
- Adds build script to API report scripts to make sure the report is always generated from the latest output.
- Adds contribution guide.


robwoodgate (approved):
This is a great improvement, as API check relies on a compile... I've lost count of the number of times I've run the API check and forgotten to compile first. LGTM.

PR #327, opened 2025-07-10
Revert change to mint/melt quote public API 

Egge21M:
# Fixes: https://github.com/cashubtc/cashu-ts/issues/323

## Description

In the latest release we introduces a change to the mint and melt operations that broke the public API. This change reverts that, while keeping the same functionality. It does this by making sure the old type definition is the same again, and introducing a narrower type that can be assigned without any issues.

```ts

const mintQuote = await wallet.createMintQuote(21); // <- Returns FullMintQuoteResponse
const mintQuote2: MintQuoteReponse = await wallet.createMintQuote(21); // This also works, as FullMintQuoteReponse is narrower

mintQuote.amount // <- this is still "number" not "number | undefined"
```

## Changes

- Revert change to `MintQuoteReponse` and `MeltQuoteReponse`
- Add new `FullMintQuoteReponse` and `FullMeltQuoteReponse` types, that represent responses without optional keys.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`
- [x] run `npm run api:update`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -32,7 +32,7 @@ import type {
 	MeltQuoteOptions,
 	SwapTransaction,
 	LockedMintQuoteResponse,
-	PartialMintQuoteResponse
+	FullMintQuoteResponse
 } from './model/types/index.js';
 import { MintQuoteState, MeltQuoteState } from './model/types/index.js';
 import { SubscriptionCanceller } from './model/types/wallet/websocket.js';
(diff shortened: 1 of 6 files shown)


gandlafbtc (approved):

src/CashuWallet.ts
```diff
@@ -688,7 +688,7 @@ class CashuWallet {
 		amount: number,
 		pubkey: string,
 		description?: string
-	): Promise<LockedMintQuoteResponse> {
+	): Promise<LockedMintQuoteResponse & { amount: number; unit: string }> {
```
gudnuf, line 691: Should we also just define this as `FullLockedMintQuoteResponse`

gudnuf (approved):
ack, just one little nit

src/CashuWallet.ts
gandlafbtc: Fair point. Type starts to look a bit ridiculous, but I think using a type is more expressive than hacking it with `&`

gandlafbtc: @Egge21M You wanna change it, or merge as is

Issue #328, opened 2025-07-10
Implement Caching layer for NUT-19


Egge21M:
https://github.com/cashubtc/nuts/blob/main/19.md allows wallets to replay mint operations in case of loss of network / state. Cashu-TS does not yet support this.

We do however expose the request instance in the Cashu mint, so theoretical a layer that caches requests could be built on that.

Egge21M:
@d4rp4t will take this on :)

robwoodgate:
Related: #331

PR #330, opened 2025-07-20
Nut19 cache

d4rp4t:
# Fixes: #328

## Description

Adds internal support for retrying cached requests according to NUT-19, based on exponential backoff.

### request.ts
- extended request() with optional retry logic
- added requestWithRetry() with exponential backoff
### MintInfo.ts
- added support for NUT-19 via checkNut19() method
### Other
- added Nut19Policy type

Note: Cached request usage is not yet implemented in the CashuMint class — waiting for approval.
## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/MintInfo.ts
+++ b/src/model/MintInfo.ts
@@ -123,7 +123,7 @@ export class MintInfo {
 
 	private checkNut19() {
 		const rawPolicy = this._mintInfo.nuts?.[19];
-		if (rawPolicy && rawPolicy.cached_endpoints.length > 0) {
+		if (rawPolicy && (rawPolicy?.cached_endpoints?.length || 0) > 0) {
 			return {
 				supported: true,
 				params: {


gandlafbtc:
thanks for the PR @d4rp4t ! do you think this could be tested with an integration test? This would not only verify it works as expected, it also gives implementors an up to date example.

src/model/types/mint/responses.ts
```diff
@@ -309,3 +310,11 @@ export type WebSocketSupport = {
...
 	signatures: Array<SerializedBlindedSignature>;
 } & ApiError;
+
+/**
+ *  Caching policy of mint endpoints
+ */
+export type Nut19Policy = {
+	ttl: number | null;
+	cached_endpoints: Array<{ method: 'GET' | 'POST'; path: string }>;
+} | null;
```
robwoodgate: Wondering if this type definition is a little too loose...? I don't love allowing it to be `null` overall.
Also, athough `null` ttl is a valid mint response, it actually means `no expiry` in the spec (ie: `Infinity`). So perhaps we can make it tighter to make that clear and avoid repeated null checking?
What do you think of this version?

```ts
/**
 * Represents the NUT-19 caching policy as advertised in the mint's info response.
 * This defines how long and for which endpoints responses can be cached.
 *
 * - `ttl` is always a finite number (seconds) or `Infinity` (indefinite caching).
 * - An empty `cached_endpoints ` array means no endpoints are cached at the mint.
 */
export type Nut19Policy = {
    ttl: number;
    cached_endpoints: Array<{ method: 'GET' | 'POST'; path: string }>;
};
```

src/model/MintInfo.ts
```diff
@@ -92,6 +95,17 @@ export class MintInfo {
...
+		if (this._mintInfo.nuts[19] && this._mintInfo.nuts[19]?.cached_endpoints.length > 0) {
+			return {
+				supported: true,
+				ttl: this._mintInfo.nuts[19].ttl,
+				params: this._mintInfo.nuts[19].cached_endpoints
+			};
+		}
+		return { supported: false };
+	}
+
```
robwoodgate, line 201: This looks solid - requiring at least one cached endpoint for the supported flag makes sense. If we adopt a stricter NUT19Policy type definition as per my suggestion, we can map `null` to `Infinity` for indefinite caching. We could also make it easier for implementors to use the policy in the request without converting it, eg:

```ts
private checkNut19() {
  const rawPolicy = this._mintInfo.nuts[19];
  if (rawPolicy && rawPolicy.cached_endpoints.length > 0) {
	  return {
		  supported: true,
		  params: {
			  ttl: rawPolicy.ttl ?? Infinity,
			  cached_endpoints: rawPolicy.cached_endpoints
		  } as NUT19Policy
	  };
  }
  return { supported: false };
}
```

We can then import NUT19Policy and add a specific `isSupported` signature for this nut around line 28, eg:

```ts
isSupported(num: 19): { supported: boolean; params?: NUT19Policy };
```

src/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
+ */
+async function requestWithRetry(options: RequestOptions): Promise<unknown> {
+	const { ttl, cached_endpoints, endpoint } = options;
+
+	const url = new URL(endpoint);
+
+	const isCachable = cached_endpoints?.some(
+		(cached_endpoint) =>
+			cached_endpoint.path === url.pathname && cached_endpoint.method === (options.method || 'GET')
+	);
```
robwoodgate: Should we make the default handling more explicit with `??` - ie just `null` or `undefined`?
Would surface bugs sooner if a caller accidentally passes an invalid but falsey method (eg ''. false).

```ts
... && cached_endpoint.method === (options.method ?? 'GET')
```

src/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
-	requestBody,
-	headers: requestHeaders,
-	...options
-}: RequestOptions): Promise<unknown> {
+/**
+ * Internal function that handles retry logic for NUT-19 cached endpoints.
+ * Non-cached endpoints are executed directly without retries.
+ */
+async function requestWithRetry(options: RequestOptions): Promise<unknown> {
+	const { ttl, cached_endpoints, endpoint } = options;
```
robwoodgate, line 39: Memo: references to `cached_endpoint` option in this function would need updating to `cachedEndpoint` if we adopt the camelCase suggestion in Nut19Policy.

robwoodgate (commented):
Great work @d4rp4t! Love the use of exponential backoff, and the implementation looks solid (eyeballed: have not had a chance to test). I've noted a couple of minor nits for your consideration.

src/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
+		} catch (e) {
+			if (e instanceof NetworkError) {
+				const totalElapsedTime = Date.now() - startTime;
+				const shouldRetry = retries < MAX_CACHED_RETRIES && (!ttl || totalElapsedTime < ttl);
+
+				if (shouldRetry) {
+					retries++;
+					const delay = Math.max(Math.pow(2, retries) * 1000, 1000);
+
+					if (ttl && totalElapsedTime + delay > ttl) {
```
robwoodgate: Suggested addition:

```ts
requestLogger.error('Network Error: request abandoned after #{retries} retries', { e, retries });
```

src/transport/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
+				const shouldRetry = retries < MAX_CACHED_RETRIES && (!ttl || totalElapsedTime < ttl);
+
+				if (shouldRetry) {
+					retries++;
+					const delay = Math.max(Math.pow(2, retries) * 1000, 1000);
+
+					if (ttl && totalElapsedTime + delay > ttl) {
+						throw e;
+					}
+
```
robwoodgate, line 182: Suggested addition:

```ts
requestLogger.info('Network Error: attempting retry #{retries} in {delay}ms', { e, retries, delay });
```

src/transport/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
+					await new Promise((resolve) => setTimeout(resolve, delay));
+					return retry();
+				}
+			}
```
robwoodgate, line 186: Suggested addition:

```ts
requestLogger.error('Request failed and could not be retried', { e });
```

robwoodgate (commented):
Would be a good spot to leverage our new logger as well

src/transport/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
+	let retries = 0;
+	const startTime = Date.now();
+
+	const retry = async (): Promise<unknown> => {
+		try {
+			return await _request(options);
+		} catch (e) {
+			if (e instanceof NetworkError) {
+				const totalElapsedTime = Date.now() - startTime;
+				const shouldRetry = retries < MAX_CACHED_RETRIES && (!ttl || totalElapsedTime < ttl);
```
robwoodgate: There is a unit mismatch here... NUT-19 specifies TTL in seconds. We are calculating `totalElapsedTime` in MS. Same further down with `delay`. So perhaps we should convert TTL here and use `ttlMs` for clarity? Also gives us an opportunity to sanitize ttl (as it's user input).

```ts
const ttlMs = typeof ttl === 'number' && !isNaN(ttl) ? Math.max(ttl, 0) * 1000 : Infinity;
```

And update ttl references in rest of function

src/request.ts
```diff
@@ -28,12 +31,55 @@ export function setRequestLogger(logger: Logger): void {
...
+		try {
+			return await _request(options);
+		} catch (e) {
+			if (e instanceof NetworkError) {
+				const totalElapsedTime = Date.now() - startTime;
+				const shouldRetry = retries < MAX_CACHED_RETRIES && (!ttl || totalElapsedTime < ttl);
+
+				if (shouldRetry) {
+					retries++;
+					const delay = Math.max(Math.pow(2, retries) * 1000, 1000);
```
robwoodgate: We should increment retries after calculating delay, so delay starts at 1000ms, otherwise it starts at 2000ms.

```suggestion
					const delay = Math.max(Math.pow(2, retries) * 1000, 1000);
					retries++;
```

robwoodgate: On a side-note... even after this tweak, our retry schedule (in MS) is:
1000, 2000, 4000, 8000, 16000, 32000, 64000, 128000, 256000, 512000

That's 1023 seconds of delay... about 17 minutes. Is an app likely to be waiting around that long?

Perhaps we should cap the delay... eg: using a 60 second cap is about 5 minutes of retries:
1000, 2000, 4000, 8000, 16000, 32000, 60000, 60000, 60000, 60000

We should also consider adding a jitter factor to avoid the thundering herd problem.
This would also potentially cut down the overall wait time.

eg: 

```ts
// Assuming:
const MAX_RETRY_DELAY = 60000; // Added around line 11
const ttlMs = typeof ttl === 'number' && !isNaN(ttl) ? Math.max(ttl, 0) * 1000 : Infinity; // in line 62

// Here's what a capped exponential backoff with jitter could look like:
if (shouldRetry) {
    // Calculate capped exponential backoff using full jitter to avoid
    // the thundering herd problem
    const cappedDelay = Math.min(Math.pow(2, retries) * 1000, MAX_RETRY_DELAY);
    const delay = Math.random() * cappedDelay;

    if (totalElapsedTime + delay > ttlMs) {
        requestLogger.error('Network Error: request abandoned after #{retries} retries', { e, retries });
        throw e;
    }

    retries++;
    await new Promise((resolve) => setTimeout(resolve, delay));
    return retry();
}
```

src/transport/request.ts
a1denvalu3: @robwoodgate  you can add suggestions with CNTRL+G

robwoodgate (approved):
Great work @d4rp4t! 

Should we also add a test to `request.test.ts` for the exponential backoff?

etc/cashu-ts.api.md
```diff
@@ -1490,7 +1509,7 @@ export type RequestArgs = {
 export type RequestFn = <T = unknown>(args: RequestOptions) => Promise<T>;
 
 // @public (undocumented)
-export type RequestOptions = RequestArgs & Omit<RequestInit, 'body' | 'headers'>;
+export type RequestOptions = RequestArgs & Omit<RequestInit, 'body' | 'headers'> & Partial<Nut19Policy>;
```
Egge21M: `Nut19Policy` is not exported. Is that on purpose?

src/wallet/Wallet.ts
```diff
@@ -355,7 +356,16 @@ class Wallet {
 		}
 
 		// Go Mintinfo?
-		this.getMintInfo();
+		const mintInfo = this.getMintInfo();
+
+		// Set up NUT-19 cache retry if supported
+		const nut19 = mintInfo.isSupported(19);
+		if (nut19.supported && nut19.params) {
+			setGlobalRequestOptions({
```
Egge21M: setGlobalRequestOptions sets the request option on a module level scope (everything that is part of the cashu-ts). But here we are mixing it with mint level options. If someone has multiple mint connections running at the same time these options will collide

src/transport/request.ts
```diff
@@ -48,6 +51,72 @@ export function setRequestLogger(logger: Logger): void {
...
+
+					if (totalElapsedTime + delay > ttl) {
+						requestLogger.error('Network Error: request abandoned after #{retries} retries', {
+							e,
+							retries,
+						});
+						throw e;
+					}
+					retries++;
+					requestLogger.info('Network Error: attempting retry #{retries} in {delay}ms', {
```
Egge21M: I don't think our Logger supports string interpolation. Also Logger Interface does not require this behavior, please either use JS string interpolation or literals

test/transport/request.node.test.ts
```diff
@@ -111,4 +113,196 @@ describe('requests', () => {
...
+			};
+			let requestCount = 0;
+			server.use(
+				http.get(endpoint, () => {
+					requestCount++;
+					return Response.error();
+				}),
+			);
+
+			vi.spyOn(Math, 'random').mockReturnValue(1); // jitter off
```
Egge21M, line 313: Please restore this mock after the test ran

test/transport/request.node.test.ts
Egge21M: Pre-Existing, but just noticed. This will alter behavior for all tests and should be cleared.

src/transport/request.ts
```diff
@@ -48,6 +51,72 @@ export function setRequestLogger(logger: Logger): void {
 	requestLogger = logger;
 }
 
+/**
+ * Internal function that handles retry logic for NUT-19 cached endpoints. Non-cached endpoints are
+ * executed directly without retries.
+ */
+const MAX_CACHED_RETRIES = 10;
+const MAX_DELAY = 5000; // 5 secs
```
Egge21M: I think this might be way too much. Let's say a mint goes offline. Wallets would retry for ~26 seconds until an error is propagated to the caller.

src/transport/request.ts
robwoodgate: Correct - our beta v3 logger had interpolation but it was a under used and over complex vs regular js string interpolation, so was removed in v3 final

d4rp4t: gotcha, i just copy-pasted it from @robwoodgate code suggestion back in the days

src/transport/request.ts
d4rp4t: now max retry time is 6.5s with mocked jitter factor at 1, so it will be ~3.250 at average

minibits-cash: Small note from wallet maintainer: I believe >99% of unfulfilled requests are on client/device and network side. So immediate retry  won't usually resolve that and keeping timers alive for too long is risky.

Alt approach would be to persist request params over the request duration. Clean it on response or on not-network exceptions. Retry on some usage pattern (I currently do on next app foreground / home screen after >1 minute).

Another topic is whether to retry specifically melts with this approach, because of their async behavior.

robwoodgate: > Small note from wallet maintainer: I believe >99% of unfulfilled requests are on client/device and network side. So immediate retry won't usually resolve that and keeping timers alive for too long is risky.
> 
> Alt approach would be to persist request params over the request duration. Clean it on response or on not-network exceptions. Retry on some usage pattern (I currently do on next app foreground / home screen after >1 minute).
> 
> Another topic is whether to retry specifically melts with this approach, because of their async behavior.

Are you suggesting cashu-ts should leave retry to wallets and not attempt to retry NUT-19 cacheable endpoints?

minibits-cash: I see now that described approach is out of the scope of the library. I would say an optionality is OK here and the PR proposes to have retries opt-in.

robwoodgate (commented):
Looking at this with fresh eyes, one thing we have not considered is network latency and timeouts. Although the delay is ~3.25s, each request could take ~10s or more. 

So though I LOVE the implementation of exponential backoff, I'm wondering if it's at the wrong layer here in cashu-ts?

src/transport/request.ts
```diff
@@ -48,6 +51,72 @@ export function setRequestLogger(logger: Logger): void {
...
+
+					if (totalElapsedTime + delay > ttl) {
+						requestLogger.error(`Network Error: request abandoned after ${retries} retries`, {
+							e,
+							retries,
+						});
+						throw e;
+					}
+					retries++;
+					requestLogger.info(`Network Error: attempting retry ${retries} in {delay}ms`, {
```
robwoodgate:
```suggestion
					requestLogger.info(`Network Error: attempting retry ${retries} in ${delay}ms`, {
```

src/transport/request.ts
```diff
@@ -48,6 +51,72 @@ export function setRequestLogger(logger: Logger): void {
...
+	let retries = 0;
+	const startTime = Date.now();
+
+	const retry = async (): Promise<unknown> => {
+		try {
+			return await _request(options);
+		} catch (e) {
+			if (e instanceof NetworkError) {
+				const totalElapsedTime = Date.now() - startTime;
+				const shouldRetry = retries < MAX_CACHED_RETRIES && (!ttl || totalElapsedTime < ttl);
```
robwoodgate, line 162: ttl is already normalized to number (or infinity), so can just use it without the guard.

```suggestion
				const shouldRetry = retries < MAX_CACHED_RETRIES && totalElapsedTime < ttl;
```

src/transport/request.ts
```diff
@@ -48,6 +51,72 @@ export function setRequestLogger(logger: Logger): void {
...
+async function requestWithRetry(options: RequestOptions): Promise<unknown> {
+	const { ttl, cached_endpoints, endpoint } = options;
+
+	const url = new URL(endpoint);
+
+	// there should be at least one cached_endpoint, also ttl is already mapped null->Infinity
+	const isCachable =
+		cached_endpoints?.some(
+			(cached_endpoint) =>
+				cached_endpoint.path === url.pathname &&
```
robwoodgate: This is fine for POST requests (eg `/v1/mint/bolt11`), but if a mint can cache GET endpoints (like `/v1/mint/quote/bolt11/[:quoteid]`), then it will need to be an includes match?

src/transport/request.ts
```diff
@@ -48,6 +51,72 @@ export function setRequestLogger(logger: Logger): void {
...
+				cached_endpoint.method === (options.method ?? 'GET'),
```
robwoodgate: Should we normalize  to UPPERCASE or is a NUT-19 `cached_endpoint.method` and `options.method` guaranteed to be uppercase? If not, the strict equality will fail.

robwoodgate:
@codex review

src/transport/request.ts
robwoodgate: On reflection, GET is not really relevant for NUT-19 - minting, swapping, and melting are all POST

src/mint/Mint.ts
```diff
@@ -711,11 +711,13 @@ class Mint {
 			...(bat ? { 'Blind-auth': bat } : {}),
 			...(cat ? { 'Clear-auth': cat } : {}),
 		};
+		const nut19 = (await this.getLazyMintInfo()).isSupported(19);
```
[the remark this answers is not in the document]
robwoodgate: fixed in b1667153

src/transport/request.ts
```diff
@@ -48,26 +58,212 @@ export function setRequestLogger(logger: Logger): void {
...
+					callerSignal.removeEventListener('abort', forwardAbort);
+					timeoutController.signal.removeEventListener('abort', forwardAbort);
+				};
+			}
+			signal = combinedController.signal;
+		}
+	}
+
+	if (callerSignal?.aborted && !timeoutController?.signal.aborted) {
+		throw new CallerAbortError('Request aborted by caller');
```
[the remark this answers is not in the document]
robwoodgate: fixed in e40e0873

src/transport/request.ts
```diff
@@ -48,26 +58,212 @@ export function setRequestLogger(logger: Logger): void {
...
+	const { ttl, cached_endpoints, endpoint } = options;
+	const endpointPathname = getEndpointPathnameSafe(endpoint);
+
+	// there should be at least one cached_endpoint, also ttl is already mapped null->Infinity
+	const isCachable =
+		endpointPathname !== undefined &&
+		cached_endpoints?.some(
+			(cached_endpoint) =>
+				cached_endpoint.path === endpointPathname &&
+				cached_endpoint.method === (options.method?.toUpperCase() ?? 'GET'),
```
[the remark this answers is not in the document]
robwoodgate: fixed in f9acf8e1

robwoodgate (approved):
@d4rp4t - hope you don't mind, I fixed up the AbortController issue and a couple of nits highlighted by codex.

This looks really solid, you've done an amazing job on it.

LGTM!

d4rp4t:
@robwoodgate woah, thanks for fixing my mistakes. glad that we finally managed to finish it!

src/mint/Mint.ts
```diff
@@ -703,19 +711,22 @@ class Mint {
 		customRequest?: RequestFn,
 	): Promise<T> {
 		const requestInstance = customRequest ?? this._request;
+		const mintInfo = await this.getLazyMintInfo(customRequest);
```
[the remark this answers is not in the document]
robwoodgate: fixed in: b5c49c9a

src/model/MintInfo.ts
```diff
@@ -116,6 +121,25 @@ export class MintInfo {
 		return { supported: false };
 	}
 
+	private checkNut19() {
+		const rawPolicy = this._mintInfo.nuts?.[19];
+		if (rawPolicy && rawPolicy.cached_endpoints.length > 0) {
```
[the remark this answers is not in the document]
robwoodgate: fixed in: 6db0716e

PR #329, opened 2025-07-10
Add onClose method to WSConnection

gudnuf:
# Fixes: #317

## Description

...

## Changes

- ...
- ...

## PR Tasks

- [x] Open PR
- [x] run `npm run test` --> no failing unit tests
- [x] run `npm run format`

Issue #317, opened 2025-06-13
Add onClosed callback to subscription methods


jbojcic1:
Currently, all subscription methods that are subscribing to socket messages (onMintQuoteUpdates, onMeltQuoteUpdates, ...) allow you to pass callback (invoked when a message is received) and errorCallback (invoked when a subscription is attempted on a non-open socket or when a message received from the mint is an error. It would be great if we could also have onClosed callback that would be triggered when the socket is closed.

The issue I am trying to solve is following. I am subscribing to messages in a react app within `useEffect`. So `useEffect` calls `onMintQuoteUpdates` for example and each time it is called it cancels the previous subscription and creates a new one (btw if the socket was closed in the meantime, for whatever reason, `onMintQuoteUpdates` will create a new connection when creating a subscription). That will happen even if the quote ids haven't changed and if it happens often it causes me to get rate limited by the mint. Thus, I had to implement the logic that checks if the quote ids have actually changed and if not, don't even call `onMintQuoteUpdates`. However, this is problematic when I am running the app on the phone. When the app goes to the background, the phone closes the socket and when the app is back in the foreground, my logic won't call `onMintQuoteUpdates` again if the quote ids haven't changed so the app stops receiving updates. 
If there would be an onClosed callback on `onMintQuoteUpdates` then my app would get notified whenever the socket is closed which would then enable me to handle this case and make sure that the cache I have is busted, and next time the app is open, the new connection will be open and new subscription created. 

A specific case that I have with the socket closed due to the app going to the background could be solved by detecting that event and busting my cache, but there are other reasons why the socket could be closed. For example, it could be because the internet connection was lost, so I would need to handle that case as well. It is much simpler if I can just be notified when the underlying socket is closed than trying to predict every possible thing that could cause it.

Another possible option, instead of adding onClosed callback to methods like `onMintQuoteUpdates`, would be to just expose onClose method on WSConnection so it could be accessed by doing `wallet.mint.webSocketConnection?.onClose`

gudnuf:
done with #317

PR #215, opened 2024-11-27
CashuWallet.ts' comment corrections

Gudnessuche:
# Fixes: #[]

## Description

Comments aren't 100% punctuated

## Changes

Made the comments more grammatically correct by adding proper punctuations as well as capitalization of words after a period.

This miniscule change is a perfectionist's view for the grammatical correctness that this file deserves.

PR #333, opened 2025-07-24
Fix linting errors

### LINKED ISSUE #337: Browser Tests are not actually running in a Browser env
A couple months ago we migrated to `vitest` to allow browser testing with `playwright`. As it turns out vitest's browser mode still runs all test files in node, but adds DOM capabilities.

This is pretty much useless for our case, as it does not catch the usage of non-browser APIs. We should therefore either have a separate test pipeline that ensures real browser tests OR add linting rules to forbid NodeJs-only APIs.

robwoodgate:
## Description

This PR enhances type safety, error handling, and code consistency by upgrading ESLINT to v9, Prettier to v3, and fixing all linting errors and warnings in the codebase.

It also adds a new "Lint Check" CI step so we can maintain code quality going forward.

Key focuses include replacing `"any"` with `"unknown"` for stricter typing, improving WebSocket/subscription logic, and minor optimizations. No new features; primarily maintenance.

## Changes

- Upgrades ESLINT to v9, adds new `eslint.config.js` file
- Upgrades prettier to v3, adds TSDOC formatting via `prettier-plugin-jsdoc`
- Upgrades vitest to v3, merges legacy `vitest.workspace.ts` into `vite.config`
- Fixes browser tests (uses chromium) via vitest projects. Adds `npm run test:prepare` script for chromium download
- Fixes issues highlighted by browser tests (buffer imports/polyfill)
- Adds Lint Check CI workflow.
- Adds `.editorconfig` to help ensure visual consistency (eg: on GitHub)
- Updates node engine to >= v22.4.x, uses v22 for CI workflows
- Ignore `/lib`, `/examples`, vite config and definition files in `eslint.config.js`
- Removes legacy `.eslintignore` and `.eslintrc.json` files
- Removes unused imports and variables
- Enforces inline type for import/export of types
- Switch `any` type to `unknown` (or a stricter type where available)
- Add generics and type parameters for better inference (e.g., in shuffleArray, addSubListener)
- Improve error handling with explicit throws, null checks, and logging (e.g., socket not open, binary search indices)
- Use `const` over `let` for immutability where possible
- Fix typos and deprecate old methods (e.g., closeSubscription → cancelSubscription)
- Refine CBOR encoding with defensive object checks
- Misc tweaks made by lint

## "Breaking" Changes

Nothing is strictly breaking, but there may be some "soft" breaking effects, including:

- **Type narrowing:** `any` → `unknown/stricter`. This improves type safety and catches potential errors earlier in development. However, it could cause compile-time warnings or errors for consumers if their code relied on the flexibility of `any` for custom behaviors (e.g., assigning incompatible values).
- **WSConnection.createSubscription:** Now throws 'Socket not open' instead of invoking errorCallback. This is in line with calling code, which also throws if there is no connection.
- **ConnectionManager.instance:** Privatized; consumers should be using getInstance() in any case.
- **WSConnection.closeSubscription:** Marked as deprecated; use cancelSubscription. Still functions meantime.
- **Error callbacks:** Now expects Error (wrapped); raw JSON-RPC properties must be unwrapped now.
- **WebSocket Implementation:** Check in ws.ts: `getWebSocketImpl` now explicitly throws if `_WS` is uninitialized (previously, it could return implicitly undefined, leading to later errors).
- **Node.js Engine Requirement**: Added `"engines": { "node": ">=22.4.0" }` to `package.json` for native fetch/Response/WebSocket support. This adds install-time warnings for older versions without blocking.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


robwoodgate:
Rebased to fix merge conflict with #329

robwoodgate:
Prettier now properly formats TSDOCs, which means #215 is no longer needed

robwoodgate:
Updated the node engine to >= v22.4.x, and CI workflows to v22.

Note that cashu-ts uses `CloseEvent` since #317, which is a node v23 feature.
Have not bumped node engine this far as it's not a core feature, but we can if appropriate.
Instead, the unsupported-features rule ignores `CloseEvent`.

.github/workflows/node.js.yml
```diff
@@ -11,13 +11,13 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [22.x]
```
gandlafbtc, line 14: I noticed we moved to the 22.x node environment everywhere. To ensure that the lib remains somewhat backwards compatible, should we maybe include other environments here? I guess that would open the question which versions we want to support

gandlafbtc (approved):
Great PR, this was very much needed!  I skimmed it, and mostly LGTM. The question about the version can be addressed here, outside of this PR:
https://github.com/cashubtc/cashu-ts/issues/339

.github/workflows/node.js.yml
Egge21M: Node 18 reached its End-Of-Life at the beginning of the year. I think its reasonable to move this goalpost with the current LTS. However our build pipeline targets es2020, which should be compatible with anything newer than node14.

Issue #332, opened 2025-07-24
Should we switch to extensionless imports?


robwoodgate:
I noticed that import statements in source files (e.g., CashuWallet.ts) include .js extensions, like:

```ts
import { serializeProof } from './crypto/client/index.js';
```

This obviously works fine, compiles with no issues etc. But perhaps we should consider using extensionless imports?

eg:

```ts
import { serializeProof } from './crypto/client/index';
```

Our `tsconfig.json` settings `(moduleResolution: "bundler", module: "ESNext")` and Vite build setup support it fully, and it would keep the code cleaner and align with broader TS best practices... unless there is another reason to keep them (e.g., for direct Node ESM compatibility without bundling)?

I've tested removing the `.js`, and it compiles cleanly.

This isn't breaking anything currently, so isn't urgent. More a style issue.

gudnuf:
We should! iirc the .js extensions are an artifact from before using vite and we just never took the time to clean up the imports

PR #335, opened 2025-07-25
Move to extensionless imports

robwoodgate:
# Fixes: #332 

## Description

This PR removes the `.js` extensions from all import statements across the library
e.g., changing `import { foo } from './bar.js';` to `import { foo } from './bar';`.

This creates cleaner and more idiomatic code: Extensionless imports align with standard TypeScript conventions and reduce visual clutter, making the codebase easier to read and maintain.

It also improves compatibility with bundlers like Vite, that handle module resolution automatically, eliminating the need for explicit extensions and avoiding potential resolution errors.

No functional changes; all tests pass.

...

## Changes

- removes all `.js` extensions from imports

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/examples/simpleWallet_example.ts
+++ b/examples/simpleWallet_example.ts
@@ -1,5 +1,5 @@
-import { CashuMint } from '../src/CashuMint.js';
-import { CashuWallet } from '../src/CashuWallet.js';
+import { CashuMint } from '../src/CashuMint';
+import { CashuWallet } from '../src/CashuWallet';
 
 import dns from 'node:dns';
 import {
@@ -9,8 +9,8 @@ import {
 	MintQuoteState,
 	Proof,
 	Token
-} from '../src/model/types/index.js';
-import { getEncodedTokenV4, sumProofs } from '../src/utils.js';
+} from '../src/model/types/index';
+import { getEncodedTokenV4, sumProofs } from '../src/utils';
 dns.setDefaultResultOrder('ipv4first');
 
 const externalInvoice =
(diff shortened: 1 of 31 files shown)


gandlafbtc:
Hi Rob,

I think the reason we had that, was because without it it had troubles being correctly imported in certain environments.
Though, this might have changed since we changed the bundler to vite. 

Could you check if it runs as expected in browser/node/bun after bundling?

Egge21M:
> I think the reason we had that, was because without it it had troubles being correctly imported in certain environments. Though, this might have changed since we changed the bundler to vite.

It did. This should only affect source files. Vite will automatically resolve modules without extensions. The output produced by the bundler will then explicitly point to files with extensions.

Example (crypto/client/NUT09.ts):

Source: 
```ts
import { HDKey } from '@scure/bip32';
import { getKeysetIdInt } from '../common/index';

```
Output: 

```js
import { HDKey as v } from "@scure/bip32";
import { getKeysetIdInt as a } from "../common.es.js";
```

Egge21M:
@gandlafbtc I gotta apologise. I think you were right. While the bundle is fine, it is the declaration files that cause an issue with downstream using "NodeNext" resolution (which expects file extensions in d.ts files).

I think we still should move on with extension-less imports, but we gotta figure out how to handle declaration files. We could either roll up our declarations, or post process them. I'll open an issue to discuss this. 

This doesn't block the release, as it is already an issue with some imports in 2.5.3

PR #336, opened 2025-07-28
Keysets ID v2 and New Secret Derivation

lollerfirst:
## Description

This PR aims to accomplish 3 main objectives:
- Support for keysets ID v2 (https://github.com/cashubtc/nuts/pull/182)
- ~~Verification of the keyset ID when a fetch is in order~~ already implemented in master
- Implement the new secret derivation (https://github.com/davidcaseria/nuts/pull/4): it uses HMAC-SHA512 instead of BIP32 for deriving secrets and blinding factors

## Changes

All changes are retro-compatible.

#### Changes for V2 keyset IDs:
- `deriveKeysetId` is now taking `unit`, `expiry` and `versionByte` and computes the keyset ID based on `versionByte`.
- `verifyKeysetId` is a new function that verifies that a given keyset ID matches the re-computed one.
- `getKeys` now verifies any keys fetched from the Mint.
- `getEncodedTokenV{3|4}` now trim the keyset ID to be a short ID: in the case of a keyset ID v1, this means a no-op.
- `getDecodedToken` can now take an optional `keysets` and tries to map any v2 short ID to a full-length ID v2.

#### Changes for the new secret derivation:
- `derive` now uses the new method
- `derive_deprecated` still retains the deprecated logic with BIP-32 and keyset ID to integer mapping.

#### Other changes
- Tests: modified the tests according to the changes.
- Tests for new secret derivation

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


src/crypto/client/NUT09.ts
```diff
@@ -1,5 +1,6 @@
-import { HDKey } from '@scure/bip32';
+import * as crypto from 'crypto';
```
Egge21M: `crypto` is a node module and will not be available in the browser. Please use `hmac` from `@noble/hashes` instead. That lib is already available in cashu-ts

a1denvalu3: @Egge21M https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API 

I think it's supported. But anyway I'll happily follow your suggestion.

Egge21M: Yeah, there definitely is hmac support for browsers, but available as a global / `window.crypto.subtle.importKey`. I am unsure if vite would transpile this correctly. Using @noble/hashes puts us on the safer side

Egge21M: I was able to confirm the issue. As it turns out our browser env tests are running nodeJs env with additional browser APIs, making them pretty useless for our case.

When used in a browser it throws as expected.
Bildschirmfoto 2025-07-28 um 21 02 07

I also noticed that the Buffer polyfill is missing in `NUT09.ts` and `utils.ts`. Please add `import { Buffer } from 'buffer';` to these files

src/crypto/client/NUT09.ts
```diff
@@ -9,22 +12,61 @@ enum DerivationType {
...
+		Buffer.from('Cashu_KDF_HMAC_SHA512'),
+		Buffer.from(keysetId, 'hex'),
+		counterBuffer,
+	]);
+
+	// Step 2: Compute HMAC-SHA512
+	const hmacDigest = hmac(sha512, seed, message);
+
+	// Step 3: Derive secret and blinding factor
+	const secret = hmacDigest.slice(0, 32); // First 32 bytes for secret
```
Egge21M: nit: It seems like this line and the following 7 could be merged

src/utils.ts
```diff
@@ -204,6 +208,17 @@ export function getEncodedTokenV3(token: Token, removeDleq?: boolean): string {
 	return TOKEN_PREFIX + TOKEN_VERSION + encodeJsonToBase64(v3TokenObj);
 }
 
+/*
+ * Convert a keyset ID into short form
+ */
+function convertToShortKeysetId(proofs: Proof[]) {
+	return proofs.map((p) => {
+		const newP = { ...p };
+		newP.id = newP.id.slice(0, 16);
```
Egge21M, line 217: v1 keyset ids are already 16 chars, so this would be a noop in this case, right?

Egge21M (approved):

src/utils.ts
a1denvalu3: Yes that's correct.

Issue #350, opened 2025-08-27
`MintInfo.supportsBolt12Description` should be changed to check for either bolt11 or bolt12


gudnuf:
[NUT-23](https://github.com/cashubtc/nuts/blob/main/23.md#example-mintmethodsetting) says that a mint can signal whether or not creating invoices with a description when minting is supported. 

I added this [`supportsBolt12Description`](https://github.com/cashubtc/cashu-ts/blob/c70e5ac4809e7188746728fce291bb410581da2a/src/model/MintInfo.ts#L137-L140) method to the `MintInfo` class, but I think it should be changed to take a `method: 'bolt11' | 'bolt12'` argument and check for description support of the specified payment method.

Then [`createMintQuote`](https://github.com/cashubtc/cashu-ts/blob/c70e5ac4809e7188746728fce291bb410581da2a/src/CashuWallet.ts#L958) should check for description support similar to [`createMintQuoteBolt12`](https://github.com/cashubtc/cashu-ts/blob/c70e5ac4809e7188746728fce291bb410581da2a/src/CashuWallet.ts#L1012)

## Note on backwards compatability

This would be a breaking change to the `MintInfo` api, but I think that is fine because no mints are supporting bolt12. Also it looks like the PR that added this method (#342) hasn't been released yet.

Egge21M:
> @Egge21M now that this method is in the 2.7 release, does my statement that its fine to break the API still hold true? Or should we instead just add a `supportsBolt11Description` method

In this case it's absolutely fine to break IMO. As you said it's cutting edge.

robwoodgate:
Thanks @KvngMikey - appreciate you pushing through with this, sorry it took so long to get there.

PR #352, opened 2025-08-29
Generic description check for bolt11 & bolt12

KvngMikey:
# Fixes: #350 

## Description
Implements NUT-23: generic per-method description support and adds the missing guard to `createMintQuote`.

## Changes

- Refactor `MintInfo.supportsBolt12Description` → `supportsDescription(method: 'bolt11' | 'bolt12')`
- Add identical description-guard to `createMintQuote` (bolt11)
- Add integration test in `wallet.node.test.ts`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/MintInfo.ts
+++ b/src/model/MintInfo.ts
@@ -130,13 +130,15 @@ export class MintInfo {
 	}
 
 	/**
-	 * Checks if the mint supports creating BOLT12 offers with a description.
+	 * Checks if the mint supports creating invoices/offers with a description
+	 * for the specified payment method.
 	 *
-	 * @returns True if the mint supports offers with a description, false otherwise.
+	 * @param method - The payment method to check ('bolt11' or 'bolt12')
+	 * @returns True if the mint supports description for the method, false otherwise.
 	 */
-	get supportsBolt12Description() {
+	supportsDescription(method: 'bolt11' | 'bolt12'): boolean {
 		return this._mintInfo.nuts[4]?.methods.some(
-			(method) => method.method === 'bolt12' && method.options?.description === true,
-		);
+			(met) => met.method === method && met.options?.description === true,
+		) ?? false;
 	}
 }
--- a/test/wallet.node.test.ts
+++ b/test/wallet.node.test.ts
@@ -156,6 +156,43 @@ describe('test info', () => {
 			{ method: 'nostr', info: 'npub1337' },
 		]);
 	});
+	test('supportsDescription and createMintQuote description check', async () => {
+		//behaviour with original mintInfoResp
+		server.use(
+			http.get(mintUrl + '/v1/info', () => {
+				return HttpResponse.json(mintInfoResp);
+			}),
+		);
+		const wallet = new CashuWallet(mint, { unit });
+		const info = await wallet.getMintInfo();
+
+		// mintInfoResp advertises description=true only for usd/eur; sat lacks it
+		expect(info.supportsDescription('bolt11')).toBe(false);
+		expect(info.supportsDescription('bolt12')).toBe(false);
+
+		//bolt11 description not supported
+		const noDescResp = JSON.parse(JSON.stringify(mintInfoResp));
+		const bolt11Method = noDescResp.nuts[4].methods.find((x: any) => x.method === 'bolt11');
+		if (bolt11Method) (bolt11Method.options ??= {}).description = false;
+
+		server.use(
+			http.get(mintUrl + '/v1/info', () => {
+				return HttpResponse.json(noDescResp);
+			}),
+		);
+
+		await expect(wallet.createMintQuote(1000, 'should fail')).rejects.toThrow(
+			/Mint does not support description for bolt11/,
+		);
+
+		//no description provided
+		server.use(
+			http.post(mintUrl + '/v1/mint/quote/bolt11', () => {
+				return HttpResponse.json({ quote: 'ok-quote', request: 'lnbc...' });
+			}),
+		);
+		await expect(wallet.createMintQuote(1000)).resolves.toHaveProperty('quote', 'ok-quote');
+	});
 });
 
 describe('test fees', () => {
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -956,6 +956,14 @@ class CashuWallet {
 	 *   specified amount and unit.
 	 */
 	async createMintQuote(amount: number, description?: string): Promise<MintQuoteResponse> {
+		// Check if mint supports description for bolt11
+		if (description) {
+			const mintInfo = await this.lazyGetMintInfo();
+			if (!mintInfo.supportsDescription('bolt11')) {
+				throw new Error('Mint does not support description for bolt11');
+			}
+		}
+
 		const mintQuotePayload: MintQuotePayload = {
 			unit: this._unit,
 			amount: amount,
@@ -1018,7 +1026,7 @@ class CashuWallet {
 	): Promise<Bolt12MintQuoteResponse> {
 		// Check if mint supports description for bolt12
 		const mintInfo = await this.lazyGetMintInfo();
-		if (options?.description && !mintInfo.supportsBolt12Description) {
+		if (options?.description && !mintInfo.supportsDescription('bolt12')) {
 			throw new Error('Mint does not support description for bolt12');
 		}
 


src/model/MintInfo.ts
```diff
@@ -130,13 +130,15 @@ export class MintInfo {
...
-	 * Checks if the mint supports creating BOLT12 offers with a description.
+	 * Checks if the mint supports creating invoices/offers with a description
+	 * for the specified payment method.
 	 *
-	 * @returns True if the mint supports offers with a description, false otherwise.
+	 * @param method - The payment method to check ('bolt11' or 'bolt12')
+	 * @returns True if the mint supports description for the method, false otherwise.
 	 */
-	get supportsBolt12Description() {
+	supportsDescription(method: 'bolt11' | 'bolt12'): boolean {
```
gudnuf: Did you run `npm run api:update` and get a diff on api? You need to include those changes otherwise the CI tests won't pass

src/model/MintInfo.ts
```diff
@@ -130,13 +130,15 @@ export class MintInfo {
...
 		return this._mintInfo.nuts[4]?.methods.some(
-			(method) => method.method === 'bolt12' && method.options?.description === true,
-		);
+			(met) => met.method === method && met.options?.description === true,
+		) ?? false;
```
gudnuf: nit, but `methods.some(...)` will return false if there are no matches, so in my opinion its redundant to add `?? false`

test/wallet.node.test.ts
```diff
@@ -156,6 +156,43 @@ describe('test info', () => {
...
+		expect(info.supportsDescription('bolt11')).toBe(false);
+		expect(info.supportsDescription('bolt12')).toBe(false);
+
+		//bolt11 description not supported
+		const noDescResp = JSON.parse(JSON.stringify(mintInfoResp));
+		const bolt11Method = noDescResp.nuts[4].methods.find((x: any) => x.method === 'bolt11');
+		if (bolt11Method) (bolt11Method.options ??= {}).description = false;
+
+		server.use(
+			http.get(mintUrl + '/v1/info', () => {
```
gudnuf: I just realized that description support is on a per-unit basis. For example, 'sat' may support invoice descriptions, but 'usd' might not. Can you add a `unit` parameter to `supportsDescription`? 

Also, I think instead of modifying the fetched info response and making this `noDescResp`, just modify the original mintInfoResp. I would think having nut4 signal description support for sats and not for usd. Then in your tests you should check both, but you only need to fetch it once.

gudnuf (changes requested):

src/model/MintInfo.ts
KvngMikey: that's right, thanks.

src/model/MintInfo.ts
KvngMikey: i ran it but no diff was produced.

test/wallet.node.test.ts
KvngMikey: done

src/model/MintInfo.ts
```diff
@@ -130,13 +130,18 @@ export class MintInfo {
...
-	 * Checks if the mint supports creating BOLT12 offers with a description.
+	 * Checks if the mint supports creating invoices/offers with a description for the specified
+	 * payment method.
 	 *
-	 * @returns True if the mint supports offers with a description, false otherwise.
+	 * @param method - The payment method to check ('bolt11' or 'bolt12')
+	 * @returns True if the mint supports description for the method, false otherwise.
 	 */
-	get supportsBolt12Description() {
+	supportsDescription(method: 'bolt11' | 'bolt12', unit?: string): boolean {
```
gudnuf: do you think unit should be optional? If undefined then this will return true if any unit for the specified method supports descriptions

gudnuf:
@KvngMikey this will be fun. We need to fix the nutshell integration tests. We should check if cashubtc/nutshell:0.16.4 and FakeWallet (this is what's [currently running in CI](https://github.com/cashubtc/cashu-ts/blob/main/.github/workflows/nutshell-integration.yml)) supports this description option. If it does, then we need to figure out how to make it return true in the mint's info or why we are reading it as not supporting descriptions.

src/model/MintInfo.ts
gudnuf: alright, the api extractor test passed so I'm not going to pretend to understand what's going on there

src/model/MintInfo.ts
KvngMikey: that's true, making unit mandatory avoids the foot-gun of ‘true if any unit supports descriptions’.

src/model/MintInfo.ts
```diff
@@ -130,13 +130,15 @@ export class MintInfo {
...
-	 * Checks if the mint supports creating BOLT12 offers with a description.
+	 * Checks if the mint supports creating invoices/offers with a description for the specified
+	 * payment method.
 	 *
-	 * @returns True if the mint supports offers with a description, false otherwise.
+	 * @param method - The payment method to check ('bolt11' or 'bolt12')
+	 * @returns True if the mint supports description for the method, false otherwise.
 	 */
-	get supportsBolt12Description() {
+	supportsDescription(method: 'bolt11' | 'bolt12', unit: string): boolean {
```
Egge21M: I think we need to come up with a less ambiguous name here to avoid confusion with the `description` info key

gudnuf: `supportsNut04Description`?

Egge21M: Yes that one is great!

KvngMikey: you're welcome @Egge21M.

alright, i'll update the PR with this naming convention and the CI fix, thank you @gudnuf

gudnuf:
@KvngMikey thanks for sticking with it and making these changes. Your fix in [e94025](https://github.com/cashubtc/cashu-ts/pull/352/commits/e94025c3270b3dbb6721d4caf9ee3ca3074c3966) is not fixing the correct test. If you look in that wallet.node.test.ts test file at the info that is being used you'll see `"nuts":{"4":{"methods":[{"method":"bolt11","unit":"sat","description":true}`.

See the integration test [action that failed](https://github.com/cashubtc/cashu-ts/actions/runs/17342361705/job/49244074698). This is the test we need to fix. The problem with this test is we are starting a actual instance of nutshell and for some reason not reading that the mint support nut04 subscriptions.

gudnuf:
@KvngMikey https://github.com/cashubtc/cashu-ts/pull/352/commits/cfd3dc6ee833b6e3cb5caa9f109a6adeac3b41c0 is not an adequate fix. The test is for invoices with a description and you just removed the description. We need to address the root cause of the problem which is what I described [here](https://github.com/cashubtc/cashu-ts/pull/352#issuecomment-3247108658) already. We want to keep the test unchanged and make it pass by fixing the nutshell integration tests

KvngMikey:
@gudnuf , okay so i have reverted the test case to keep it as is, based on what you're saying it seems i need to make an update to the nutshell-integration.yml file or what exactly do you have in mind?

gudnuf:
> @gudnuf , okay so i have reverted the test case to keep it as is, based on what you're saying it seems i need to make an update to the nutshell-integration.yml file or what exactly do you have in mind?

Yea I think that's what you need to do, but I'm not exactly sure why nutshell isn't already returning enabled descriptions. You should try to run the same nutshell version that is running in nutshell-integration.yml locally on your machine, then point the integration tests to that local instance and see if they pass. If they pass, then we know its a CI problem.

I would need to dig in more to give you an exact path

src/model/types/mint/responses.ts
```diff
@@ -319,6 +319,7 @@ export type SwapMethod = {
 	unit: string;
 	min_amount: number;
 	max_amount: number;
+	description?: boolean; // <-- add this for Nutshell 0.16.4 compatibility
```
gudnuf: Good catch! it seems even the latest of nutshell has this spec incompatibility. You can check [lnvoltz](https://mint.lnvoltz.com/v1/info) for example which is on v0.17.0.  Do you want to open an issue to nutshell about this? We should open an issue to nutshell and then mention the incompatibility with a link to the issue in the comment here. See the relevant part of nut23 [here](https://github.com/cashubtc/nuts/blob/main/23.md#example-mintmethodsetting)

test/wallet.node.test.ts
```diff
@@ -156,6 +156,34 @@ describe('test info', () => {
...
+	test('supportsNut04Description respects per-unit description support', async () => {
+		server.use(
+			http.get(mintUrl + '/v1/info', () => {
+				return HttpResponse.json(mintInfoResp);
+			}),
+		);
+		const wallet = new CashuWallet(mint, { unit: 'sat' });
+		const info = await wallet.getMintInfo();
+
+		expect(info.supportsNut04Description('bolt11', 'sat')).toBe(false);
```
gudnuf: this should fail right? the mintInfoResp we are using has `[{"method":"bolt11","unit":"sat","description":true}...`. I would expect this to fail because it should support description.

src/model/types/mint/responses.ts
gudnuf: There is an open PR to nutshell for this now (https://github.com/cashubtc/nutshell/pull/783). I would still add a comment linking to the fix so that we know why this extra description added.

KvngMikey: @gudnuf thanks for doing this, I couldn't get to it yesterday due to some problems with my computer...

this PR has been merged by Calle.

I have also added the PR link in the comment of the mint responses for future reference.

Is there anything else on here you want me to do for you ?

test/wallet.node.test.ts
KvngMikey: you’re totally right, that line was a leftover from an earlier draft where I was testing something out.
Since the mintInfoResp does have "description": true for sat, the expectation should obviously be true. I’ve flipped it in the latest commit so the test now correctly asserts that our guard allows the call when the mint signals support. Thanks for spotting it!

gudnuf (approved):
ACK

robwoodgate (approved):
@KvngMikey - I fixed up the failing CI test. V3 wallet needs an `await wallet.loadMint();` to fully init. I also added a usd keyset so the wallet didn't complain about keyset not being init.

ACK @ 4e8f66a

etc/cashu-ts.api.md
```diff
@@ -803,7 +803,7 @@ export class MintInfo {
     requiresBlindAuthToken(method: 'GET' | 'POST', path: string): boolean;
     // (undocumented)
     requiresClearAuthToken(method: 'GET' | 'POST', path: string): boolean;
-    get supportsBolt12Description(): boolean;
```
Egge21M, line 809: We should not delete this old getter. Instead we should try to wire it up with the new function `supportsNut04Description` and mark it as deprecated

KvngMikey: this has been done.

PR #355, opened 2025-09-04
V3-alpha: Refactor CashuWallet to Wallet v3

robwoodgate:
## Overview
Introduces a new Wallet class (v3) as a work-in-progress redesign of CashuWallet (v2), aiming for improved separation of concerns, simplified options handling, and a cleaner, more flexible API through the use of tagged unions.

Retains full functionality while adding beginner-friendly helpers (e.g., sendAsDefault, receiveAsP2PK) and transaction pipeline optimizations.

There are also additions to the OutputData class and utils.ts which are not used by v2.

## Key Changes

- **Refactored types folder** - moved wallet and mint to their own domain folders and located specific types with each for better DX. Shared types folder (model/types) flattened and reorganized.
- **Refactored Crypto Library** - Crypto library flattened and reorganized by NUT.
- **Removed subpackage exports** - everything public now exports from src/index (@cashu-ts)
- **New Class Structure:** CashuWallet (v2) remains unchanged; new Wallet (v3) class with similar constructor but enhanced methods.
- **Simpler instantiation** - Wallet can be instantiated with a mint url - no need to instantiate a mint first. 
- **New KeyChain / Keyset classes** - abstracts keychain logic away from wallet. Can be init in wallet constructor or during loadMint(). Caching method (`keyChain.getCache`) makes offline Wallet instantiation simpler.
- **Output Configuration:** Replaces scattered options (e.g., outputAmounts, pubkey) with OutputType union (random, deterministic, p2pk, factory, custom) and OutputConfig for send/keep.
- **Method Helpers:** Added type-safe wrappers like sendAsDeterministic, mintProofsAsP2PK for common patterns; core methods (e.g., send, receive) now take OutputType/OutputConfig.
- **Initialization:** Unified loadMint() handles info, keysets, and keys; getters throw if not loaded.
- **Offline Send:** New sendOffline for exact-match selections without swaps.
- **Fees & Selection:** Refined selectProofsToSend with optional exactMatch; retained fee handling in outputs.
- **Deterministic Outputs:** Auto-offsets counters in deterministic sends to avoid overlaps.
- **Defaults**: Provided DEFAULT_OUTPUT and DEFAULT_OUTPUT_CONFIG for easy customization.

## Breaking Changes from V2

- **Type imports:** Type folder structure has changed. Public types can now be imported from src/index (@cashu-ts)
- **Crypto Imports** - The sub-package exports are now gone. Import crypto functions from src/index (@cashu-ts)
- **Class Rename:** CashuWallet -> Wallet (v3); CashuMint -> Mint (v3); import and instantiate the new class for v3 features.
- **Method Signatures**: receive, send, mintProofs, meltProofs now use OutputType/OutputConfig instead of flat options (e.g., no more pubkey, counter as separate params). See test/walletv3.node.test for examples.
- Removed options objects with mixed props; use tagged unions.
- Removed wallet getters for key/keyset related calls. Moved into KeyChain class
- swap is now an alias of send.
- **Removed Constants:** Dropped DEFAULT_DENOMINATION_TARGET, DEFAULT_UNIT; integrated into class props.
- **Seed Requirement:** Deterministic outputs strictly require BIP39 seed in constructor.
- **Custom Outputs:** custom type bypasses splitting/fees; may cause mismatches if misused.
- **Deprecated Patterns:** v2's keepFactory, outputData option in send/receive replaced by OutputType.factory/custom.
- **Error Handling:** More throws for invalid states (e.g., unloaded mint); call loadMint() first.
- **Mint** - removed static methods. Added custom request param to instance methods for flexibilty

See the READMEv3.md files for examples of use.

Feedback and testing welcome on v3 API!

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/types/keyset.ts
+++ b/src/model/types/keyset.ts
@@ -67,6 +67,7 @@ export type MintKeyset = {
 	 * Input fee for keyset (in ppk)
 	 */
 	input_fee_ppk?: number;
+
 	/**
 	 * Expiry of the keyset.
 	 */


robwoodgate:
Rebased to development to include #358

robwoodgate:
Now exports ALL core utils, including `splitAmount`, `sumProofs`, `verifyKeysetId`, etc

robwoodgate:
Hey @lollerfirst - I've fixed the references to deprecations in @noble.

fc85c6cbf11ad63d4d69f9c7ba04bf1e9f5a7130 - fixes the crypto library
08b0c405c5b7bff777b70f2fc6950eec2cd51cb8 - renames createRandomPrivateKey and fixes @noble references in tests

All tests passed, but please can you cast and eye over them to make sure I've migrated the functions correctly.
Also, the `Point.Fn.add` and `Point.Fn.mul` functions are apparently constant-time safe.
Please can you confirm, and if so, we can remove the warning on `createDLEQProof`

src/crypto/NUT12.ts
```diff
@@ -0,0 +1,63 @@
+import { type DLEQ, hash_e, hashToCurve, createRandomSecretKey } from './core';
+import { bytesToNumber } from '../utils';
+import { type WeierstrassPoint } from '@noble/curves/abstract/weierstrass';
+import { numberToBytesBE } from '@noble/curves/abstract/utils';
+import { secp256k1 } from '@noble/curves/secp256k1';
+
+function arraysEqual(arr1: Uint8Array, arr2: Uint8Array) {
```
a1denvalu3, line 7: Not really an issue, but we could make this comparison constant-time since it's in the crypto library.
You can do it by point-wise XORing each element arr1[i] ^ arr2[i] and then ORing the result with a running variable, which should initially be zero'd out. If the running variable isn't zero at the end of the procedure, return false else return true.

If you want I can also do it.

robwoodgate: It's a good idea, and I'd love you to do it please... how about push as a new PR to the v3-alpha-dev branch once this baseline PR is merged.

src/crypto/NUT01.ts
```diff
@@ -0,0 +1,80 @@
...
+		mintKeys[p] = hexToBytes(serializedMintKeys[p]);
+	});
+	return mintKeys;
+}
+
+export function getPubKeyFromPrivKey(privKey: Uint8Array) {
+	return secp256k1.getPublicKey(privKey, true);
+}
+
+export function createNewMintKeys(pow2height: IntRange<0, 65>, seed?: Uint8Array): KeysetPair {
```
Egge21M, line 48: This seems like a mint side util. Do we want to keep it for v3?

src/wallet/KeyChain.ts
```diff
@@ -0,0 +1,155 @@
+import { Keyset } from './Keyset';
+import { type Mint } from '../mint';
+import type { MintKeyset, MintKeys, MintAllKeysets, MintActiveKeys } from '../model/types/keyset';
+
+export class KeyChain {
+	private mint: Mint;
+	private unit: string;
+	private keysets: { [id: string]: Keyset } = {};
+	private _activeKeysetId: string | undefined;
```
Egge21M: Personally I always found the concept of the cached "active" keyset a bit confusing. In cashu-ts this is effectively one of the active keysets with the lowest fee. I wonder whether we should get rid of this member and simple run getCheapestKeyset when appropriate.

src/model/BlindedMessage.ts
```diff
@@ -1,11 +1,11 @@
 import { type SerializedBlindedMessage } from './types/index';
-import { type ProjPointType } from '@noble/curves/abstract/weierstrass';
+import { type WeierstrassPoint } from '@noble/curves/abstract/weierstrass';
```
Egge21M, line 2: Is that change intentional?

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+		keyset: Keyset,
+		outputType: OutputType,
+	): OutputDataLike[] {
+		if (amount < 0) {
+			// we can accept zero (for blanks) or positive values
+			this._logger.warn('Amount was negative');
+			return [];
+		}
+		if (
+			'custom' != outputType.type &&
```
Egge21M, line 589: nit: This comparison is technically correct, but semantically backwards

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+			adjustedAmount += receiveFee;
+			denominations = [...denominations, ...receiveFeeAmounts];
+		}
+
+		const effectiveOutputType: OutputType = { ...outputType, denominations };
+		return this.createOutputData(adjustedAmount, keyset, effectiveOutputType);
+	}
+
+	/**
+	 * Creates a swap transaction with sorted outputs for mint compatibility.
```
Egge21M: nit: Outputs are not sorted for compatibility, but rather to preserve privacy. By sorting the mint can not tell which outputs might be kept and which will be sent

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+	 *
+	 * @remarks
+	 * For apps storing P2PK locked proofs. Uses NUT-11 options for locking.
+	 * @param token Cashu token.
+	 * @param options P2PK locking options (e.g., pubkey, locktime).
+	 * @param denominations Optional custom amounts for splitting outputs.
+	 * @param config Optional parameters.
+	 * @returns The proofs received from the token, P2PK-locked.
+	 */
+	async receiveAsP2PK(
```
Egge21M: I am going back and forth about whether we really need all these helpers or whether it would be enough to point to the OutputType type in the regular receive call. Looking at the type definitions of OutputType should give consumers a pretty good idea about what is possible.

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+	 * complexity O(n log n) and space complexity O(n).
+	 * @param proofs Array of Proof objects available to select from.
+	 * @param amountToSend The target amount to send.
+	 * @param includeFees Optional boolean to include fees; Default: false.
+	 * @param exactMatch Optional boolean to require exact match; Default: false.
+	 * @returns SendResponse containing proofs to keep and proofs to send.
+	 * @throws Throws an error if an exact match cannot be found within MAX_TIMEMS.
+	 * @see https://crypto.ethz.ch/publications/files/Przyda02.pdf
+	 */
+	selectProofsToSend(
```
Egge21M, line 1026: I might be overlooking something, but I think this function is stateless and could be extracted from this module

Egge21M: PS: I see that it uses the Wallet's fee calculation logic, but it seems to me that this does not really belong in this class either. Not dying on this hill though, what do you think?

src/model/BlindedMessage.ts
robwoodgate: Yes, there is a deprecation warning in noble which recommends the change

src/wallet/Wallet.ts
robwoodgate: Hahaha, you must have read my mind... am already working on v3-alpha-2 which abstracts it!

src/wallet/Wallet.ts
robwoodgate: I'm already going back on the idea - I think you'll like what I'm working on RN with v3-alpha-2

src/wallet/Wallet.ts
robwoodgate: I think it's the error logging message that confuses it. Custom output type has no denominations, everything else does... so the test is "if you supplied denominations, they better add up to the amount".

src/crypto/NUT01.ts
robwoodgate: I don't mind either way - it will be shaken out of most builds, so probably does no harm and "completes" the crypto side of the library, but it probably will never be used ... except maybe in future unit tests?

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+		if (config?.privkey) {
+			proofs = this.signP2PKProofs(proofs, config?.privkey);
+		}
+		// Check DLEQs if needed
+		const keyset = this.keyChain.getKeyset(config?.keysetId);
+		if (config?.requireDleq && proofs.some((p) => !hasValidDleq(p, keyset))) {
+			const message = 'Token contains proofs with invalid or missing DLEQ';
+			this._logger.error(message);
+			throw new Error(message);
+		}
```
robwoodgate, line 760: I kept the logic same as v2 here, but I'm wondering if it might be wrong.

It ensures all proofs are from the SPECIFIED keyset with valid DLEQ. If the token contains proofs from other keysets, the receive will be rejected, even if their DLEQ is valid.

We might just want to check the proofs have a valid DLEQ - regardless of keyset. 
This is [what the docs suggest](https://cashu-ts.dev/docs/API/type-aliases/ReceiveOptions).

So maybe something like:

```typescript 
if (requireDleq) {
  for (const p of proofs) {
    const ks = this.keyChain.getKeyset(p.id);
    if (!hasValidDleq(p, ks)) {
      throw new Error('Token contains proofs with invalid or missing DLEQ');
    }
  }
}
```

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+		meltQuote: T extends 'bolt11' ? MeltQuoteResponse : Bolt12MeltQuoteResponse,
+		proofsToSend: Proof[],
+		outputType: OutputType = DEFAULT_OUTPUT,
+		config?: MeltProofsConfig,
+	): Promise<MeltProofsResponse> {
+		const { keysetId, onChangeOutputsCreated } = config || {};
+		const keyset = this.keyChain.getKeyset(keysetId);
+		const feeReserve = sumProofs(proofsToSend) - meltQuote.amount;
+		let outputData: OutputDataLike[] = [];
+
```
robwoodgate: Should we throw if sumProofs is less than meltQuote amount (v2 did not)?
Or is a partial melt ever ok?

Bolt12 - "wallet.meltProofsBolt12 delegates and returns" test fails if we do

src/wallet/Wallet.ts
robwoodgate: Clarified in v3-alpha-2

src/wallet/Wallet.ts
robwoodgate: v3-alpha-2 removes them all in favour of a WalletOps builder

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+	 * @param config Optional parameters for the swap.
+	 * @returns SendResponse with keep/send proofs.
+	 * @throws Throws if the send cannot be completed offline or if funds are insufficient.
+	 */
+	async send(
+		amount: number,
+		proofs: Proof[],
+		outputConfig: OutputConfig = DEFAULT_OUTPUT_CONFIG,
+		config?: SendConfig,
+	): Promise<SendResponse> {
```
robwoodgate, line 858: In v3-alpha-2, I've reversed the 3rd and 4th params in send/receive/mint/melt - it makes for less of a breaking change upgrading from v2, and this removes the need for the DEFAULT_OUTPUT etc. if you want to apply a config only

So it is currently a breaking change to the signature in thie PR that goes away in next version

src/wallet/Wallet.ts
robwoodgate: Clarified in v3-alpha-2

src/wallet/Wallet.ts
a1denvalu3: In theory you could pay only `x` out of `y` of a melt quote if the method is -for example- bolt12. But I don't think we should allow it. CDK errors when this situation is presented.

src/utils/core.ts
```diff
@@ -44,22 +44,50 @@ export function splitAmount(
...
-		for (let i = 0; i < q; ++i) split?.push(amt);
+		for (let i = 0; i < q; ++i) split.push(amt);
 		value %= amt;
 	});
+	if (value !== 0) {
+		throw new Error(`Unable to split remaining amount: ${value}`);
+	}
+
+	// Only sort when we performed a fill; exact custom splits returned earlier
 	return split.sort((a, b) => (order === 'desc' ? b - a : a - b));
```
a1denvalu3: Could we potentially remove this sort, and let it be handled elsewhere? For example the sorting and saving the original order is already handled in your `createSwapTransaction`. I don't exactly remember why we need this sort here.

a1denvalu3: Or at least change the `order` parameter such that when unspecified, we don't enforce any ordering on the outputs

src/wallet/Wallet.ts
robwoodgate: Ok, so a throw is preferable and the test should be updated

src/utils/core.ts
robwoodgate: I think the logic for sorting if a fill was needed was to make partial denomination requests less obvious... eg, I definitely want [8,4] but don't mind how the rest comes.

Without sorting, it looks something like: [8,4,32,8,1]

But as all outputs are sorted before being sent to the mint, this is kind of redundant.

@Egge21M - do you recall why sorting was applied here originally?

robwoodgate: > Or at least change the `order` parameter such that when unspecified, we don't enforce any ordering on the outputs

I think this is the best of all worlds... allows sorting if desired, don't do it by default. Will break a bunch of tests though, as default sort order was ASC

a1denvalu3: > I think the logic for sorting if a fill was needed was to make partial denomination requests less obvious...

yes that was the reason but since now we handle that more precisely i don't see why we need this sort here.

src/wallet/Wallet.ts
a1denvalu3: Yes this was an oversight. The way you described it should fix it.

robwoodgate: Added in v3-alpha-2

src/utils/core.ts
robwoodgate: ok, sort made optional in v3-alpha-2

src/wallet/Wallet.ts
robwoodgate: added to v3-alpha-2

src/wallet/Wallet.ts
```diff
@@ -0,0 +1,2192 @@
...
+		// Ensure witnesses are serialized, strip DLEQ if not required
+		const sendPrepared = this._prepareInputsForMint(send, requireDleq);
+		return { keep, send: sendPrepared };
+	}
+
+	/**
+	 * Sends proofs using Default (random) secrets for both send and keep outputs.
+	 *
+	 * @remarks
+	 * Beginner-friendly default for privacy-focused sends. Uses random blinding to avoid linkability.
```
TheRealCheebs: I like/appreciate these beginner-friendly callouts.

src/crypto/NUT12.ts
robwoodgate: Looking again, we already have a constant time array equals in the Bytes class: `Bytes.equals()`. We could just switch to using that.

PR #356, opened 2025-09-08
Chore/cdk ci migration

d4rp4t:
# Fixes: #344

## Description

Migration from Nutshell Mint to CDK mintd. Reused as much mint parameters as possible.

## Changes

- Renamed `nutshell-integration.yml` to `cdk-mint-integration.yml`. Changed the docker image to mintd
- Updated integration tests to match the new implementation.
- Left two TODOs:
  1. Once the CDK Mintd release addresses the already raised [issue](https://github.com/cashubtc/cdk/issues/1033), we can uncomment one.
  2. The second TODO is more complex because the CDK Mint websocket connection disconnect time is around 6000ms, whereas with Nutshell it was about 500ms.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/test/integration.test.ts
+++ b/test/integration.test.ts
@@ -271,6 +271,7 @@ describe('mint api', () => {
 
 		// Mint some proofs
 		const request = await wallet.createMintQuoteBolt11(128);
+		await untilMintQuotePaid(wallet, request);
 		const mintedProofs = await wallet.mintProofs(128, request.quote);
 
 		// Send them P2PK locked to Alice, with extra tags


TheRealCheebs:
I just pushed changes to @d4rp4t branch. Sorry I got a bit carried away. After the call we had figured it would be good to keep the nutshell mint in addition to the cdk mint. I didn't like how the two files were identical so I pulled the functionality of spinning up the image and running the tests into one file and created two integration-mint-<mint> files.

I would like to also add something like renovate so if we have minor/patch changes it will open a pr and someone can approve that we bump mint versions.

robwoodgate:
@d4rp4t - I've bumped Nutshell to latest version (0.18.1), and set a 1 second fakewallet delay on both Nutshell and CDK.

I then fixed up the tests to properly wait for the Mint Quote to be paid - I used the new v3 wallet event to listen for it, rather than a plain timeout. I also added global timeout to the integration test suite.

This also lead me to fix up a potential cancel leak in Events... as tests were timing out... and eventually find out Nutshell was rate limiting after 20 calls, hahaha... so it now has an expanded rate limit.

Finally, I added mint up/down instructions to the integration test, and a Makefile at the root of the package to make it easy for us to spin up a CDK/Nutshell mint locally. I also added instructions to the CONTRIBUTING file.

PR #360, opened 2025-09-10
Add support for `Prefer: respond-async` on melt operations

Egge21M:
## Description
This adds a new flag on `MeltProofsOptions` to signal a client prefers an async response (don't block until melt completes, instead respond immediately with a pending state, see [NUT](https://github.com/cashubtc/nuts/pull/285)). I was hesitant to add another flag, but in this case it seems to most reasonable approach.

## Changes

- Creates new methods on CashuMint for async melt operations
- Adds `preferAsync` flag to `MintProofOptions`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


robwoodgate:
This is great, @Egge21M  - I wonder if we can harmonize with the approach I've taken in v3? https://github.com/cashubtc/cashu-ts/pull/355/commits/3d732f7e7af4def7ab26784f20f8073871479952

This adds an `onChangeOutputsCreated` callback option (which could replace your async flag) If the callback is specified, it returns the change outputs so change can be recovered async. 

For v2->v3, the option wouldn't then change, and the consumer adding a callback would use the preferSync pathway

Otherwise we end up with two options to achieve the same thing (or a breaking change right after the option is put in).

PR #363, opened 2025-09-14
change checkProofsStates type to only require secrets

gudnuf:
# Fixes: 
`checkProofsStates` says it only requires `secret`s, but the type says otherwise.
## Description
This is a non-breaking change that allows users to only pass the secret to `checkProofsStates` instead of a whole proof
...

## Changes

- ...
- ...

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -165,7 +165,7 @@ export class CashuWallet {
     // (undocumented)
     checkMintQuote(quote: string): Promise<PartialMintQuoteResponse>;
     checkMintQuoteBolt12(quote: string): Promise<Bolt12MintQuoteResponse>;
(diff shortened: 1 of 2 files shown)


Egge21M:
Good change, thank you! Just curious: In what situation would you end up with p = {secret}?
I wondered whether this function should be overloaded instead to work on `Proof[] | string[]`

gudnuf:
Yea that would be perfect for what I need now @robwoodgate 

My use case went away because I realized I had to restore before checking state, and once I restore I get the full proof.

I will just close this unless you guys @Egge21M think we should function overload with `string[]`, but I don't need this anymore

Egge21M:
@gudnuf this got merged in v3. Do we close this?

PR #368, opened 2025-10-05
V3 alpha 2

robwoodgate:
## Description

Version 3 Alpha 2 of Cashu-TS. This version adds:

- Deterministic counters with callbacks for persistence (WalletCounters / CounterSource)
- Improved wallet event subscriptions (WalletEvents)
- Transaction builder (WalletOps)
- P2PK options builder (P2PKBuilder)
- Abstracted coin selection (SelectProofs) for easy plugin/replacements
- Improved v2 compatibility vs v3-alpha-1
- Improved test coverage
- Clearer wallet layout / sections

It also adds detailed documentation to README, migration guide, and updates examples.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


src/index.ts
```diff
@@ -1,37 +1,53 @@
...
+	MeltBuilder,
+	WalletEvents,
+	WalletOps,
+	Wallet,
+} from './wallet';
+export type * from './wallet/types/payloads';
+export * from './mint/types';
 
 // Core models & primitives
 export * from './model/types';
```
Egge21M: Do we want to keep exporting all these types, or do we want to start to be more selective here?

src/wallet/Wallet.ts
```diff
@@ -417,201 +686,99 @@ class Wallet {
...
+		outputType?: OutputType,
 	): Promise<Proof[]> {
-		const outputType: OutputType = {
-			type: 'deterministic',
-			counter,
-			denominations,
-		};
-		return this.receive(token, outputType, config);
-	}
+		const { keysetId, privkey, requireDleq, proofsWeHave, onCountersReserved } = config || {};
```
Egge21M, line 718: Do we need the onCountersReserved ballback here, now that we have the class-wide event system? What would be the usecase?

src/wallet/Wallet.ts
```diff
@@ -213,14 +301,266 @@ class Wallet {
...
+		if (totalOutputs <= 0) return { start: 0, count: 0 };
+		return this._counterSource.reserve(keysetId, totalOutputs);
+	}
+
+	private countersNeeded(ot: OutputType): number {
+		if (ot.type !== 'deterministic' || ot.counter !== 0) return 0;
+		return (ot.denominations ?? []).length;
+	}
+
+	private async setAutoCounters(
```
Egge21M: I think this function could use a more descriptive name, so that it is clear what it does when contributors encounter it in the code. How about `addCountersToOutputTypes`

src/wallet/WalletCounters.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Developer friendly view of the wallet's deterministic output counters.
+ */
+export class WalletCounters {
+	constructor(private readonly src: CounterSource) {}
+	/**
+	 * Returns the "next" counter for a specified keyset.
+	 */
+	async peekNext(keysetId: string): Promise<number> {
+		const r = await this.src.reserve(keysetId, 0);
```
Egge21M, line 12: Calling `reserve` with n=0 is supposed to not mutate state, but wouldn't it be better to have the source implement an explicit peek method to not "overload" reserve to also peek?

src/wallet/Wallet.ts
```diff
@@ -213,14 +301,266 @@ class Wallet {
...
-		}
+		this.failIfNullish(this._mintInfo, 'Mint info not initialized; call loadMint first');
 		return this._mintInfo;
 	}
 
+	/**
+	 * The keyset ID bound to this wallet instance.
+	 */
+	get keysetId(): string {
+		this.failIf(this._boundKeysetId === '__PENDING__', 'Wallet not initialised, call loadMint');
```
Egge21M: What is the reasoning behind using a string literal instead of checking for undefined?

src/wallet/Wallet.ts
```diff
@@ -213,14 +301,266 @@ class Wallet {
...
+	 * Use this to operate on multiple keysets concurrently without mutating your original wallet.
+	 * Counters remain monotonic across instances because the same CounterSource is reused.
+	 *
+	 * Do NOT pass a fresh CounterSource for the same seed unless you know exactly why. Reusing
+	 * counters can recreate secrets that a mint will reject.
+	 *
+	 * @param id The keyset identifier to bind to.
+	 * @throws If keyset not found, if it has no keys loaded, or if its unit is not the wallet unit.
+	 */
+	public withKeyset(id: string, opts?: { counterSource?: CounterSource }): Wallet {
```
Egge21M, line 441: What would be the rational to use this over changing back and forth using the `bindKeyset` method?

robwoodgate: The idea was to allow a consumer to use wallets with different keysets but same CounterSource at the same time. See Deterministic Counters, example 5 in the README:

```
// 5) Parallel keysets without mutation
const wA = wallet; // bound to '0111111'
const wB = wallet.withKeyset('0122222'); // bound to '0122222', same CounterSource
await wB.counters.advanceToAtLeast('0122222', 10);
await wA.counters.snapshot(); // { '0111111': 137, '0122222': 10 }
await wB.counters.snapshot(); // { '0111111': 137, '0122222': 10 }
wA.keysetId; // '0111111'
wB.keysetId; // '0122222'
```

robwoodgate: Whether that's useful vs just switching keysets with `bindKeyset` is not a hill I'll die on, so we can remove if not needed.

src/wallet/WalletCounters.ts
robwoodgate: The idea here was to keep the `CounterSource` interface as easy to fulfil as possible. `peekNext` is a WalletCounters convenience. But we absolutely could make it an optional on the interface.

Overloading reserve seemed ok, because what else would it return if asked to reserve 0 counters, except maybe a throw, which seemed a bit harsh.

src/wallet/Wallet.ts
robwoodgate: Great idea!

src/wallet/Wallet.ts
robwoodgate: The `onCountersReserved` is a per method convenience - an alternative to setting up the global event.

I added them before designing the global event, as a way to simply get counters for a single operation.

We may decide we don't want any per method callback... in which case am happy to remove them.

But if we decide they are useful, we need them for receive as the outputtype can be deterministic.

src/index.ts
robwoodgate: Great question. In theory, we only need to export the types made public by exported methods / functions. 

Am very happy to being selective - is that your preference here?

src/wallet/Wallet.ts
robwoodgate: Great question. I used a sentinel mainly to keep `_boundKeysetId` a plain string, avoiding `string | undefined` checks throughout. It also makes intent explicit: `__PENDING__` always means "not yet bound by loadMint."

If someone skips loadMint or the KeyChain isn’t populated, we can give a clear, deterministic error instead of having to guess why the keyset id is missing.

robwoodgate: We could absolutely turn it into a constant, rather than a string literal. Would be more robust and self-documenting.

```ts
const PENDING_KEYSET_ID = '__PENDING__';
```

robwoodgate: Added constant in 024358fce786ae30b30f7e5d543b39f75eab0c2e

src/wallet/Wallet.ts
robwoodgate: Renamed in 2f43f25659112c730040332d33145ceef9db5925

src/index.ts
robwoodgate: I just checked and we pretty much reference all the mint types publicly, so yes, we need to export them.

robwoodgate: I've been through the API and removed the private types, also removed a couple of totally unused types.

Changes - https://github.com/cashubtc/cashu-ts/pull/368/commits/8426cd1703b98ac90e874af39412a7ff50e1d6bf

PR #377, opened 2025-10-19
ECDH derived Pay-to-Blinded-Key (P2BK)

robwoodgate:
## Description

Implements Pay to Blinded Key (P2BK) as an *ECDH-derived* blinding scheme.

https://github.com/cashubtc/nuts/pull/300

## Changes

- Adds `p2pk_e` (33-byte SEC1 pubkey) per proof (stored as `pe` in token v4 format)
- Adds `nut26` flag to payment request
- Adds blinding option to `P2PKBuilder`
- Adds blinding capability to `OutputData`
- Adds NUT-26 crypto module (though NUT number TBC)

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/crypto/NUT11.ts
+++ b/src/crypto/NUT11.ts
@@ -271,7 +271,7 @@ export const signP2PKProofs = (
 ): Proof[] => {
 	return proofs.map((proof, index) => {
 		try {
-			const privateKeys: string[] = deriveP2BKSecretsForProof(privateKey, proof);
+			const privateKeys: string[] = maybeDeriveP2BKPrivateKeys(privateKey, proof);
 			let signedProof = proof;
 			for (const priv of privateKeys) {
 				try {
@@ -400,7 +400,7 @@ export const getSignedOutputs = (
  * @returns Deduplicated list of derived secret keys (hex, 64 chars)
  * @alpha
  */
-export function deriveP2BKSecretsForProof(privateKey: string | string[], proof: Proof): string[] {
+export function maybeDeriveP2BKPrivateKeys(privateKey: string | string[], proof: Proof): string[] {
 	const privs = Array.isArray(privateKey) ? privateKey : [privateKey];
 	const Ehex: string | undefined = proof?.p2pk_e;
 	if (!Ehex) {


src/crypto/NUT26.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @throws If r reduces to zero after the retry.
+ */
+function deriveP2BKBlindingTweakFromECDH(
+	point: WeierstrassPoint<bigint>, // E or P
+	scalar: bigint, // p or e
+	keysetId: Uint8Array, // kid
+	slotIndex: number, // i
+): bigint {
+	// Calculate ECDH shared point (Z)
+	const Zx = point.multiply(scalar).toBytes(false).slice(1, 33);
```
a1denvalu3: I think noble does have ECDH builtin, right?

```typescript
const shared = await secp.getSharedSecret(scalar, publicKey);
```

robwoodgate: Yes. It does the same thing - only it takes raw byte inputs, not points and scalars.

```ts
function getSharedSecret(
    secretKeyA: Uint8Array,
    publicKeyB: Uint8Array,
    isCompressed = true
  ): Uint8Array {
    if (isProbPub(secretKeyA) === true) throw new Error('first arg must be private key');
    if (isProbPub(publicKeyB) === false) throw new Error('second arg must be public key');
    const s = Fn.fromBytes(secretKeyA);
    const b = Point.fromBytes(publicKeyB); // checks for being on-curve
    return b.multiply(s).toBytes(isCompressed);
  }
```

For this initial PoC draft, I chose to follow the crypto module trend of passing around the crypto Point and scalar, so using the Noble `getSharedSecret` helper would add needless conversion overhead.

But I will revisit in a final refactor once the [p2pk_e / nonce issue](https://github.com/cashubtc/nuts/pull/300#issuecomment-3467739848) is ironed out and see if it makes life simpler in this module.

src/crypto/NUT11.ts
```diff
@@ -263,7 +271,7 @@ export const signP2PKProofs = (
 ): Proof[] => {
 	return proofs.map((proof, index) => {
 		try {
-			const privateKeys: string[] = Array.isArray(privateKey) ? privateKey : [privateKey];
+			const privateKeys: string[] = deriveP2BKSecretsForProof(privateKey, proof);
```
Egge21M: Maybe I am missing something here, but isn't this basically replacing the old P2PK flow with P2bK? Should cashu-ts not offer both?

robwoodgate: No, if `p2pk_e` is not defined, and there is nothing to do, it just returns the unique private keys untouched..

```ts
export function deriveP2BKSecretsForProof(privateKey: string | string[], proof: Proof): string[] {
	const privs = Array.isArray(privateKey) ? privateKey : [privateKey];
	const Ehex: string | undefined = proof?.p2pk_e;
	if (!Ehex) {
		return Array.from(new Set(privs));
	}
...
```

robwoodgate: I can rename the method `maybeDeriveP2BKSecretsForProof ` if that makes it clearer.

PR #378, opened 2025-10-20
Add support for Prefer: respond-async on melt operations

robwoodgate:
## Description

The v3 implementation of #360 . This implementation uses the `onChangeOutputsCreated` option as a flag - if the consumer has set it, the mint request will use `prefer-async`. If not, the mint will await the response as normal.

Also harmonises the Mint/Melt family of methods to have a "Bolt11" suffix (detailed in migration v3 doc).

## Changes

- Updates the Mint `melt()` and `MeltBolt12()` methods to allow preferAsync.
- Wallet sets the flag depending on the status of `onChangeOutputsCreated`.
- Harmonizes the Mint/Melt family of methods with "Bolt11" suffix
- Updates Migration Docs for v3

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,9 +111,9 @@ const mintUrl = 'http://localhost:3338';
 const wallet = new Wallet(mintUrl);
 await wallet.loadMint(); // wallet is now ready to use
 
-const mintQuote = await wallet.createMintQuote(64);
+const mintQuote = await wallet.createMintQuoteBolt11(64);
 // pay the invoice here before you continue...
-const mintQuoteChecked = await wallet.checkMintQuote(mintQuote.quote);
+const mintQuoteChecked = await wallet.checkMintQuoteBolt11(mintQuote.quote);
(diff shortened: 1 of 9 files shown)


robwoodgate:
Note for our future selves. We decided to use the `onChangeOutputsCreated` as a flag because:

- v2 consumers could only synchronously await the result. There was no 'fire and forget' option before v3.
- The `onChangeOutputsCreated` callback in combination with the `completeMelt` method allows clients to choose async.
- The `prefer-async` header only makes sense in an async operation (ie MUST be used with `onChangeOutputsCreated`)
- Change outputs **would be lost** if a consumer did `prefer-async` but did not use `onChangeOutputsCreated` for the blanks.

We did consider the case where a client might want to store the blanks, but await the operation synchronously... perhaps as a protection against network failure etc.

However, it is already possible to await the async `meltQuotePaid` event and then run the `meltCompleted` method. 

We therefore decided to "fail safe" and prevent the possiblity of lost change.

PR #380, opened 2025-10-21
refactor: compress builder pattern

Egge21M:
## Description

This is a proposal to compress the builder pattern in WalletOps.ts. Instead of redeclaring all the OutputType methods all the time we use composition and a central OutputTypeBuilder. The benefit is less repeated code. The drawback is that the API becomes slightly more verbose due to composition:

```ts
// Old API
wallet.ops.send()
.asRandom()
.keepAsDeterministic()
.run();

// New API
wallet.ops.send()
.sendOutput.asRandom()
.keepOutput.asDeterministic()
.run();
```

## Changes

- Add a OutputTypeBuilder
- Compose OutputTypeBuilder into WalletOps builders

## PR Tasks

- [x] Open PR
- [ ] run `npm run lint` --> no warnings or errors
- [ ] run `npm run format`

### DIFF
--- a/src/wallet/WalletOps.ts
+++ b/src/wallet/WalletOps.ts
@@ -60,8 +60,8 @@ export class WalletOps {
  *     	.run();
  */
 export class SendBuilder {
-	readonly sendOutputType = new OutputTypeBuilder();
-	readonly keepOutputType = new OutputTypeBuilder();
+	readonly sendOutputType = new OutputTypeBuilder(this);
+	readonly keepOutputType = new OutputTypeBuilder(this);
 	private config: SendConfig = {};
 	private offlineExact?: { requireDleq: boolean };
 	private offlineClose?: { requireDleq: boolean };
@@ -183,7 +183,7 @@ export class SendBuilder {
  *     	.run();
  */
 export class ReceiveBuilder {
-	readonly outputType = new OutputTypeBuilder();
+	readonly outputType = new OutputTypeBuilder(this);
 	private config: ReceiveConfig = {};
 
 	constructor(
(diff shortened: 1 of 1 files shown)


robwoodgate:
> Oh, great point. I missed that. We could fix that by giving `OutputTypeBuilder` a reference to the enclosing builder. Then it could return the reference and still allow chaining.
> 
> The reason why I stumbled upon this was that I wanted to separate out WalletOps.mint into mintBolt11 and mintBolt12 and realised that we need to add about 50 lines of repeating code for every Ops that we want to add in the future.

Can you not just repeat the pattern I used for MeltBuilder?

```ts
meltBolt11(quote: MeltQuoteResponse, proofs: Proof[]) {
    return new MeltBuilder(this.wallet, 'bolt11', quote, proofs);
}
meltBolt12(quote: Bolt12MeltQuoteResponse, proofs: Proof[]) {
    return new MeltBuilder(this.wallet, 'bolt12', quote, proofs);
}
```

Egge21M:
Overloads make it worse... I think trying to be smart about reusing methods for multiple things makes us loose quite some type safety as well. For example with the current implementation there is nothing to enforce a LockedMintQuote is run with a privkey set.

Right now I think the best way would be to have separate builder for bolt11, bolt11Locked and bolt12 so that we can enforce whatever required parameters (privkey) at compile time. But my heads spinning from all the type magic, so maybe I am wrong about this

Egge21M:
Yes, you are right. Let's keep it as is and revisit if it ever goes out of hand.unfortunately I tend towards premature optimisation haha

PR #382, opened 2025-10-21
V3 WalletOps support for mintbolt12

robwoodgate:
## Description

This PR adds mintBolt12 support to the MintBuilder. It updates tests to use `mintBolt11()` and `mintBolt12`.

EDIT: Includes a compile-time check for `mintBolt12` to ensure `.privkey()` has been used before `.run()`.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/wallet/WalletOps.ts
+++ b/src/wallet/WalletOps.ts
@@ -257,16 +257,12 @@ export class SendBuilder {
 			});
 		}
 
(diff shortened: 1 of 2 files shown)


Egge21M:
I like the simplicity of this, but I think we should enforce the presence of privkey at compile time. Right now consumers can use .privkey to add one, but it is not suggested by the code

Issue #376, opened 2025-10-16
v3 keep denominations are very small if no denomination is passed


Egge21M:
When relying on the default denominations for keep outputs a very small split is used.

e.g. We have 3 proofs (16, 4, 1) and send 6 sats the resulting split looks like this:
send: 4,2
keep: 1,1,1,2,2,2,2,4

Egge21M:
But isn't this why we have `proofsWeHave`. Passing this option used to enable this behavior. If it is now the default why keep that option? I get that the idea is to find more exact matches to save on fees, but I have seen many implementations out there that will preselect proofs and not pass everything to the wallet methods. In this case this would result to lots and lots of small proofs and therefore larger tokens. 

I think this should remain an opt-in feature

robwoodgate:
> the solution you propose only works under the assumption that consumers will not do any preselection on storage level (e.g. running total window in SQL to optimize networking) in which case they would have simply not have used the `proofsWeHave` of v2.

You are absolutely right - and on reflection am happy we either re-add `proofsWeHave` (if passing in two arrays of proofs is most flexible/beneficial), or an `optimizeChange` type boolean option (I don't mind the default). Can I leave it with you to decide and PR the optimal solution?

PR #384, opened 2025-10-24
V3 proofs we have

Egge21M:
# Fixes: https://github.com/cashubtc/cashu-ts/issues/376

## Description
This PR adds proofsWeHave back into Wallet.send and adds the corresponding method to SendBuilder. Now these operations will create the change with the least proofs possible if `proofsWeHave` is not explicitly passed and do optimization if it is

## Changes

- Add proofsWeHave to SendConfig
- Pass config.proofsWeHave instead of unselectedProofs in Wallet.send
- Add SendBuilder.proofsWeHave to adjust config in builder

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -85,7 +85,7 @@ export function splitAmount(
 		// Calculate how many of amt fit into remaining value
 		const requireCount = Math.floor(value / amt);
 		// Add them to the split and reduce the target value by added amounts
-		split.push(...Array(requireCount).fill(amt));
+		split.push(...Array<number>(requireCount).fill(amt));
 		value -= amt * requireCount;
 		// Break early once target is satisfied
 		if (value === 0) break;


Egge21M:
@robwoodgate FYI this also includes a tiny modification to splitAmounts. I just found the "algorithm" very difficult to read and reason about, so I replaced it with a less functional-style approach, thats more readable IMHO. Let me know if you prefer this too, otherwise I can always revert that.

src/utils/core.ts
```diff
@@ -80,12 +80,14 @@ export function splitAmount(
...
-		value %= amt;
-	});
+	for (const amt of sortedKeyAmounts) {
+		if (amt <= 0) continue;
+		while (value >= amt) {
+			split.push(amt);
+			value -= amt;
+		}
+		if (value === 0) break;
+	}
```
robwoodgate, line 92: I agree it was hard to read and reason about before, and love that you've replaced the forEach with a for..of loop, and breaking early when value is zero. Much more efficient.

The looped subtraction + push is a bit less efficient when a larger number of each key amount is needed, being O(q) vs O(1) doing it via a fill.

Wondering if a hybrid approach with comments will give us the best of both worlds?

But in practical terms, it probably doesn't matter, unless a mint only offers a 1 sat key and consumer wants to split 10,000 or more.

```suggestion
	for (const amt of sortedKeyAmounts) {
		if (amt <= 0 || value < amt) continue;
		// Calculate how many of this denomination
		// we need (q) and add them to our split array
		const q = Math.floor(value / amt);
		split.push(...Array<number>(q).fill(amt));
		// Calculate remaining value to split, or exit if done
		value -= q * amt;
		if (value === 0) break;
	}
```

src/wallet/Wallet.ts
```diff
@@ -953,7 +953,7 @@ class Wallet {
 			keyset,
 			outputConfig.keep ?? this.defaultOutputType(),
 			false,
-			unselectedProofs,
+			config?.proofsWeHave,
```
robwoodgate, line 955: The comment at the top of this block is now inaccurate and should probably be changed or removed.

```
// Uses unselectedProofs to optimize denominations if needed
```

test/integration.test.ts
```diff
@@ -214,8 +214,7 @@ describe('mint api', () => {
 		// The 32 would have been selected (fee: 1 sat), leaving 4,64 unspent
 		// Keep is optimized to _denominationTarget (default: 3) vs unselected
 		// So we expect: 1,1,1,2,2,2,4,4,4 change + 4,64 unspent = 11 proofs (total 89)
-		console.log(`sumProofs(sendResponse.keep) = ${sumProofs(sendResponse.keep)}`);
-		expect(sendResponse.keep.length).toBe(11);
+		expect(sendResponse.keep.length).toBe(5);
```
robwoodgate, line 216: The comment at the start of this block is no longer accurate:

```
// Keep is optimized to _denominationTarget (default: 3) vs unselected
// So we expect: 1,1,1,2,2,2,4,4,4 change + 4,64 unspent = 11 proofs (total 89)
```

will be something like:

```
// We expect: 1,4,16 change + 4,64 unspent = 5 proofs (total 89)
```

robwoodgate (approved):
Couple of super-minor comment nits, and a suggestion (don't mind if you use it or not).

ACK whatever you choose.

robwoodgate:
Looking good, hollywood! LGTM!

PR #395, opened 2025-11-01
MintInfo method for amountless

### LINKED ISSUE #225: Add MintInfo method for amountless

KvngMikey:
# Fixes #225

## Description
This PR updates how amount-less melt support is handled to with the current Cashu spec in https://github.com/cashubtc/nuts/pull/173

The wallet now properly detects and uses amount-less melt support per NUT-05 and the current Cashu spec, it is clean, backwards-compatible change for clients and everything conforms to the comments dropped by @Egge21M  in this closed PR https://github.com/cashubtc/cashu-ts/pull/230#issue-2755713161
...

## Changes

- Added a new `supportsAmountless` method in `MintInfo.ts` that reads from `this._mintInfo.nuts["5"].methods.melt` to determine if a mint supports amount-less melts for a given method/unit. and returns `boolean — true` if a matching method entry has `amountless: true.`
-  Moved `amountless?: boolean` in `/model/responses.ts` into `MeltMethodSettings,` matching the spec layout under `nuts[5].methods.`
-  Updated `createMeltQuoteBolt11()` in `wallet.ts` to include an `options.amountless.amount_msat` field only if the mint supports amount-less melts `(supportsAmountless() returns true)`, and the caller explicitly passes an `amount` argument.
- In `types/wallet/payload.ts' I confirmed `MeltQuotePayload` already supports `options?: MeltQuoteOptions`, no change required.

## PR Tasks

- [x] Open PR
- [x] run `npm run test` --> all project tests passed except one unrelated import failure in `base64.test.ts`.
- [x] run `npm run format`

### DIFF
--- a/test/utils/core.test.ts
+++ b/test/utils/core.test.ts
@@ -833,7 +833,7 @@ describe('invoiceHasAmountInHRP()', () => {
 			//uppercase LN — should be valid as BOLT11 is case insensitive
 			'lNbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa',
 
-			//pico invoice — should be valid as BOLT11 is case insensitive
+			//pico invoice — should be valid last digit of amount is 0
 			'lnbc9678785340p1pwmna7lpp5gc3xfm08u9qy06djf8dfflhugl6p7lgza6dsjxq454gxhj9t7a0sd8dgfkx7cmtwd68yetpd5s9xar0wfjn5gpc8qhrsdfq24f5ggrxdaezqsnvda3kkum5wfjkzmfqf3jkgem9wgsyuctwdus9xgrcyqcjcgpzgfskx6eqf9hzqnteypzxz7fzypfhg6trddjhygrcyqezcgpzfysywmm5ypxxjemgw3hxjmn8yptk7untd9hxwg3q2d6xjcmtv4ezq7pqxgsxzmnyyqcjqmt0wfjjq6t5v4khxsp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygsxqyjw5qcqp2rzjq0gxwkzc8w6323m55m4jyxcjwmy7stt9hwkwe2qxmy8zpsgg7jcuwz87fcqqeuqqqyqqqqlgqqqqn3qq9q9qrsgqrvgkpnmps664wgkp43l22qsgdw4ve24aca4nymnxddlnp8vh9v2sdxlu5ywdxefsfvm0fq3sesf08uf6q9a2ke0hc9j6z6wlxg5z5kqpu2v9wz',
 		];
 
@@ -844,7 +844,7 @@ describe('invoiceHasAmountInHRP()', () => {
 		const invalid = [
 			'lnbc0210n1...', // leading zero in amount → invalid per spec
 			'lnsomething', // incomplete HRP
-			'lnbc9678785343p1pwmna7lpp5g...', // pico invoice — case insensitive
+			'lnbc9678785343p1pwmna7lpp5g...', // pico invoice — amount not ending with 0
 		];
 
 		invalid.forEach((inv) => expect(invoiceHasAmountInHRP(inv)).toBe(false));


src/wallet/types/payloads.ts
```diff
@@ -33,9 +33,21 @@ export type MeltQuotePayload = {
...
+export type MeltQuotePayloadAmountless = {
+	/**
+	 * Optional property for specifying an amountless option.
+	 */
+	amountless?: {
+		/**
+		 * The amount in milli-satoshis.
+		 */
+		amount_msat: number;
+	};
```
robwoodgate: This seems redundant. `MeltQuoteOptions` already includes an `amountless?: AmountlessOption ` entry, and `AmountlessOption` specifies the `amount_msat`.

```suggestion
	options?: MeltQuoteOptions;
};
```

src/mint/types/responses.ts
```diff
@@ -281,6 +281,7 @@ export type SwapMethod = {
 	options?: {
 		description?: boolean;
 	};
+	amountless?: boolean | null;
 };
```
robwoodgate: Shouldn't this be inside the options [Per NUT-23](https://github.com/cashubtc/nuts/blob/main/23.md?plain=1#L252)

```suggestion
	options?: {
		description?: boolean;
		amountless?: boolean | null;
	};
};
```

src/model/MintInfo.ts
```diff
@@ -200,4 +200,14 @@ export class MintInfo {
...
 		);
 	}
+
+	supportsAmountless(method: string = 'bolt11', unit: string = 'sat'): boolean {
+		const meltMethods = this._mintInfo?.nuts?.[5]?.methods ?? [];
+
+		if (!Array.isArray(meltMethods)) return false;
+
+		return meltMethods.some(
+			(met) => met.method === method && met.unit === unit && met.amountless === true,
```
robwoodgate: [Per NUT-23](https://github.com/cashubtc/nuts/blob/main/23.md?plain=1#L252), shouldn't amountless should be in options?

```suggestion
			(met) => met.method === method && met.unit === unit && met.options?.amountless === true,
```

robwoodgate (changes requested):
Looks mostly solid @KvngMikey - I think the MintInfo `amountless` setting may be in the wrong place (as per suggestions), can you take another look to confirm.

Also, would be really useful to get some unit test coverage for this please.

KvngMikey:
thank you for reviewing @robwoodgate, i'll take a look at all and provide updates.

src/wallet/types/payloads.ts
KvngMikey: updated this

src/mint/types/responses.ts
KvngMikey: done

src/model/MintInfo.ts
KvngMikey: fixed

src/wallet/Wallet.ts
Egge21M, line 1692: TSdocs are missing `amountMsat`

Egge21M (changes requested):
Additionally to the suggestion we need to add an additonal check for the invoice amount here.
The problem here is, that users might put an invoice that is NOT amountless, but still add `amountMsat`. In this case we should throw. So we need to make sure its actually an amountless invoice and throw if the user supplied `amountMsat` but the invoice already encodes an amount.

Parsing the amount from an invoice does not require any special tools. Here is the specification: https://github.com/lightning/bolts/blob/master/11-payment-encoding.md

src/wallet/Wallet.ts
KvngMikey: fixed

src/wallet/Wallet.ts
```diff
@@ -1507,22 +1507,47 @@ class Wallet {
...
+	 * @param amountMsat Optional amount in millisatoshis to attach for amountless invoices, must not
+	 *   be provided for invoices that already encode an amount.
 	 * @returns The mint will create and return a melt quote for the invoice with an amount and fee
 	 *   reserve.
 	 */
-	async createMeltQuoteBolt11(invoice: string): Promise<MeltQuoteResponse> {
+	async createMeltQuoteBolt11(invoice: string, amountMsat?: number): Promise<MeltQuoteResponse> {
+		if (amountMsat != null) {
+			// invoice is amount-less <=> HumanReadablePart has NO amount segment
+			const hasAmountInHRP = /^ln[a-z0-9]*\d+[munp]?$/.test(invoice);
```
robwoodgate:
```suggestion
			const hasAmountInHRP = /^ln[a-z]{2,}[1-9][0-9]*(?:[mun]|0p)?1/.test(invoice);
```

This one should do the job...

robwoodgate: `^ln` - starts with prefix (literal characters)
`[a-z]{2,}` - 2 or more characters in lowercase a-z (currency)
`[1-9][0-9]*` - a positive integer with no leading zero... literally a digit 1-9 with zero or more 0-9 digits following
`(?:[mun]|0p)?` - non-capturing group (?:...) with zero or 1 of: m, u, n, or 0p (to meet pico requirement)
`1` - the 1 separator signifying end of HRP

robwoodgate: Might be an idea to move the whole regex check and throw to a utility functiion, then you can run unit tests to it without doing a full melt quote.

These are the ln invoices I tested the regex suggestion against:

```
0 amount:
lnbc1p53lqw7pp5d8ntp7kfaqcqtxfgks0n32xd4lng2hhx5z3gvfcm9teyn4vee35sdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqr8pqsp5wdg4qaq6ktrvfm4z99ry98y4qrmg3krnc4mhf2rwce230hyyeu4s9qxpqysgqgz7lt5hnxcq3wrpd5qe64a37msj0lhqfa0ky6ppagyedd79lz86zrcg20p78csjtqv3sc2m06uu24ykh8q0jzhu30yr820sysh9wv8gpz44nvz

21 sats:
lnbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa

1btc (no multiplier):
lnbc11p53lqsgpp5mxs67qwmh34wu3jy7um8u490n2dtgy7dsrhzpdup008g8ygj2eysdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5y7x6c82fj2cfd59pjkyfljlfyr06lnt0vttpss6c8ped9cpsw95s9qxpqysgquq5f3ahjwqgcwdh47ht492l9urwpnf5vyns982tev2puj644wx6s2ezdxnjetqc05gjrhwyh9dzgkdc4gshm97h0cws2az5s04mrjesq6x02q4
```

KvngMikey: @robwoodgate, you're amazing, great catch, let me work on this now and push a fix for this PR, thank you.

test/utils/core.test.ts
```diff
@@ -808,3 +813,34 @@ describe('test deprecated base64 keyset id derivation', () => {
...
+			// 1 BTC — no multiplier
+			'lnbc11p53lqsgpp5mxs67qwmh34wu3jy7um8u490n2dtgy7dsrhzpdup008g8ygj2eysdp82pshjgr5dusyymrfde4jq4mpd3kx2',
+		];
+
+		withAmount.forEach((inv) => expect(invoiceHasAmountInHRP(inv)).toBe(true));
+	});
+
+	test('rejects malformed or invalid HRP structure', () => {
+		const invalid = [
+			'lNbc210n1...', // uppercase LN (valid BOLT-11 but won’t match this HRP regex → fine)
```
robwoodgate: Bravo for spotting that uppercase is allowed in BOLT-11. I overlooked that.
But we should fix the regex, not bake in the shortcomings of the regex.

So can you move this test to the 'valid' pile, and let's add the case insensitive flag to the regex: `/i`

src/utils/core.ts
```diff
@@ -814,3 +814,11 @@ function removePrefix(token: string): string {
 	});
 	return token;
 }
+
+/**
+ * Detects whether a BOLT-11 Lightning invoice encodes a non-zero amount in the Human-Readable Part
+ * (HRP).
+ */
+export function invoiceHasAmountInHRP(invoice: string): boolean {
+	return /^ln[a-z]{2,}[1-9][0-9]*(?:[mun]|0p)?1/.test(invoice);
```
robwoodgate:
```suggestion
	return /^ln[a-z]{2,}[1-9][0-9]*(?:[mun]|0p)?1/i.test(invoice);
```

As per other comment - let's fix the regex by making it case insensitive to support uppercase characters as per BOLT-11

KvngMikey: worked on both requests.

src/utils/core.ts
```diff
@@ -814,3 +814,11 @@ function removePrefix(token: string): string {
 	});
 	return token;
 }
+
+/**
+ * Detects whether a BOLT-11 Lightning invoice encodes a non-zero amount in the Human-Readable Part
+ * (HRP).
+ */
+export function invoiceHasAmountInHRP(invoice: string): boolean {
+	return /^ln[a-z]{2,}[1-9][0-9]*(?:[mun]|0p)?1/i.test(invoice);
```
zk-bits, line 823: The regex doesn't handle `p` (pico) invoices correctly, is that intentional?  or should it be `(?:[munp])?` the 0 seems to always be consumed as part of the amount.

src/wallet/Wallet.ts
```diff
@@ -1680,22 +1662,45 @@ class Wallet {
...
+			);
+		}
+
+		const supportsAmountless = this._mintInfo?.supportsAmountless?.('bolt11', this._unit) ?? false;
+
 		const meltQuotePayload: MeltQuoteBolt11Request = {
 			unit: this._unit,
 			request: invoice,
+
+			...(supportsAmountless && amountMsat
```
zk-bits: If someone passes `amountMsat: 0`, it passes the null check in line 1683 but gets dropped in the spread (0 is falsy).

zk-bits: Should `amountMsat` be validated with `assertAmount` when provided? That would catch 0, negatives, and non-integers.

Egge21M: Using the new `assertAmount` util is a great idea!

src/wallet/Wallet.ts
```diff
@@ -1680,22 +1662,45 @@ class Wallet {
...
+	 *   be provided for invoices that already encode an amount.
 	 * @returns The mint will create and return a melt quote for the invoice with an amount and fee
 	 *   reserve.
 	 */
-	async createMeltQuoteBolt11(invoice: string): Promise<MeltQuoteBolt11Response> {
+	async createMeltQuoteBolt11(
+		invoice: string,
+		amountMsat?: number,
+	): Promise<MeltQuoteBolt11Response> {
+		if (amountMsat != null && invoiceHasAmountInHRP(invoice)) {
```
Egge21M: If `amountMsat` is not passed (evaluating to `undefined`) this branch still triggers. Instead it should check whether `amountMsat` is defined and then use the `assertAmount` util as @zk-bits mentioned.

src/utils/core.ts
robwoodgate: The pico handling is intended as bolt11 spec requires:

If the p multiplier is used the last decimal of amount MUST be 0.

https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#requirements-1

zk-bits: The regex is too greedy. Pico bolt11 invoices will not trigger. As the 0 is consumed as part of the previous statement '0p' does not trigger in any instance the 0 portion of 0p is captured by [1-9][0-9]* always

zk-bits: I was wrong. |0p works correctly, regex can be tricky, sorry!

robwoodgate: No worries, yeah, they are a bit (pun intended).

But a vaild point - we should add a pico invoice to the tests.

test/utils/core.test.ts
```diff
@@ -808,3 +813,36 @@ describe('test deprecated base64 keyset id derivation', () => {
...
+	test('detects invoices with amount', () => {
+		const withAmount = [
+			// 21 sats (210n → valid)
+			'lnbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa',
+
+			// 1 BTC — no multiplier
+			'lnbc11p53lqsgpp5mxs67qwmh34wu3jy7um8u490n2dtgy7dsrhzpdup008g8ygj2eysdp82pshjgr5dusyymrfde4jq4mpd3kx2',
+
+			//uppercase LN — should be valid as BOLT11 is case insensitive
+			'lNbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa',
```
robwoodgate, line 834: We should test the pico part too, as it's a bit special.

```suggestion
			//uppercase LN — should be valid as BOLT11 is case insensitive
			'lNbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa',
			
			//pico invoice — should be valid as BOLT11 is case insensitive
			'lnbc9678785340p1pwmna7lpp5gc3xfm08u9qy06djf8dfflhugl6p7lgza6dsjxq454gxhj9t7a0sd8dgfkx7cmtwd68yetpd5s9xar0wfjn5gpc8qhrsdfq24f5ggrxdaezqsnvda3kkum5wfjkzmfqf3jkgem9wgsyuctwdus9xgrcyqcjcgpzgfskx6eqf9hzqnteypzxz7fzypfhg6trddjhygrcyqezcgpzfysywmm5ypxxjemgw3hxjmn8yptk7untd9hxwg3q2d6xjcmtv4ezq7pqxgsxzmnyyqcjqmt0wfjjq6t5v4khxsp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygsxqyjw5qcqp2rzjq0gxwkzc8w6323m55m4jyxcjwmy7stt9hwkwe2qxmy8zpsgg7jcuwz87fcqqeuqqqyqqqqlgqqqqn3qq9q9qrsgqrvgkpnmps664wgkp43l22qsgdw4ve24aca4nymnxddlnp8vh9v2sdxlu5ywdxefsfvm0fq3sesf08uf6q9a2ke0hc9j6z6wlxg5z5kqpu2v9wz',
```

test/utils/core.test.ts
```diff
@@ -808,3 +813,36 @@ describe('test deprecated base64 keyset id derivation', () => {
...
+		];
+
+		withAmount.forEach((inv) => expect(invoiceHasAmountInHRP(inv)).toBe(true));
+	});
+
+	test('rejects malformed or invalid HRP structure', () => {
+		const invalid = [
+			'lnbc0210n1...', // leading zero in amount → invalid per spec
+			'lnsomething', // incomplete HRP
```
robwoodgate, line 846:
```suggestion
			'lnbc0210n1...', // leading zero in amount → invalid per spec
			'lnsomething', // incomplete HRP
			'lnbc9678785343p1pwmn', // pico not ending with 0
```

src/model/types/NUT06.ts
```diff
@@ -91,6 +91,7 @@ export type SwapMethod = {
 	description?: boolean; //added this for Nutshell =>0.16.4 compatibility, see https://github.com/cashubtc/nutshell/pull/783
 	options?: {
 		description?: boolean;
+		amountless?: boolean | null;
```
robwoodgate: Should this have a `null` here? I think it should probably just be:
```suggestion
		amountless?: boolean;
```

as the key is already optional. Ref:
https://github.com/cashubtc/nuts/blob/main/23.md#melt-quote
https://github.com/cashubtc/nuts/blob/main/23.md#melt-quote

src/wallet/Wallet.ts
robwoodgate: The check is loose equality `!=`, so although it looks like it shouldn't pass, it will, because `undefined` and `null` are equivalent in loose equality.

So it's technically correct, but would be clearer as:

```ts
if (amountMsat != undefined && invoiceHasAmountInHRP(invoice)) {
```

That said, amount `0` is a nonsense input... so we should decide whether to throw or silently discard it. If we change the test to a truthy one, it will pass the check and silently discard `0` at the spread:

```ts
if (amountMsat && invoiceHasAmountInHRP(invoice)) {
```

Assuming we want to throw on `0` and use `assertAmount()` as @zk-bits  suggests, we can do the HRP check and throw the general assert for zero, non-integer, negative and other nonsense.

Something iike:

```ts
if (amountMsat != undefined) {
	if (invoiceHasAmountInHRP(invoice)) {
		throw new Error(
			'amountMsat supplied but invoice already contains an amount, Leave amountMsat undefined for non-zero invoices.',
		);
	}
	assertAmount(amountMsat);
}
```

robwoodgate: In fact, as we are in Wallet, you can lean on the `this.failIf()` which does logging plus throw:

```ts
if (amountMsat != undefined) {
	this.failIf(invoiceHasAmountInHRP(invoice), 'amountMsat supplied but invoice already contains an amount. Leave amountMsat undefined for non-zero invoices.');
	assertAmount(amountMsat);
}
```

test/utils/core.test.ts
```diff
@@ -808,3 +813,40 @@ describe('test deprecated base64 keyset id derivation', () => {
...
+			// 21 sats (210n → valid)
+			'lnbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa',
+
+			// 1 BTC — no multiplier
+			'lnbc11p53lqsgpp5mxs67qwmh34wu3jy7um8u490n2dtgy7dsrhzpdup008g8ygj2eysdp82pshjgr5dusyymrfde4jq4mpd3kx2',
+
+			//uppercase LN — should be valid as BOLT11 is case insensitive
+			'lNbc210n1p53lq0wpp5tsmnj3c6znsdyu5v8t2k3y8xw33m9hnd6exzwspxa4pqz3hze8rsdp82pshjgr5dusyymrfde4jq4mpd3kx2apq24ek2uscqzpuxqrwzqsp5jgr8l0yx8zpxfez9hns5t25j9m90yrzjz34gpacssd6lwr7an40q9qxpqysgqws7g2g9hh6awk2n6vhzpqjyf6matulx0cc0ct099nz6kudzv8xmy9clu4kyvurrt99zkr7y03hse85c2jvm7jm8qlqnvzawudn4e3vsq0m6qpa',
+
+			//pico invoice — should be valid as BOLT11 is case insensitive
```
robwoodgate:
```suggestion
			//pico invoice — should be valid as last digit of amount is 0
```

test/utils/core.test.ts
```diff
@@ -808,3 +813,40 @@ describe('test deprecated base64 keyset id derivation', () => {
...
+		];
+
+		withAmount.forEach((inv) => expect(invoiceHasAmountInHRP(inv)).toBe(true));
+	});
+
+	test('rejects malformed or invalid HRP structure', () => {
+		const invalid = [
+			'lnbc0210n1...', // leading zero in amount → invalid per spec
+			'lnsomething', // incomplete HRP
+			'lnbc9678785343p1pwmna7lpp5g...', // pico invoice — case insensitive
```
robwoodgate:
```suggestion
			'lnbc9678785343p1pwmna7lpp5g...', // pico invoice — amount not ending with 0
```

Issue #394, opened 2025-10-31
Add arbitrary tags to NUT-11 (P2PK) secrets


robwoodgate:
We could modify the `P2PKOptions` to allow a user to add arbitrary tags to the P2PK secret.

`OutputData.create[Single]P2PKData` can then build these tags into the secret.

This would be useful, for example, to add an `e` (event) tag for NutZaps.

robwoodgate:
This can perhaps be done as part of the upcoming P2BK implementation, as that already refactors OutputData.

PR #396, opened 2025-11-01
V3: Adds support for arbitrary NUT11 secret tags

robwoodgate:
# Fixes: #394 

## Description

Adds support for up to 5 arbitrary NUT-11 secret tags, allowing (for example) NutZap proofs to have `e` and `p` tags.

## Changes

- Adds AddTag/AddTags methods to P2PKBuilder
- Adds additionalTags option in P2PKOptions
- Adds tag handling in OutputData,
- Includes test coverage and README update
- createSingleP2PKData refactor from ECDH brought forward to this PR

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/OutputData.ts
+++ b/src/model/OutputData.ts
@@ -139,8 +139,9 @@ export class OutputData implements OutputDataLike {
 		// build P2PK Tags (NUT-11)
 		const tags: string[][] = [];
 
-		if (typeof p2pk.locktime === 'number') {
-			tags.push(['locktime', String(p2pk.locktime)]);
+		const ts = p2pk.locktime ?? NaN;
+		if (Number.isSafeInteger(ts) && ts >= 0) {
+			tags.push(['locktime', String(ts)]);
 		}
 
 		if (pubkeys.length > 0) {


robwoodgate:
As discussed in the meeting today, the arbitrary 5 tag limit has been removed and a max character length check introduced to ensure compatiblity with Nutshell default secret length limits.

etc/cashu-ts.api.md
```diff
@@ -991,25 +991,13 @@ export class OutputData implements OutputDataLike {
...
     static createDeterministicData(amount: number, seed: Uint8Array, counter: number, keyset: MintKeys | Keyset, customSplit?: number[]): OutputData[];
     // (undocumented)
-    static createP2PKData(p2pk: {
-        pubkey: string | string[];
-        locktime?: number;
-        refundKeys?: string[];
-        requiredSignatures?: number;
-        requiredRefundSignatures?: number;
-    }, amount: number, keyset: MintKeys | Keyset, customSplit?: number[]): OutputData[];
+    static createP2PKData(p2pk: P2PKOptions, amount: number, keyset: MintKeys | Keyset, customSplit?: number[]): OutputData[];
```
Egge21M, line 994: Review Note: This is not a breaking change, as P2PKOptions is broader than the original literal

etc/cashu-ts.api.md
```diff
@@ -1467,6 +1460,9 @@ export type SwapTransaction = {
     sortedIndices: number[];
 };
 
+// @public (undocumented)
+export type TagTuple = [key: string, ...values: string[]];
```
Egge21M: naming nit (non blocking): Calling it tuple might be confusing as tags are arbitrary length and are mutabable. But I think this is fine and leaving it up to you.

src/model/OutputData.ts
```diff
@@ -92,58 +108,84 @@ export class OutputData implements OutputDataLike {
...
-		keysetId: string,
-	) {
-		// Standardize pubkey (backwards compat), clamp n_sigs between 1 and total pubkeys
-		// clamp n_sigs_refund between 1 and total refundKeys, and create secret
-		const pubkeys: string[] = Array.isArray(p2pk.pubkey) ? p2pk.pubkey : [p2pk.pubkey];
-		const n_sigs: number = Math.max(1, Math.min(p2pk.requiredSignatures || 1, pubkeys.length));
-		const n_sigs_refund: number = Math.max(
+	static createSingleP2PKData(p2pk: P2PKOptions, amount: number, keysetId: string) {
+		// normalise keys and clamp required signature counts to available keys
+		const lockKeys: string[] = Array.isArray(p2pk.pubkey) ? p2pk.pubkey : [p2pk.pubkey];
```
Egge21M, line 126: We should guard against `p2pk.pubkey = []`. This is valid, as `[]` satisfies `string[]` and can lead to unspendable ecash (or ecash that can be spent by anyone... not sure how mint handle this).

src/model/OutputData.ts
```diff
@@ -92,58 +108,84 @@ export class OutputData implements OutputDataLike {
...
+		const refundKeys: string[] = p2pk.refundKeys ?? [];
+		const reqLock = Math.max(1, Math.min(p2pk.requiredSignatures ?? 1, lockKeys.length));
+		const reqRefund = Math.max(
 			1,
-			Math.min(p2pk.requiredRefundSignatures || 1, p2pk.refundKeys ? p2pk.refundKeys.length : 1),
+			Math.min(p2pk.requiredRefundSignatures ?? 1, refundKeys.length || 1),
```
Egge21M, line 131: This will lead to `reqRefund` be at least 1, even if `requiredRefundSignatures` is `undefined` and `refundKeys` is `[]`. Is that the behavior that we expect? If not, I'd suggest replacing the `||` with `??` to make sure 0 makes it through.

Edit: I noticed later on that we only add the flag if `reqRefund` is greater than 1. So this might be fine

src/model/OutputData.ts
```diff
@@ -92,58 +108,84 @@ export class OutputData implements OutputDataLike {
...
+
+		// Init vars
+		const data = lockKeys[0];
+		const pubkeys = lockKeys.slice(1);
+		const refund = refundKeys;
+
+		// build P2PK Tags (NUT-11)
+		const tags: string[][] = [];
+
+		if (p2pk.locktime !== undefined) {
```
Egge21M:
```suggestion
		if (typeof p2pk.locktime === 'number') {
```

This guards from null passing this check (useful for JS consumers or consumers that pass in JSON parsed data without validation)

src/wallet/P2PKBuilder.ts
```diff
@@ -52,6 +54,21 @@ export class P2PKBuilder {
 		return this;
 	}
 
+	addTag(key: string, values?: string[] | string) {
+		if (!key || typeof key !== 'string') throw new Error('tag key must be a non empty string');
```
Egge21M: nit: could be abstracted into a util as its non-DRY with the code in OutputData

etc/cashu-ts.api.md
robwoodgate: The word 'Tuple" hurts my brain for some reason, so more than happy to change it. How about "P2PKTag"

src/model/OutputData.ts
robwoodgate: This is purely a normalization step. Later on, it is guarded at use time with`if (pubkeys.length > 0) {...}`

src/model/OutputData.ts
robwoodgate: Again, purely a normalization step at this point in the flow.. `reqRefund` = 0 makes no sense if `refund` tag is provided. And is not required if `refund` tag is not provided. In other words, it's always got to be `>= 1` if set.

And yes, later on we only add it if a) we have refund tag and b) `reqRefund` > 1 (as the default assertion in NUT-11 spec is _"at least ONE of the refund pubkeys"_).

src/wallet/P2PKBuilder.ts
robwoodgate: good catch... I wrote the P2PKBuilder first, then thought I'd better guard for direct creation in OutputData. I'll abstract it.

src/model/OutputData.ts
robwoodgate: Good catch @Egge21M  - I was incorrectly leaning on the type guard, and missed the runtime. I've applied your suggestion as the minimel fix, but do we want to go further with the runtime check?
```ts
		const lt = p2pk.locktime;
		if (typeof lt === 'number' && Number.isSafeInteger(lt) && lt > 0 && lt < 1e10) {
```

src/wallet/P2PKBuilder.ts
robwoodgate: Abstracted to `assertValidTagKey()`

etc/cashu-ts.api.md
robwoodgate: Renamed to `P2PKTag`

src/model/OutputData.ts
Egge21M: Thank you! I think more validation is not strictly necessary, but when we are at it anyways, why not? What about adding a check for valid Date?

Number.isSafeInteger(ls) && ts >= 0 && !isNaN(new Date(ts * 1000))

I think we can get rid of the typeof check because `isSafeInteger` will take care of that too

Issue #369, opened 2025-10-09
CBOR library guardrails


robwoodgate:
Our current [cbor.ts library](https://github.com/cashubtc/cashu-ts/blob/main/src/cbor.ts) is a slimmed down, Cashu-specific implementation of CBOR.

We should perhaps consider adding some guardrails (eg throws) for operations which are out of bound for Cashu.

Two immediate examples:

- `encodeObject` can only handle objects with < 24 keys

it encodes the key count as an unsigned integer, then OR’s the last pushed byte with 0xa0. That only works when the length is less than 24, because `encodeUnsigned` writes a single initial byte in that case. For lengths 24 or greater, `encodeUnsigned` writes two or more bytes, and you are OR’ing the length byte, not the initial major type byte, which corrupts the length.

We could either support larger objects, or throw if keys >23 to avoid silent corruption.

- Numbers are always encoded as unsigned in `encodeItem`

Again, we could either support negative numbers or throw if not expected/supported (the current case).

After brief discussion with @Egge21M  - we could also consider installing a full implementation of CBOR as a devDependency and figure out some tests that will warn us if our slimmed down version goes out of bounds in the future.

PR #398, opened 2025-11-05
Chore/cbor guardrails

TheRealCheebs:
# Fixes: #369 

## Description
Added additional guardrails and filled in testing

- Fix map encoding bug: write correct major-type 5 initial byte and length bytes (previous approach OR'd the wrong byte for lengths >= 24 and corrupted output). Add guardrail to throw if map key count >= 2**32.

- Implement robust number encoding:
  - Add encodeNumber to dispatch integer vs float encoding.
  - Add encodeSigned for CBOR major type 1 (negative ints).
  - Add encodeFloat64 for non-integer JS numbers.

- Preserve and clarify unsigned integer encoding (encodeUnsigned). Add explicit size limits and consistent error messages for unsupported sizes.

- Add bounds checks to decoding:
  - Add ensureAvailable checks before DataView reads to normalize truncated-input errors to "Unexpected end of data" instead of leaking DataView RangeError.

- Improve decoder robustness:
  - decodeLength correctly handles 1/2/4/8-byte length forms with bounds checks.
  - decodeSimpleAndFloat uses bounds checks for 1/2/4/8 byte reads and reports unknown simple/float values consistently.

- Add many focused tests to exercise enc/dec code paths and error cases:
  - Map encoding: 0..23 (short), 24, 255 (1-byte), 256 (2-byte), 65536 (4-byte).
  - Array header forms and lengths (24, 256).
  - Byte string & text string header forms (24, 256, 65536).
  - Negative integer encodings: -1, -10, -25 (1-byte), -257 (2-byte), -1000, -65537 (4-byte).
  - Non-integer numbers encoded as float64 (1.5, NaN, Infinity).
  - float16 decoding: subnormal, +/-1.0, Infinity, NaN, negative variants to exercise sign branch.
  - decodeSimpleAndFloat error branches (unknown simple / unknown simple or float).
  - decodeLength 8-byte length branch (hi/lo small and hi/lo too-large -> length-exceeds).
  - "Unexpected end of data" tests for empty and truncated buffers.
  - Decode error cases: unsupported major type, invalid map key type, byte/string overflow.
  - Encode error cases: unsupported JS types (Symbol, function, BigInt), encode-side "too long" guards for byte string / string / array, and integer size errors.
  - Guardrail test for object with >= 2**32 keys (monkeypatched Object.keys to avoid huge allocation).

- Result: fixes silent corruption bugs for larger maps/numbers, normalizes error behavior on truncated input, and significantly increases test coverage for CBOR encode/decode logic.

Notes:
- The decoder now throws consistent 'Unexpected end of data' for truncated inputs. If you prefer preserving native RangeError types, we can switch to throwing RangeError('Unexpected end of data') instead.
- Some tests create 65k-length inputs to exercise 4-byte length code paths; these are reasonably small (~128 KB) but can be replaced with fakes if you prefer avoiding large allocations.

## Changes
Files changed:
- src/cbor.ts     (encoding fixes, signed/float handling, bounds checks)
- test/cbor.test.ts (many new unit tests covering enc/dec branches and errors)

## PR Tasks

- [x] Open PR
- [x] run `npm run format`
- [x] run `npm run api:check` --> run `npm run api:update` for changes to the API (N/A)


src/utils/cbor.ts
```diff
@@ -62,6 +62,52 @@ function encodeUnsigned(value: number, buffer: number[]) {
...
+			0x3a,
+			unsigned >> 24,
+			(unsigned >> 16) & 0xff,
+			(unsigned >> 8) & 0xff,
+			unsigned & 0xff,
+		);
+	} else {
+		throw new Error('Unsupported integer size');
+	}
+}
```
robwoodgate, line 136: Wondering if we should mask the high byte in the 32 bit branches to avoid relying on Uint8Array clamping?

```ts
function encodeUnsigned(value: number, buffer: number[]) {
	if (value < 24) {
		buffer.push(value);
	} else if (value < 256) {
		buffer.push(0x18, value);
	} else if (value < 65536) {
		buffer.push(0x19, (value >> 8) & 0xff, value & 0xff);
	} else if (value < 4294967296) {
		buffer.push(
			0x1a,
			(value >>> 24) & 0xff,
			(value >>> 16) & 0xff,
			(value >>> 8) & 0xff,
			value & 0xff,
		);
	} else {
		throw new Error('Unsupported integer size');
	}
}

function encodeSigned(value: number, buffer: number[]) {
	// CBOR negative integer encoding: store -1 - value as unsigned under major type 1
	const unsigned = -1 - value;
	if (unsigned < 24) {
		buffer.push(0x20 | unsigned);
	} else if (unsigned < 256) {
		buffer.push(0x38, unsigned);
	} else if (unsigned < 65536) {
		buffer.push(0x39, (unsigned >> 8) & 0xff, unsigned & 0xff);
	} else if (unsigned < 4294967296) {
		buffer.push(
			0x3a,
			(unsigned >>> 24) & 0xff,
			(unsigned >>> 16) & 0xff,
			(unsigned >>> 8) & 0xff,
			unsigned & 0xff,
		);
	} else {
		throw new Error('Unsupported integer size');
	}
}
```

I didn't add as a suggestion because it wasn't in the original cbor (encodeUnsigned is unchanged in this PR)

test/utils/cbor.test.ts
```diff
@@ -202,18 +208,430 @@ const tests = [
...
+		}) as any;
+		(globalThis as any).TextEncoder = function () {
+			return { encode: (s: string) => ({ length: 4294967296 }) as any };
+		};
+		try {
+			expect(() => encodeCBOR('x' as any)).toThrow(/String too long to encode/);
+		} finally {
+			(globalThis as any).TextEncoder = RealTextEncoder;
+		}
+	});
```
robwoodgate, line 458: Test coverage is already 100%, so not strictly necessary, but would catch  any future endianness or off by one mistakes.

```suggestion
	});
	
	test('encodes max 32 bit unsigned integer and roundtrips', () => {
		const n = 0xffffffff; // 4294967295
		const enc = encodeCBOR(n as any);
		expect(Array.from(enc)).toEqual([0x1a, 0xff, 0xff, 0xff, 0xff]);
		expect(decodeCBOR(enc)).toBe(n);
	});
	
	test('encodes max 32 bit unsigned integer and roundtrips', () => {
		const n = 0xffffffff; // 4294967295
		const enc = encodeCBOR(n as any);
		expect(Array.from(enc)).toEqual([0x1a, 0xff, 0xff, 0xff, 0xff]);
		expect(decodeCBOR(enc)).toBe(n);
	});
```

robwoodgate (approved):
@TheRealCheebs - awesome work, and lovely code. Thank you!

Looks like you started this in v2, as the `cbor.ts` file is in the `src` directory (it's moved in v3 to `src/utils`). Please can you rebase to the `development` branch.

I've tried it locally, and it's a clean rebase, but if you need assistance LMK.

TheRealCheebs:
Thanks for catching all of that @robwoodgate! I hadn't fetched the development branch despite starting my worktree from there. I'm also pro fixing things when we are in here so the clamping comment is great.

In going through the reference tests the current implementation failed a ton. I went ahead and explicitly called out what was and wasn't covered in this implementation at the top of the file. Please check that. 

I thought it best to keep the entire appendix_a.json and call out what was being skipped in the test. Prior to this change the reference tests had been cherrypicked at the top. As someone new to cbor all of that was "magic" and even without me asking AI what they each tested its all magic but I think the comments and exclusions in that test help. 

My hope is that the comments help the next person to look at the file.

robwoodgate:
> Thanks for catching all of that @robwoodgate! I hadn't fetched the development branch despite starting my worktree from there. I'm also pro fixing things when we are in here so the clamping comment is great.
> 
> In going through the reference tests the current implementation failed a ton. I went ahead and explicitly called out what was and wasn't covered in this implementation at the top of the file. Please check that.
> 
> I thought it best to keep the entire appendix_a.json and call out what was being skipped in the test. Prior to this change the reference tests had been cherrypicked at the top. As someone new to cbor all of that was "magic" and even without me asking AI what they each tested its all magic but I think the comments and exclusions in that test help.
> 
> My hope is that the comments help the next person to look at the file.

Don't worry, CBOR is pretty much magic to me too... haha... the clamping suggestion was from an AI analysis.

Amazing to have the official test suite in place. And I agree with your approach - keep and call out deviations (I'll try and dive into your amendments today if possible).

Fior future reference, if a test requires node only features, you can simply add `.node.` to the test file name - eg: `cbor.node.test.ts`. Our vite config will then skip browser testing for that file. Likewise you can use `.browser.` in the name if the test suite is browser only (we don't have any of those).

test/utils/cbor.test.ts
```diff
@@ -14,207 +17,448 @@ function base64urlDecode(str: string): Uint8Array {
...
+		expect(dec.length).toBe(len);
+		expect(dec.slice(0, 3)).toBe('aaa');
+	});
+
+	test('encodes max 32 bit unsigned integer and roundtrips', () => {
+		const n = 0xffffffff; // 4294967295
+		const enc = encodeCBOR(n as any);
+		expect(Array.from(enc)).toEqual([0x1a, 0xff, 0xff, 0xff, 0xff]);
+		expect(decodeCBOR(enc)).toBe(n);
+	});
```
robwoodgate, line 305: Sorry @TheRealCheebs - in my last suggestion I pasted the max test twice. here is the "missing" min clamp test if you want it.

```suggestion
	});
	
	test('encodes min 32 bit unsigned integer and roundtrips', () => {
		const n = 0x00010000; // 65536
		const enc = encodeCBOR(n as any);
		expect(Array.from(enc)).toEqual([0x1a, 0x00, 0x01, 0x00, 0x00]);
		expect(decodeCBOR(enc)).toBe(n);
	});
```

robwoodgate:
ACK. Great work @TheRealCheebs

PR #404, opened 2025-11-10
feat: add getTokenMetadata (v3)

Egge21M:
## Description

This adds a new core util that decodes a encoded Token an retrieves metadata from it (amount, unit, mint, memo).

## Changes

- add getTokenMetadata util
- add TokenMetadata type

## PR Tasks

- [x] Open PR
- [x] run `npm run test` --> no failing unit tests
- [x] run `npm run lint` --> no warnings or errors
- [x] run `npm run format`
- [x] run `npm run api:check` --> run `npm run api:update` for changes to the API

PR #405, opened 2025-11-10
feat: add getTokenMetadata

Egge21M:
## Description

v2 impl. of https://github.com/cashubtc/cashu-ts/pull/404
See v3 PR for more info

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -991,6 +991,7 @@ export type TokenMetadata = {
     memo?: string;
     mint: string;
     amount: number;
+    incompleteProofs: Array<Omit<Proof, 'id'>>;
 };
 
 // @public @deprecated (undocumented)
--- a/src/model/types/wallet/tokens.ts
+++ b/src/model/types/wallet/tokens.ts
@@ -153,5 +153,5 @@ export type TokenMetadata = {
 	/**
 	 * Proofs without an id.
 	 */
-	incompleteProofs: Omit<Proof, 'id'>[];
+	incompleteProofs: Array<Omit<Proof, 'id'>>;
 };


src/utils.ts
```diff
@@ -373,20 +374,38 @@ function tokenFromTemplate(template: TokenV4Template): Token {
...
+		mint: tokenObj.mint,
+		amount: sumProofs(tokenObj.proofs),
+		incompleteProofs: tokenObj.proofs.map((p) => ({
+			secret: p.secret,
+			C: p.C,
+			amount: p.amount,
+			...(p.dleq && {
+				dleq: p.dleq,
+			}),
+		})),
```
robwoodgate, line 404: I think it would be simpler and more future-proof if we just isolate and extract the id, rather than naming fields to keep - this is a pattern I've used in v3 (eg for DLEQ removal):

```suggestion
		incompleteProofs: tokenObj.proofs.map((p) => {
			const { id, ...rest } = p; // extract keyset id
			void id; // intentionally unused
			return rest;
		}),
```

Egge21M: I went back and forth on this one, as changing it to the rest approach would mean changes to the proof struct would "bubble" up into this. This can be nice, if we want it, but if there are things in the future that should not be part of this we would need to remember to remove them here. So it goes in both directions. The current code enforces a clear contract that is resistent to changes

robwoodgate: True, and a valid approach. My only thought is that in this case, we are explicitly removing the keyset ID because it can't be calculated properly.

We COULD return any other property in Proof - the "contract" is a proof without keyset iD.

And removing what we don't want is more explicit that including what we know about today.

robwoodgate: But honestly fine either way. I just like the simplicity of extraction - it's more beautiful haha

PR #412, opened 2025-11-17
fix: make sure v2 keysets can get selected activeKeyset

Egge21M:
## Description
This fixes an issue with `getActiveKeyset` in v2 CashuWallet. This function filtered all keyset ids that did not start with `00`. So v2 keysets could never get elected active keysets.

## Changes

- Use isBase64String to filter out base64 keyset ids
- Adjusted isBase64String to not misidentify matching hex strings
- Added tests

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/CashuWallet.ts
+++ b/src/CashuWallet.ts
@@ -42,6 +42,7 @@ import {
 	getDecodedToken,
 	getKeepAmounts,
 	hasValidDleq,
+	isValidHex,
 	splitAmount,
 	stripDleq,
 	sumProofs,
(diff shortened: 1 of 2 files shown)


src/base64.ts
```diff
@@ -39,6 +39,11 @@ function base64urlFromBase64(str: string) {
 function isBase64String(s: string): boolean {
 	if (typeof s !== 'string' || s.length === 0) return false;
 
+	// Check if it's a hex string (keyset IDs are hex, not base64)
+	// This prevents hex strings like '00500550f0494146' from being misidentified as base64
+	const isHex = /^[a-fA-F0-9]+$/.test(s);
+	if (isHex) return false;
+
```
robwoodgate: This is the wrong place for this kind of check. It will create false positives for valid base64 strings. We should do keyset specific stuff closer to the keyset edge

Egge21M: Yeah, you are absolutely right. I removed the check from the base64 util and added an inline check for hex in the wallet method.

robwoodgate: Perfect!

robwoodgate (approved):
LGTM  @  c4ea760afb8a21ef9516d79f909f7cdbfd8d24c2

Issue #386, opened 2025-10-24
Add support for SIG_ALL in P2PK (NUT-11)


robwoodgate:
We've been holding off doing this as the specification was only implemented in mints recently (~ 2 months), and we were focusing on v3.

Once the latest amendment to specification is finalized (https://github.com/cashubtc/nuts/pull/302), and mints have adopted, we should add support to Cashu-TS v3.

PR #413, opened 2025-11-17
Adds SIG_ALL handling, transaction preview and generic melt

robwoodgate:
# Fixes: #386 

## Description

This PR adds "prepare" and "complete" steps for send, receive and melt to allow SIG_ALL signing and transaction preview. It also adds a strongly typed generic melt API, and cleans up legacy blinded message signing.

## Changes

- Introduces "prepare" and "complete" methods for send, receive and melt that build inputs, outputs and a typed preview structure before signing.
- Extends the wallet `signP2PKProofs` method to detect SIG_ALL inputs, enforce `assertSigAllInputs`, and sign all three (legacy, interim and final) SIG_ALL message formats onto the first proof.
- Adds NUT 11 helpers for SIG_ALL aggregation, including: `buildLegacyP2PKSigAllMessage`, `buildInterimP2PKSigAllMessage`, `buildP2PKSigAllMessage` and `assertSigAllInputs`.
- Hardens `verifyP2PKSig` and `hasP2PKSignedProof` with guards so SIG_ALL proofs cannot be improperly checked.
- Tides up legacy blinded message signing paths to use the new OutputDataLike based flow and improves logging around optional, non fatal signature attempts.
- Replaces `MeltBlanks` with the new `MeltPreview` model so melts now have an explicit preview phase, updating onChangeOutputsCreated, meltBlanksCreated, and completeMelt to use MeltPreview and support SIG_ALL signing.

## TODO

- Waiting for [CDK 0.13.5+ ](https://github.com/cashubtc/cdk/releases)to fix SIG_ALL signing (This is why CDK test fails)
- HTLC handling still pending (waiting for clarification around [locktime treatment](https://github.com/cashubtc/nuts/issues/310))
- Refactor tests to accomodate NUT05 shape tests ([Mint, L437-442](https://github.com/robwoodgate/cashu-ts/blob/v3-sigall/src/mint/Mint.ts#L434-L442))

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/crypto/core.ts
+++ b/src/crypto/core.ts
@@ -1,7 +1,7 @@
 import { type WeierstrassPoint } from '@noble/curves/abstract/weierstrass';
 import { secp256k1 } from '@noble/curves/secp256k1';
 import { sha256 } from '@noble/hashes/sha2';
-import { randomBytes, bytesToHex, hexToBytes } from '@noble/curves/utils';
+import { randomBytes, bytesToHex, hexToBytes, type PrivKey } from '@noble/curves/utils';
 import { Bytes, bytesToNumber, hexToNumber, encodeBase64toUint8 } from '../utils';
 import { type P2PKWitness } from '../model/types';
 
@@ -11,11 +11,16 @@ export type BlindSignature = {
 	id: string;
 };
 
-export type BlindMessage = {
+export type RawBlindedMessage = {
 	B_: WeierstrassPoint<bigint>;
 	r: bigint;
 	secret: Uint8Array;
 };
+
+/**
+ * @deprecated - Use RawBlindedMessage.
+ */
+export type BlindedMessage = RawBlindedMessage;
 
 export type DLEQ = {
 	s: Uint8Array; // signature
@@ -97,14 +102,22 @@ export function createBlindSignature(
 }
 
 /**
+ * @deprecated - Use createRandomRawBlindedMessage()
+ */
+export function createRandomBlindedMessage(_deprecated?: PrivKey): BlindedMessage {
+	void _deprecated; // intentionally unused
+	return createRandomRawBlindedMessage();
+}
+
+/**
  * Creates a random blinded message.
  *
  * @remarks
  * The secret is a UTF-8 encoded 64-character lowercase hex string, generated from 32 random bytes
  * as recommended by NUT-00.
- * @returns A BlindedMessage: {B_, r, secret}
+ * @returns A RawBlindedMessage: {B_, r, secret}
  */
-export function createRandomBlindMessage(): BlindMessage {
+export function createRandomRawBlindedMessage(): RawBlindedMessage {
 	const secretStr = bytesToHex(randomBytes(32)); // 64 char ASCII hex string
 	const secretBytes = new TextEncoder().encode(secretStr); // UTF-8 of the hex
 	return blindMessage(secretBytes);
@@ -115,9 +128,9 @@ export function createRandomBlindMessage(): BlindMessage {
  *
  * @param secret A UTF-8 byte encoded string.
  * @param r Optional. Deterministic blinding scalar to use (eg: for testing / seeded)
- * @returns A BlindedMessage: {B_, r, secret}
+ * @returns A RawBlindedMessage: {B_, r, secret}
  */
-export function blindMessage(secret: Uint8Array, r?: bigint): BlindMessage {
+export function blindMessage(secret: Uint8Array, r?: bigint): RawBlindedMessage {
 	const Y = hashToCurve(secret);
 	if (!r) {
 		r = bytesToNumber(createRandomSecretKey());
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -82,15 +82,11 @@ export type BlindAuthMintResponse = {
     signatures: SerializedBlindedSignature[];
 } & ApiError;
 
-// @public (undocumented)
-export type BlindMessage = {
-    B_: WeierstrassPoint<bigint>;
-    r: bigint;
-    secret: Uint8Array;
-};
+// @public @deprecated (undocumented)
+export type BlindedMessage = RawBlindedMessage;
 
 // @public
-export function blindMessage(secret: Uint8Array, r?: bigint): BlindMessage;
+export function blindMessage(secret: Uint8Array, r?: bigint): RawBlindedMessage;
 
 // @public (undocumented)
 export type BlindSignature = {
@@ -215,8 +211,11 @@ export function createNewMintKeys(pow2height: IntRange<0, 65>, seed?: Uint8Array
 // @public (undocumented)
 export const createP2PKsecret: (pubkey: string) => string;
 
+// @public @deprecated (undocumented)
+export function createRandomBlindedMessage(_deprecated?: PrivKey): BlindedMessage;
+
 // @public
-export function createRandomBlindMessage(): BlindMessage;
+export function createRandomRawBlindedMessage(): RawBlindedMessage;
 
 // @public (undocumented)
 export function createRandomSecretKey(): Uint8Array<ArrayBufferLike>;
@@ -1238,6 +1237,13 @@ export type ProofState = {
     Y: string;
     state: CheckStateEnum;
     witness: string | null;
+};
+
+// @public (undocumented)
+export type RawBlindedMessage = {
+    B_: WeierstrassPoint<bigint>;
+    r: bigint;
+    secret: Uint8Array;
 };
 
 // @public (undocumented)
--- a/test/crypto/NUT12.test.ts
+++ b/test/crypto/NUT12.test.ts
@@ -10,7 +10,7 @@ import {
 	verifyDLEQProof,
 	verifyDLEQProof_reblind,
 	constructProofFromPromise,
-	createRandomBlindMessage,
+	createRandomRawBlindedMessage,
 } from '../../src/crypto';
 
 describe('test hash_e', () => {
@@ -33,7 +33,7 @@ describe('test DLEQ scheme', () => {
 		const mintPubKey = pointFromBytes(secp256k1.getPublicKey(mintPrivKey, true));
 
 		// Wallet(Alice)
-		const blindMessage = createRandomBlindMessage();
+		const blindMessage = createRandomRawBlindedMessage();
 
 		// Mint
 		const blindSignature = createBlindSignature(blindMessage.B_, mintPrivKey, 1, '');
@@ -48,7 +48,7 @@ describe('test DLEQ scheme', () => {
 		const mintPubKey = pointFromBytes(secp256k1.getPublicKey(mintPrivKey, true));
 
 		// Wallet(Alice)
-		const blindMessage = createRandomBlindMessage();
+		const blindMessage = createRandomRawBlindedMessage();
 
 		// Mint
 		const blindSignature = createBlindSignature(blindMessage.B_, mintPrivKey, 1, '');
--- a/test/crypto/core.test.ts
+++ b/test/crypto/core.test.ts
@@ -6,7 +6,7 @@ import {
 	unblindSignature,
 	createBlindSignature,
 	constructProofFromPromise,
-	createRandomBlindMessage,
+	createRandomRawBlindedMessage,
 	serializeProof,
 	deserializeProof,
 	getKeysetIdInt,
@@ -27,7 +27,7 @@ describe('test crypto scheme', () => {
 		const mintPubKey = secp256k1.getPublicKey(mintPrivKey, true);
 
 		//Wallet(Bob)
-		const blindMessage = createRandomBlindMessage();
+		const blindMessage = createRandomRawBlindedMessage();
 
 		//Mint
 		const blindSignature = createBlindSignature(blindMessage.B_, mintPrivKey, 1, '');


src/mint/types/responses.ts
```diff
@@ -108,9 +108,9 @@ export type GetInfoResponse = {
 };
 
 /**
- * Response from the mint after requesting a melt quote.
+ * Base melt quote response - all melt quotes have these fields (NUT-05)
  */
-export type PartialMeltQuoteResponse = {
+export type NUT05MeltQuoteResponse = {
```
robwoodgate, line 113: Review note: This was an interim name - in #414 it is moved and renamed to `MeltQuoteBaseResponse`

src/mint/types/responses.ts
```diff
@@ -131,6 +131,23 @@ export type PartialMeltQuoteResponse = {
 	 * Timestamp of when the quote expires.
 	 */
 	expiry: number;
+	/**
+	 * Optional change from overpaid fees. If blanks were provided in `outputs`, the mint may return
+	 * signatures here.
+	 */
+	change?: SerializedBlindedSignature[];
```
robwoodgate, line 138: Reviewer note: Although not part of the NUT-05 base spec, I added this as a design decision as it is shared by bolt11 an bolt12, assumed optional by Wallet, and may later be useful for generic payment methods.

src/mint/types/responses.ts
```diff
@@ -144,17 +161,48 @@ export type PartialMeltQuoteResponse = {
...
+	fee_reserve: number;
+	/**
+	 * Preimage of the paid invoice. is null if it the invoice has not been paid yet. can be null,
+	 * depending on which LN-backend the mint uses.
+	 */
+	payment_preimage: string | null;
+	/**
+	 * Return/Change from overpaid fees. This happens due to Lighting fee estimation being inaccurate.
+	 */
+	change?: SerializedBlindedSignature[];
```
robwoodgate, line 187: Reviewer note: This superfluous key was removed in the tidy up in PR #414  (I moved it to the base type)

src/mint/Mint.ts
```diff
@@ -394,6 +395,56 @@ class Mint {
...
+			customRequest?: RequestFn;
+			preferAsync?: boolean;
+		},
+	): Promise<NUT05MeltQuoteResponse & TRes> {
+		// Set headers as needed
+		const headers: Record<string, string> = {
+			...(options?.preferAsync ? { Prefer: 'respond-async' } : {}),
+		};
+		// Sanitize method string and make request
+		if (!/^[a-z0-9-]+$/.test(method)) {
```
robwoodgate, line 421: Reviewer note: This was changed in #414 to include underscore, and abstracted to a helper ready for work on generics

src/mint/Mint.ts
```diff
@@ -394,6 +395,56 @@ class Mint {
...
+		const data = await this.requestWithAuth<NUT05MeltQuoteResponse & TRes>(
+			'POST',
+			`/v1/melt/${method}`,
+			{ requestBody: meltPayload, headers },
+			options?.customRequest,
+		);
+
+		// Runtime shape check for basic NUT05MeltQuoteResponse
+		// @todo - Tests need updating before we can do full shape check!
+		if (
```
robwoodgate, line 433: Question: Do we need more than an object existence check? More would only be needed if we don't trust the mints to return what the NUTs say they should return.

src/wallet/Wallet.ts
```diff
@@ -661,13 +671,15 @@ class Wallet {
...
 
 		// Sort ASC by amount for privacy, but keep indices to return order afterwards
+		// But ONLY if the transaction is NOT SIG_ALL (as order is fixed for signing)
 		const mergedBlindingData = [...keepOutputs, ...sendOutputs];
-		const indices = mergedBlindingData
-			.map((_, i) => i)
-			.sort(
+		const indices = mergedBlindingData.map((_, i) => i);
+		if (!isP2PKSigAll(inputs)) {
+			indices.sort(
```
robwoodgate, line 678: Reviewer note: This was a design decision - we have to fix the order or outputs for SIG_ALL, and that must happen before signing. So SIG_ALL consumers are responsible for doing their own output sorting before completing the swap.

src/wallet/Wallet.ts
```diff
@@ -1764,83 +1925,84 @@ class Wallet {
...
-		blanks: MeltBlanks<T>,
-	): Promise<MeltProofsResponse> {
-		const meltResponse =
-			blanks.method === 'bolt12'
-				? await this.mint.meltBolt12(blanks.payload)
-				: await this.mint.meltBolt11(blanks.payload);
-
-		// Check for too many signatures before mapping
+	async completeMelt<TQuote extends NUT05MeltQuoteResponse>(
+		meltPreview: MeltPreview<TQuote>,
```
robwoodgate, line 1960: Reviewer note: The switch from the MeltBlanks type to MeltPreview is technically a breaking change.

```ts
export interface MeltBlanks<T extends MeltQuoteResponse = MeltQuoteResponse> {
	method: 'bolt11' | 'bolt12';
	payload: {
  	    quote: string;
  	    inputs: Proof[];
  	    outputs: SerializedBlindedMessage[];
    };
	outputData: OutputDataLike[];
	keyset: Keyset;
	quote: T;
}
```

to 

```ts
/**
 * Preview of a Melt transaction created by prepareMelt.
 */
export interface MeltPreview<TQuote extends MeltQuoteBaseResponse = MeltQuoteBolt11Response> {
	method: string;
	inputs: Proof[];
	outputData: OutputDataLike[];
	keysetId: string;
	quote: TQuote;
}
```

However, the simplicity of the new shape, which reduces duplication of data and reduces nesting, makes it easier to work with and store if needed... as well as making it generic melt friendly.

robwoodgate: I had a cunning plan... in #414  I have deprecated the meltBlanksCreated / onChangeOutputsCreated events as the new prepareMelt() method makes it simple to get the blank outputs. I have also reinstated MeltBlanks for those deprecated callbacks, and added a simple converter in completeMelt to handle the legacy format.

src/crypto/core.ts
```diff
@@ -98,24 +96,34 @@ export function createBlindSignature(
...
-	);
+/**
+ * Creates a random blinded message.
+ *
+ * @remarks
+ * The secret is a UTF-8 encoded 64-character lowercase hex string, generated from 32 random bytes
+ * as recommended by NUT-00.
+ * @returns A BlindedMessage: {B_, r, secret}
+ */
+export function createRandomBlindMessage(): BlindMessage {
```
robwoodgate: Review note: I renamed this function and it's return type to a) harmonize it with createBlindSignature, and b) disambiguate it from the BlindedMessage class.

It's technically a breaking change, but since it was not used in the library outside of tests, and was part of a legacy output signing scheme, I figured it won't matter.

If it does, we can rename it back or add a deprecated pointer function for now.

src/crypto/core.ts
```diff
@@ -1,22 +1,20 @@
...
-import { getSignedOutput } from './NUT11';
 
 export type BlindSignature = {
 	C_: WeierstrassPoint<bigint>;
 	amount: number;
 	id: string;
 };
 
-export type BlindedMessage = {
+export type BlindMessage = {
```
Egge21M: This diverges from the naming conventions established by https://github.com/cashubtc/nuts/blob/main/00.md

src/crypto/core.ts
Egge21M: I see. Can we not adjust this utility to return a BlindedMessage instance instead of the plain object? IMHO everything that is not actively used should be marked as deprecated as soon as possible, so that we do not forget about deleting it later.

src/crypto/NUT11.ts
```diff
@@ -18,6 +18,21 @@ export type SecretData = {
...
 };
 
+/**
+ * Validates proof secret is P2PK.
+ *
+ * @param secretStr - The Proof secret.
+ * @throws If secret is not P2PK.
+ * @internal
+ */
+const validateP2PKSecret = (secretStr: string | Secret): Secret => {
```
Egge21M, line 28: LGTM, but I wonder whether we should rollup `parse` and `validate`, because I personally would expect `parseP2PKSecret` to throw already if the secret is not a P2PK one.

src/mint/types/responses.ts
Egge21M: I assume the same is true for the Payload?

src/wallet/Wallet.ts
```diff
@@ -1045,11 +1173,41 @@ class Wallet {
...
+		// SIG_ALL is in flux currently, so let's generate all known message formats
+		// and sign the first proof only against each message...
+		const [first, ...rest] = proofs;
+		let signedFirst = first;
+		const messages = [
+			buildLegacyP2PKSigAllMessage(proofs, outputData, quoteId),
+			buildInterimP2PKSigAllMessage(proofs, outputData, quoteId),
+			buildP2PKSigAllMessage(proofs, outputData, quoteId),
+		];
+		messages.map(
```
Egge21M: nit: `map` looks like this is intended to mutate the message array, which however is disregarded after this operation. `forEach` would be more readable

src/wallet/Wallet.ts
```diff
@@ -1764,83 +1925,84 @@ class Wallet {
...
 		const change =
-			meltResponse.change?.map((s, i) => blanks.outputData[i].toProof(s, blanks.keyset)) ?? [];
+			meltResponse.change?.map((s, i) => meltPreview.outputData[i].toProof(s, keyset)) ?? [];
 
-		this._logger.debug('COMPLETE MELT', { changeAmounts: change.map((p) => p.amount) });
-		return {
-			quote: { ...meltResponse, unit: blanks.quote.unit, request: blanks.quote.request },
-			change,
-		};
+		this._logger.debug('MELT COMPLETED', { changeAmounts: change.map((p) => p.amount) });
```
Egge21M, line 1998: Melt might acutally not be completed at this point is preferAsync is used

src/wallet/WalletOps.ts
```diff
@@ -236,7 +267,21 @@ export class SendBuilder {
 	}
 
 	/**
-	 * Execute the send or swap.
+	 * Prepare the swap to send.
+	 *
+	 * @returns A SwapPreview containing inputs, outputs, amount, fee and unselectedProofs.
+	 */
+	async prepare() {
```
Egge21M, line 274: This breaks the builder pattern by not returning the class instance. What is the appropriate use of this method?

Egge21M (commented):
Generally this L(very)GTM. I think splitting up the operations is a great addition. I have left a few comments.

src/mint/types/responses.ts
robwoodgate: MeltQuoteBaseResponse *is* the payload. MeltQuoteBaseRequest is the request - so yes, both were named using same convention.

src/crypto/core.ts
robwoodgate: I agree that we should start deprecating as soon as we stop using something. This function is still used internally for tests (for NUT-12 DLEQ creation) so not entirely ready to hang up its boots yet. 

But since we have the OutputData class for output creation, and it previously had a private key param for legacy output signing, I'd be surprised if anyone was using the raw crypto function in production.

It would not make sense to return a BlindedMessage instance in the crypto library. The BlindedMessage class is a higher level wrapper for the lower level blinded message (B_).

The best illustration is in OutputData:createSingleRandomData()

```ts
static createSingleRandomData(amount: number, keysetId: string) {
		const randomHex = bytesToHex(randomBytes(32));
		const secretBytes = new TextEncoder().encode(randomHex);
		const { r, B_ } = blindMessage(secretBytes);
		return new OutputData(
			new BlindedMessage(amount, B_, keysetId).getSerializedBlindedMessage(),
			r,
			secretBytes,
		);
	}
```

In theory, `createRandomBlindMessage` could be the first part of `createSingleRandomData`:

```ts
static createSingleRandomData(amount: number, keysetId: string) {
		const { r, B_, secret } = createRandomBlindMessage();
		return new OutputData(
			new BlindedMessage(amount, B_, keysetId).getSerializedBlindedMessage(),
			r,
			secret,
		);
	}
```

But it didn't seem worth importing as it's the only usable call site, and it's clearer to have the detailed steps in the method,, like the other methods in the class do.

All that to say, we COULD mark it deprecated, and then later make it `@internal` to remove it from public API.

src/crypto/core.ts
robwoodgate: It's a grey area - a NUT-00 BlindedMessage includes an amount:

```ta
{
  "amount": int,
  "id": hex_str,
  "B_": hex_str
}
```

The crypto level `BlindMessage` doesn't have an amount, and includes the secret. It's a lower level crypto building block to a full `BlindedMessage`. So that's why I renamed it.

robwoodgate: I chose `BlindMessage` because it is primarily the output of the `blindMessage` fumction.

But maybe `RawBlindedMessage` would be better?

src/crypto/NUT11.ts
robwoodgate: The only reason I didn't is because we had `parseP2PKSecret` as part of the public API already, and it would mean altering its signature and handling a messy string | Uint8Array | Secret union.

So I made `validateP2PKSecret` an internal (no export) helper simply to reduce the boilerplate checks in all the other functions and left parse alone.

src/wallet/Wallet.ts
robwoodgate: Totally fair point - I changed it in [c1ae7c4](https://github.com/cashubtc/cashu-ts/pull/413/commits/c1ae7c47ee2aa9e14718c5db5f6d68e57baebb42)

src/wallet/Wallet.ts
robwoodgate: Good spot - addressed in #414 (for easier merge)

src/wallet/WalletOps.ts
robwoodgate: This allows the consumer to run the `prepare` version of `send` and receive a swap preview only, for later completion with `completeSwap()`. It doesn't return the class instance because it's a final command (like `run()`).

robwoodgate: Added example to README in #414, commit [b80d469](https://github.com/cashubtc/cashu-ts/pull/414/commits/b80d4691bfe7406810277e078d8ebd4e483cf300)

src/crypto/NUT11.ts
robwoodgate: We will have to revisit to accommodate the upcoming changes in NUT-11 / 14, so lets reconsider it then.

src/crypto/core.ts
robwoodgate: I reinstated the original `BlindedMessage` type and `createRandomBlindedMessage` function and marked them as deprecated. So no breaking change any more.

src/crypto/core.ts
robwoodgate: I renamed to `RawBlindedMessage `

Issue #419, opened 2025-11-24
Integer checking


robwoodgate:
Some wallet methods take an `amount` as input, which is generally expected to be a positive integer (we don't have fractional or negative key amounts in Cashu).

We should assert amount inputs are what we expect, failing loudly if a consumer passes a negative or float value.

Enforcing a firm boundary makes it easier for a consumer to chase down errors in their wallet too.

PR #420, opened 2025-11-24
V3 assert integer

robwoodgate:
# Fixes: #419 

## Description

Adds private `assertAmount` method to Wallet, and asserts on the send/receive/selectProofs/mint etc methods.

Also adds and tidies up tests.

## Changes

- Added `assertAmount` private method to wallet
- Adds assertion to relevant Wallet methods
- Adds `MINTINFORESP` and `MINTINFOCACHE` constants to test/consts
- Adds tests for `assertAmount`
- Tidies up tests that used a mintInfoCache to use the MINTINFOCACHE const

## NOTE

Built on top of the `v3-mint-ment-types` branch. Should be merged AFTER #414 

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -259,6 +259,8 @@ class Wallet {
 	/**
 	 * Asserts amount is a positive integer.
 	 *
+	 * @param amount To check.
+	 * @param op Caller method name (or other identifier) for debug.
 	 * @throws If not.
 	 */
 	private assertAmount(amount: unknown, op: string): asserts amount is number {


src/wallet/Wallet.ts
```diff
@@ -1599,6 +1611,7 @@ class Wallet {
 		config?: MintProofsConfig,
 		outputType?: OutputType,
 	): Promise<Proof[]> {
+		this.assertAmount(amount, `_mintProofs: ${method}`);
```
TheRealCheebs, line 1621: General question based on this function having the this.failIf(amount <= 0, 'Invalid mint amount...) a few lines later. Is that needed, the assertAmount allows for 0, while this function doesn't. But the question is should assertAmount allow 0?

robwoodgate: Great question - my original reasoning was it makes it a more flexible assert - as zero is not an "out of bounds fail" but more like "operationally pointless" kind of fail (caught upstream if needed).

But looking again at all the call sites, there isn't a single case where zero would make sense.
So it would strengthen them all to throw on zero, and remove the need for the second narrowing in _mintProofs.

Issue #425, opened 2025-12-09
Wallet init from cached data in sync-only environments


robwoodgate:
Wallet does a single async[ `loadMint()`](https://github.com/cashubtc/cashu-ts/blob/development/src/wallet/Wallet.ts#L280) to populate the mint caches and init the wallet, speeding up operations and preventing race conditions etc.

In some cases, a Wallet might need to be called in a sync-only context, such as a Vue component (example: Cashu.me). This is possible using cached Mint data.

The problem is that `loadMint` runs both async Mint fetches AND the final init, which makes it unreliable to use in sync constrained contexts.

The final init part of `loadMint` (below) needs to be separated into its own (private?) non-async `init()` method 

```ts
this._logger.debug('KeyChain', { keychain: this.keyChain.getCache() });

if (this._boundKeysetId === PENDING_KEYSET_ID) {
	this._boundKeysetId = this.keyChain.getCheapestKeyset().id;
} else {
	// Ensure the bound id is still present and keyed
	const k = this.keyChain.getKeyset(this._boundKeysetId);
	this.failIf(!k.hasKeys, 'Wallet keyset has no keys after refresh', { keyset: k.id });
}
```

This can then be called as the final part of `loadMint`, keeping the existing contract for async consumers.

It can then AlSO be run in the constructor IF options `keys`, `keysets`, and `mintInfo` were passed in.

That way, Wallet init with cached data is ready on construct, and Wallet init from scratch is ready on loacMint.

PR #426, opened 2025-12-09
Refactor wallet init with cached data

robwoodgate:
# Fixes: #425 

## Description

The wallet class and keychain classes have been refactored to allow full init with cached data in synchronous contexts.

Instantiation can now be completed async as normal using `loadMint` or synchronously with cached data using `loadMintFromCache`.

Example (from updated README):

```ts
// Simplest: With a mint URL
const mintUrl = 'http://localhost:3338';
const wallet1 = new Wallet(mintUrl); // unit is 'sat'
await wallet1.loadMint(); // wallet is now ready to use

const cache = wallet1.keyChain.getCache(); // persist keychain data in your app
const mintInfo = wallet1.getMintInfo().cache; // persist mint info in your app

// Advanced: With cached mint data (reduces API calls)
const wallet2 = new Wallet(cache.mintUrl, { unit: cache.unit });
wallet2.loadMintFromCache(mintInfo, cache.keysets, cache.keys);
// wallet2 is now ready to use
```

## Changes

- Deprecates the wallet constructor options for cached data (mintinfo, keys, keysets)
- Adds `loadMintFromCache` method for wallet init from caches
- Adds `cache` getter to MintInfo
- Adds `loadFromCache` method to KeyChain to allow synchronous init
- Wallet keychain now stored in `_keyChain` to allow internal reset with public readonly getter
- Adds private `finishInit` method to keep post init tasks DRY
- Updates tests to use new `loadMintFromCache` construction

## Incidental Changes (code quality)

- Boosted test coverage for MintInfo class

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -238,7 +238,7 @@ class Wallet {
 		this._keyChain = new KeyChain(this.mint, this._unit);
 		this._denominationTarget = options?.denominationTarget ?? this._denominationTarget;
 
-		// Deprecated: cache init - to be removed later
+		// @todo: Deprecated cache init - to be removed with deprecated options
 		if (options?.keysets && options?.keys && options?.mintInfo) {
 			this.loadMintFromCache(options.mintInfo, options.keysets, options.keys);
 		}
@@ -2163,7 +2163,7 @@ class Wallet {
 		const Ys = proofs.map((p: Pick<Proof, 'secret'>) =>
 			hashToCurve(enc.encode(p.secret)).toHex(true),
 		);
-		// TODO: Replace this with a value from the info endpoint of the mint eventually
+		// @todo: Replace this with a value from the info endpoint of the mint eventually
 		const BATCH_SIZE = 100;
 		const states: ProofState[] = [];
 		for (let i = 0; i < Ys.length; i += BATCH_SIZE) {


src/wallet/Wallet.ts
```diff
@@ -225,9 +234,13 @@ class Wallet {
 			this._counterSource = new EphemeralCounterSource(options?.counterInit);
 		}
 		this.counters = new WalletCounters(this._counterSource);
-		this.keyChain = new KeyChain(this.mint, this._unit, options?.keysets, options?.keys);
-		this._mintInfo = options?.mintInfo ? new MintInfo(options.mintInfo) : this._mintInfo;
+		this._keyChain = new KeyChain(this.mint, this._unit);
 		this._denominationTarget = options?.denominationTarget ?? this._denominationTarget;
+
+		// Deprecated: cache init - to be removed later
```
TheRealCheebs: just a clerical note / question, are there tags in the code to remind folks to remove this later?

robwoodgate: Great question @TheRealCheebs .

Yes, the  `@deprecated blah blah` tags are there as a signal to the developer via their IDE... it will probably underline or flag calls to deprecated options so they know they need to update. The blah blah tells them what to do instead.

The comment you flagged above is purely internal for us to remember that this call is part of the deprecated init flow.

TheRealCheebs: Yes, I was referring to the internal comment for us. Because we will want to remove that in ~6ish months (or some amount of time for folks to get updated) I was wondering if the project had a way to not forget we want that removed at some point. Like a quick grep of the codebase on <insert tag here> to pull up the list of things to not forget about.

dumb question.

robwoodgate: Not a dumb question at all... generally, I use `@todo` in code blocks that future me need to remember to look at.

I didn't do it here, but it would be good if we were consistent on such stuff.

robwoodgate: Updated with `@todo`

PR #427, opened 2025-12-09
Domain separator as ASCII text

robwoodgate:
## Description

For auditability and peace of mind, we should store our DST in plain ASCII text, and not a hex string.

That way, we can see at a glance it hasn't been tampered with.

## Changes

- DST changed from hextobytes(hex) to `utf8ToBytes('Secp256k1_HashToCurve_Cashu_')`

## PR Tasks

- [x] Open PR
- [x] run `npm run format`


TheRealCheebs:
Good thinking!

Issue #264, opened 2025-03-10
Clean up Key types


Egge21M:
We have quite a lot of types dealing with a mint's keysets. However the type's names are ambiguous. We should clean these up or move them into their own class.

Example: Here we have the `MintKeys` type describing a key set with all its key-pairs. Then we have `MintKeyset` which represents the a single keyset entry without the keys. `MintActiveKeys` is an Array of `MintKeys`, but `MintAllKeysets` is an Array of `MintKeysets`. 

Proposed solution: Merge all these types into a single Keyset Class or type.

```ts
export type Keys = { [amount: number]: string };

export type MintActiveKeys = {
	keysets: Array<MintKeys>;
};

export type MintAllKeysets = {
	keysets: Array<MintKeyset>;
};

export type MintKeys = {
	id: string;
	unit: string;
	keys: Keys;
};

export type MintKeyset = {
	id: string;
	unit: string;
	active: boolean;
	input_fee_ppk?: number;
};
```

robwoodgate:
#355 makes a start on this - MintActiveKeys and MintAllKeysets are not used in the new Wallet class, and MintKeys / MintKeyset are only used in the constructor for loading cached api data. It also implements a KeyChain and Keyset class to abstract key management from the wallet. Keychain has a compatible shape to MintKeys and MintKeyset so utilites like `getDecodedToken` can use both objects.

PR #429, opened 2025-12-10
Refactor wallet init with cached data + Clean up key types

robwoodgate:
# Fixes: #264, #425, Supercedes: #426

### Refactor wallet init with cached data (from #426)

This PR supercedes #426 by rolling up, and building on the changes in that PR (included suggested fixes).

It redefines keychain caching and the `loadMintFromCache` method to use a single, unified `KeyChainCache` object.

EXAMPLE: new wallet instantiation from README:

```typescript
import { Wallet } from '@cashu/cashu-ts';

// Simplest: With a mint URL
const mintUrl = 'http://localhost:3338';
const wallet1 = new Wallet(mintUrl); // unit is 'sat'
await wallet1.loadMint(); // wallet is now ready to use

// Persist these in your app
const keychainCache = wallet1.keyChain.cache; // KeyChainCache
const mintInfoCache = wallet1.getMintInfo().cache; // GetInfoResponse

// Advanced: With cached mint data (avoids network calls on startup)
const wallet2 = new Wallet(keychainCache.mintUrl, { unit: keychainCache.unit });
wallet2.loadMintFromCache(mintInfoCache, keychainCache);
// wallet2 is now ready to use
```

### Clean up Mint Key types

It also cleans up Mint key types to bring them closer to the NUTs, deprecates some remaining v2 shapes and removes the risk of future circular dependencies by decoupling the core utils from the wallet module.

We now have these more easily understandable Mint API types:

```ts
// Raw Mint API DTOs
GetKeysResponse - now reflects NUT01 naming  (was MintActiveKeys)
GetKeysetsResponse - now reflects NUT02 naming (was MintAllKeysets)

// Mint DTOs (unchanged)
MintKeyset - the metadata for a Mint keyset (from GetKeysetsResponse)
MintKeys - the keys for a Mint keyset (from GetKeysResponse)
```

## Changes

- Mint API DTO types renamed to `GetKeysResponse`, `GetKeysetsResponse` as per NUT01/02
- Adds `KeysetCache` and `KeyChainCache` types for caching mint data
- Adds `cache` getter to MintInfo
- Makes OutputData more generic to decouple it from MintKeys DTO and Keychain
- Introduces a `KeyLike` minimal required base type which callers can expand on.
- Makes core utils generic (`getDecodedToken`, `mapShortKeysetIds`, `hasValidDleq`) to decouple from wallet layer
- Deprecates `verifyKeysetId` in core utils - now a static helper in Keyset class along with `fromMintApi `
- Adds static helpers to KeyChain - `fromCache`, `mintToCacheDTO`, `cacheToMintDTO`
- Adds `cache` getter, `loadFromCache` and `getAllKeys` methods to KeyChain
- Deprecates the v2 keys for Keyset class, adds `fromMintApi` static helper
- Wallet now exposes KeyChain as a private param + getter, instead of public readonly
- Wallet deprecates the cache options in the constructor, adds `loadMintFromCache` and `finishInit`
- Completes test coverage for KeyChain Class, adds tests for caches

## Incidental Changes (code quality)

- Fixes api-extractor output so it stops flip-flopping on newline and indentation. Now enforces LF and 4 space indentation to hopefully reduce the noise and weirdness on API updates.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/OutputData.ts
+++ b/src/model/OutputData.ts
@@ -26,8 +26,11 @@ import { bytesToNumber, numberToHexPadded64, splitAmount } from '../utils';
 
 /**
  * Note: OutputData helpers only require keyset `id` and `keys`. If you want richer keyset typing at
- * the call site, use `OutputDataLike<YourType>`. In v4 we may simplify this further by fixing the
- * keyset type to `HasKeysetKeys` and removing the generic.
+ * the call site, use `OutputDataLike<YourType>`.
+ *
+ * @remarks
+ * WARNING: In v4 we may simplify this further by fixing the keyset type to `HasKeysetKeys` and
+ * removing the generic.
  */
 export interface OutputDataLike<TKeyset extends HasKeysetKeys = HasKeysetKeys> {
 	blindedMessage: SerializedBlindedMessage;
@@ -39,8 +42,11 @@ export interface OutputDataLike<TKeyset extends HasKeysetKeys = HasKeysetKeys> {
 
 /**
  * Note: OutputData helpers only require keyset `id` and `keys`. If you want richer keyset typing at
- * the call site, use `OutputDataLike<YourType>`. In v4 we may simplify this further by fixing the
- * keyset type to `HasKeysetKeys` and removing the generic.
+ * the call site, use `OutputDataLike<YourType>`.
+ *
+ * @remarks
+ * WARNING: In v4 we may simplify this further by fixing the keyset type to `HasKeysetKeys` and
+ * removing the generic.
  */
 export type OutputDataFactory<TKeyset extends HasKeysetKeys = HasKeysetKeys> = (
 	amount: number,
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -394,7 +394,7 @@ function tokenFromTemplate(template: TokenV4Template): Token {
  */
 export function getDecodedToken(tokenString: string, keysetIds?: readonly string[]): Token;
 /**
- * @deprecated Pass keyset ids as `string[]` instead.
+ * @deprecated Pass keyset ids as `string[]` instead, eg: using KeyChain.getAllKeysetIds()
  */
 export function getDecodedToken(tokenString: string, keysetIds?: readonly HasKeysetId[]): Token;
 export function getDecodedToken(
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -1371,11 +1371,8 @@ class Wallet {
 	 * @returns Token object.
 	 */
 	public decodeToken(token: string): Token {
-		const keysets = this._keyChain.getKeysets();
-		return getDecodedToken(
-			token,
-			keysets.map((k) => k.id),
-		);
+		const keysetIds = this._keyChain.getAllKeysetIds();
+		return getDecodedToken(token, keysetIds);
 	}
 
 	// -----------------------------------------------------------------


robwoodgate:
@zk-bits, @TheRealCheebs - if you have time to cast an eye over this one, it would be most appreciated. Thanks!

src/model/OutputData.ts
```diff
@@ -20,15 +21,26 @@ import { BlindedMessage } from './BlindedMessage';
...
-export interface OutputDataLike {
+/**
+ * Minimal key carrier shape for OutputData helpers.
+ *
+ * Any type with `id`, and `keys` can be used as TKeyset in OutputDataLike and OutputDataFactory,
+ * including MintKeys and wallet Keyset.
+ */
+export type KeyLike = { id: string; keys: Keys };
+
+export interface OutputDataLike<TKeyset extends KeyLike = MintKeys | Keyset> {
```
Egge21M: Why make this type generic? Shouldn't fixing the type to KeyLike be enough?

src/utils/core.ts
```diff
@@ -388,21 +386,36 @@ function tokenFromTemplate(template: TokenV4Template): Token {
...
+ *
+ * // Strongly typed: (keysets param is a Keyset)
+ * const token = getDecodedToken<Keyset>(tokenString, keysets);
+ * ```
+ *
+ * @param token An encoded cashu token (cashuB...)
  * @returns Cashu token object.
  */
-export function getDecodedToken(tokenString: string, keysets?: MintKeyset[] | Keyset[]) {
+export function getDecodedToken<T extends { id: string }>(
```
Egge21M: nit: I think using a type parameter here disguises the real purpose of this. If the goal is to be less restrictive on the params we could do something like

`keysets?: {id: string}` or more explicit `keyset?: Pick<Keyset, "id">`

src/wallet/Wallet.ts
```diff
@@ -226,9 +235,21 @@ class Wallet {
 			this._counterSource = new EphemeralCounterSource(options?.counterInit);
 		}
 		this.counters = new WalletCounters(this._counterSource);
-		this.keyChain = new KeyChain(this.mint, this._unit, options?.keysets, options?.keys);
-		this._mintInfo = options?.mintInfo ? new MintInfo(options.mintInfo) : this._mintInfo;
+		this._keyChain = new KeyChain(this.mint, this._unit);
 		this._denominationTarget = options?.denominationTarget ?? this._denominationTarget;
+
+		// @todo: Deprecated cache init - to be removed with deprecated constructor options
```
Egge21M: convention is to use `// TODO:` so that common plugins like [better-comments](https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments) will pick it up

src/model/OutputData.ts
```diff
@@ -20,15 +21,26 @@ import { BlindedMessage } from './BlindedMessage';
...
 import { bytesToNumber, numberToHexPadded64, splitAmount } from '../utils';
 
-export interface OutputDataLike {
+/**
+ * Minimal key carrier shape for OutputData helpers.
+ *
+ * Any type with `id`, and `keys` can be used as TKeyset in OutputDataLike and OutputDataFactory,
+ * including MintKeys and wallet Keyset.
+ */
+export type KeyLike = { id: string; keys: Keys };
```
Egge21M: nit: I think this export would probably fit better into the keyset related files

.editorconfig
```diff
@@ -12,3 +12,8 @@ trim_trailing_whitespace = true
 [*.{yml,yaml}]
 indent_style = space
 indent_size = 2
+
+# Override for API Extractor
+[etc/*.api.md]
```
Egge21M, line 17: Looks like the whole api extractor file got overwritten in this PR. Maybe thats related?

etc/cashu-ts.api.md
Egge21M, line 1: I noticed your comment regarding the formatting changes to the extractor process. Is this complete overwrite a sideeffect of this and expected?

src/utils/core.ts
KvngMikey: yes, If the real purpose is input flexibility, a union or structural type may be clearer.

src/model/OutputData.ts
robwoodgate: I left it here for now as the only consumer of it is OutputData. It's only exported because it appears in public signatures and would otherwise be a forgotten export.

But if we use it in other spots, we definitely should move it to another place, such as `model/types/keyset.ts`

.editorconfig
robwoodgate: Yes, I got fed up with it flip-flopping because of newline styles and tabs/spaces. So I standardized it on LF (so we don't see horrible Windows `^M` characters) and 4 spaces, which is what it wants to do when not fighting with our project preference for tabs. Should be the last time it gets a total rewrite (hopefully)

etc/cashu-ts.api.md
robwoodgate: Yes, and yes :)

src/wallet/Wallet.ts
robwoodgate: Ok cool, I'll change it.

src/utils/core.ts
robwoodgate: These are lower level utilities which really should not have knowledge or coupling to higher level domain types.

The function doesn't actually need "a keyset", it needs “something I can read a keyset id from”.

By removing the coupling, especially to Keyset, we avoid type level import dependencies (utils are lower level constructs and should not really be importing things from the modules they are being used in).

We could use `Pick<Keyset, 'id'>`, but it effectively collapses to `{ readonly id: string }`, and couples a domain model for very little actual gain (and keyset is the param name).

All that said, I agree the generic type parameter is not buying us much here since `T` is not used anywhere except the param.

I’d prefer something like  `HasId` or even reuse the `KeyLike` type over `Pick<Keyset, "id">`,  because Pick still leaks a domain model into a low level util surface, and it can show up in generated types.

So maybe something like:

```ts
type HasId = { id: string };
export function getDecodedToken(
    tokenString: string,
    keysets?: readonly HasId[],
): Token {
    // remove prefixes
    const tokenStr = removePrefix(tokenString);
    const token: Token = handleTokens(tokenStr);
    token.proofs = mapShortKeysetIds(token.proofs, keysets);
    return token;
}
```

What do you think?

src/model/OutputData.ts
robwoodgate: I was trying to decouple without breaking the existing API, since callers may already have factories or implementations typed to MintKeys or Keyset, or both.

I think that if we fix toProof to accept KeyLike, then any implementation that only accepts Keyset (or only MintKeys) stops being assignable under strictFunctionTypes, because the interface would be promising it can handle any KeyLike.

Whereas the generic keeps the keyset type consistent end to end, factory input and toProof agree on the same type, without leaking the higher level domain types into the helper surface.

Unless we are happy to break this (any consumers using factories are likely to be advanced users), we can look to drop the generic entirely, making KeyLike the fixed type as a deliberate breaking change in v4?

src/utils/core.ts
robwoodgate: Or your exact suggestion works for me too - `keysets?: {id: string}`

One less type in the graph. LMK which you prefer.

src/wallet/KeyChain.ts
```diff
@@ -51,6 +144,27 @@ export class KeyChain {
...
+	 * Synchronously load keysets and keys from cached data.
+	 *
+	 * @remarks
+	 * Does not hit the network. Intended for callers that already have a KeyChainCache and want a
+	 * synchronous path.
+	 */
+	loadFromCache(cache: KeyChainCache): void {
+		if (cache.unit !== this.unit) {
+			throw new Error(
+				`KeyChain unit mismatch in cache, expected '${this.unit}', got '${cache.unit}'`,
```
zk-bits: When the cache unit doesn't match, the error message could be more helpful by including the mintUrl from the cache to help developers debug where the mismatched cache came from. Its a nit, but could really prevent a headache for devs. something small along the lines of:

throw new Error(
    `KeyChain unit mismatch in cache, expected '${this.unit}', got '${cache.unit}' (from ${cache.mintUrl})`
);

src/utils/core.ts
Egge21M: I see your point, however ID by itself is VERY ambiguous. getDecodedToken needs a list of Keyset ids not just any identifiers. Pick<Keyset, "id> might collapse to the same as the literal, but it carries information. We could assume that consumers can figure this link out by themselves from the parameter name, but it's more abstract.

If you veto Pick<> I am perfectly fine with {id: string} for now. We should leave a comment and break to `keysetIds: string[]` (which is the best solution IMO) in v4

robwoodgate: If it's ok with you, I would like to veto  `Pick<Keyset, "id">`, as it is not much of an improvement to coupling over the union we had before.

I think that a good compromise is a descriptive type to represent the fact that these functions can accept way more types than originally specified, including: `MintKeyset`, `MintKeys`, `KeyLike`, `Keyset`, `KeysetCache`, even `Proof` (probably never going to be used to decode tokens, but you get my point!)

Maybe something like `HasKeysetId`?

```ts
/**
* Any object with a keyset `id` - including: `MintKeyset`, `MintKeys`, `KeyLike`, `Keyset`, `KeysetCache`
*/
type HasKeysetId = { id: string };
```

If you can think of a better name, am happy with anything you like that resolves the ambiguity.

We can then decide whether to collapse that down from an object with id to plain array of strings as a breaking change to v4. 

I suspect, we might like the flexibility of being able to pass any object with a keyset id, but if we decide to collapse to an array of strings, it would be  simple to add a helper to the KeyChain class like `getAllKeysetIds`.

robwoodgate: Or even `Pick< KeyLike , "id">`, then `KeyLike` can be used in all three functions (including hasValidDleq)

src/wallet/KeyChain.ts
robwoodgate: Added

src/model/OutputData.ts
robwoodgate: Moved to `model/types/keyset.ts` and renamed to `HasKeysetKeys`.

src/utils/core.ts
robwoodgate: I LOVE your `keysetIds: string[] ` suggestion for v4, and i totally agree this is the ultimate destination.

So I've prepped us for a smooth change in v4 (or even later in v3) by deprecating the object-like keyset param in the signature and adding a new `keysetIds?: readonly string[]` overload signature. The method now maps on the fly to accomodate both param types. Same for `mapShortKeysetIds`.

I also added a `KeyChain.getAllKeysetIds()` method to make it simple to get an array of keyset IDs from the keychain going forward.

I think this gets us where we want to be with the utils, removing the coupling in a backwards compatible way, and nudging consumers towards `string[]`, without intrducing ambiguity.

src/model/OutputData.ts
robwoodgate: I've tweaked to default to `HasKeysetKeys`, which decouples and nudges consumers the right way. It is a minor type-level (not runtime) breaking change for any early-adopter of v3 who was using a custom factory, but a simple "fix", which is noted in the TSDoc and the migration guide.

I seriously doubt anyone with such advanced features as custom factories made the leap to v3 yet, and if they have, it will be a simple tweak to fix.

I've also added a //TODO: to remind us to decide if we want to simplify further in v4.

d4rp4t (approved):

PR #430, opened 2025-12-11
Upgrade @noble/curves and @noble/hashes to v2.x

happylemonprogramming:
# Fixes: Dependency update with @noble/* v2.x packages

## Description

Upgrades `@noble/curves` from `^1.3.0` to `^2.0.1` and `@noble/hashes` from `^1.3.3` to `^2.0.1` to resolve Shakespeare.diy build errors.

This resolves an issue where `esm.sh` cannot handle version conflicts that npm allows through nested installations. Packages like `coco-cashu-core` directly depend on `@noble/*` v2.x, while `@cashu/cashu-ts` previously required v1.x, creating a conflict for `esm.sh`.

## Changes

- **Dependencies**: Upgraded `@noble/curves` to `^2.0.1` and `@noble/hashes` to `^2.0.1`
- **Import paths**: Updated all imports to include `.js` extensions (required by v2 ESM-only API)
  - `@noble/curves/abstract/utils.js` → `@noble/curves/utils.js`
  - `@noble/hashes/sha256.js` → `@noble/hashes/sha2.js`
- **API updates**:
  - `ProjPointType` → `WeierstrassPoint`
  - `ProjectivePoint` → `Point`
  - `randomPrivateKey()` → `randomSecretKey()`
  - `fromPrivateKey()` → `Point.BASE.multiply()` or `getPublicKey()` + `fromHex()`
  - `CURVE.n` → `Point.CURVE().n`
  - `PrivKey` type → `Uint8Array`
- **Type conversions**: Fixed string-to-Uint8Array conversions for `sha256()` and `schnorr.verify()` calls
- **Test files**: Updated all test imports to use correct v2.x paths

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/crypto/mint/NUT12.ts
+++ b/src/crypto/mint/NUT12.ts
@@ -1,8 +1,7 @@
 import { hash_e, createRandomPrivateKey, type DLEQ } from '../common/index';
 import { type WeierstrassPoint } from '@noble/curves/abstract/weierstrass.js';
-import { bytesToHex, numberToBytesBE } from '@noble/curves/utils.js';
+import { numberToBytesBE } from '@noble/curves/utils.js';
 import { secp256k1 } from '@noble/curves/secp256k1.js';
-import { bytesToNumber, hexToNumber } from '../util/utils';
 
 /**
  * !!! WARNING !!! Not recommended for production use, due to non-constant time operations See:
@@ -10,16 +9,16 @@ import { bytesToNumber, hexToNumber } from '../util/utils';
  * https://en.wikipedia.org/wiki/Timing_attack for information about timing attacks.
  */
 export const createDLEQProof = (B_: WeierstrassPoint<bigint>, a: Uint8Array): DLEQ => {
-	const r = bytesToHex(createRandomPrivateKey()); // r <- random
-	const n_r = hexToNumber(r);
-	const R_1 = secp256k1.Point.BASE.multiply(n_r); // R1 = rG
-	const R_2 = B_.multiply(n_r); // R2 = rB_
-	const C_ = B_.multiply(bytesToNumber(a)); // C_ = aB_
-	const n_a = bytesToNumber(a);
-	const A = secp256k1.Point.BASE.multiply(n_a); // A = aG
+	const r = secp256k1.Point.Fn.fromBytes(createRandomPrivateKey()); // r <- random (Uint8Array)
+	const R_1 = secp256k1.Point.BASE.multiply(r); // R1 = rG
+	const R_2 = B_.multiply(r); // R2 = rB_
+	const scalar_a = secp256k1.Point.Fn.fromBytes(a);
+	const C_ = B_.multiply(scalar_a); // C_ = aB_
+	const A = secp256k1.Point.BASE.multiply(scalar_a); // A = aG
 	const e = hash_e([R_1, R_2, A, C_]); // e = hash(R1, R2, A, C_)
-	const n_e = bytesToNumber(e);
-	// WARNING: NON-CONSTANT TIME OPERATIONS?
-	const s = numberToBytesBE((n_r + n_e * n_a) % secp256k1.Point.CURVE().n, 32); // (r + ea) mod n
+	const scalar_e = secp256k1.Point.Fn.fromBytes(e);
+	// Use field operations for constant-time addition and multiplication
+	const s_scalar = secp256k1.Point.Fn.add(r, secp256k1.Point.Fn.mul(scalar_e, scalar_a));
+	const s = numberToBytesBE(s_scalar, 32); // s = (r + e * a) mod n
 	return { s, e };
 };


src/crypto/client/NUT12.ts
```diff
@@ -14,11 +13,11 @@ function arraysEqual(arr1: Uint8Array, arr2: Uint8Array) {
...
 	dleq: DLEQ,
-	B_: ProjPointType<bigint>,
-	C_: ProjPointType<bigint>,
-	A: ProjPointType<bigint>,
+	B_: WeierstrassPoint<bigint>,
+	C_: WeierstrassPoint<bigint>,
+	A: WeierstrassPoint<bigint>,
 ) => {
-	const sG = secp256k1.ProjectivePoint.fromPrivateKey(bytesToHex(dleq.s));
+	const sG = secp256k1.Point.BASE.multiply(bytesToNumber(dleq.s));
```
robwoodgate: If I read [the documentation](https://github.com/paulmillr/noble-curves) correctly, I think this should probably be:

```suggestion
	const sG = secp256k1.Point.BASE.multiply(secp256k1.Point.Fn.fromBytes(dleq.s));
```

Please can you double check and see if you agree.

robwoodgate: On reflection, I think you are ok here, as the point multiplication reduces modulo `n`.

But maybe it would be more aligned with docs to use `Point.Fn.fromBytes` for `s` and `e`?

```ts
const s = secp256k1.Point.Fn.fromBytes(dleq.s);
const e = secp256k1.Point.Fn.fromBytes(dleq.e);
const eA = A.multiply(e);
const sB_ = B_.multiply(s);
... etc
```

zk-bits: I have made this into an issue  #433. also have a specific PR to attach to it of we decide to fix it across the code base there are several files that could benefit from this clean up

happylemonprogramming: > On reflection, I think you are ok here, as the point multiplication reduces modulo `n`.
> 
> But maybe it would be more aligned with docs to use `Point.Fn.fromBytes` for `s` and `e`?
> 
> ```ts
> const s = secp256k1.Point.Fn.fromBytes(dleq.s);
> const e = secp256k1.Point.Fn.fromBytes(dleq.e);
> const eA = A.multiply(e);
> const sB_ = B_.multiply(s);
> ... etc
> ```

Updated per your recommendation, thank you!

src/crypto/client/NUT11.ts
```diff
@@ -18,14 +18,14 @@ export const createP2PKsecret = (pubkey: string): string => {
 	return JSON.stringify(newSecret);
 };
 
-export const signP2PKSecret = (secret: string, privateKey: PrivKey): string => {
-	const msghash = sha256(secret);
+export const signP2PKSecret = (secret: string, privateKey: Uint8Array): string => {
+	const msghash = sha256(new TextEncoder().encode(secret));
```
Egge21M, line 22: Review note: v2 no longer allows hashing on strings. UTF-8 encoding mimics the old behavior

src/crypto/client/NUT11.ts
```diff
@@ -262,19 +262,21 @@ export const signP2PKProofs = (
  * requires a signature from the key.
  *
  * @param proof - A proof to sign.
- * @param privateKey - A single private key.
+ * @param privateKey - A single private key (hex string or Uint8Array).
  * @throws Error if signature is not required or proof is already signed.
  */
-export const signP2PKProof = (proof: Proof, privateKey: string): Proof => {
+export const signP2PKProof = (proof: Proof, privateKey: string | Uint8Array): Proof => {
```
robwoodgate, line 268: Widening privateKey to string | Uint8Array will inadvertently create an undocumented breaking change for v3, but it's not a bad idea, so we should add it to https://github.com/cashubtc/cashu-ts/pull/432/

src/crypto/mint/NUT12.ts
```diff
@@ -1,25 +1,25 @@
...
+	const A = secp256k1.Point.BASE.multiply(n_a); // A = aG
 	const e = hash_e([R_1, R_2, A, C_]); // e = hash(R1, R2, A, C_)
-	const n_r = hexToNumber(r);
 	const n_e = bytesToNumber(e);
-	const n_a = bytesToNumber(a);
 	// WARNING: NON-CONSTANT TIME OPERATIONS?
-	const s = numberToBytesBE((n_r + n_e * n_a) % secp256k1.CURVE.n, 32); // (r + ea) mod n
+	const s = numberToBytesBE((n_r + n_e * n_a) % secp256k1.Point.CURVE().n, 32); // (r + ea) mod n
 	return { s, e };
 };
```
robwoodgate, line 24: This function should be as per v3:

```ts
export const createDLEQProof = (B_: WeierstrassPoint<bigint>, a: Uint8Array): DLEQ => {
	const r = secp256k1.Point.Fn.fromBytes(createRandomPrivateKey()); // r <- random (Uint8Array)
	const R_1 = secp256k1.Point.BASE.multiply(r); // R1 = rG
	const R_2 = B_.multiply(r); // R2 = rB_
	const scalar_a = secp256k1.Point.Fn.fromBytes(a);
	const C_ = B_.multiply(scalar_a); // C_ = aB_
	const A = secp256k1.Point.BASE.multiply(scalar_a); // A = aG
	const e = hash_e([R_1, R_2, A, C_]); // e = hash(R1, R2, A, C_)
	const scalar_e = secp256k1.Point.Fn.fromBytes(e);
	// Use field operations for constant-time addition and multiplication
	const s_scalar = secp256k1.Point.Fn.add(r, secp256k1.Point.Fn.mul(scalar_e, scalar_a));
	const s = numberToBytesBE(s_scalar, 32); // s = (r + e * a) mod n
	return { s, e };
};
```

ersclaytDocumentsProgrammingcashu-ts
```diff
@@ -0,0 +1,40 @@
+{
```
robwoodgate: Not sure what this change  to the consumer-reactnative smoke test (which seems to be failing) has do do with the noble upgrade. Is this intended?

robwoodgate (changes requested):
Thanks again for taking care of this, much appreciated. I've highlighted a couple of areas we need to address - otherwise am happy to LGTM.

ersclaytDocumentsProgrammingcashu-ts
Egge21M: Great catch, I am not sure how I missed this, but this is the reason why CI is failing. I don't think the CI failing is a blocker for the feature itself, but I also don't get why this is here.

src/crypto/client/NUT11.ts
robwoodgate: Have raised in the other PR, so marking this as resolved.

ersclaytDocumentsProgrammingcashu-ts
Egge21M: @happylemonprogramming what was the reason to add this?

happylemonprogramming: The "package.json" was added in an attempt to configure Jest to handle the @noble v2.x ES modules and address the test error, but it was not the right approach. I removed it.

src/crypto/mint/NUT12.ts
happylemonprogramming: Updated!

ersclaytDocumentsProgrammingcashu-ts
Egge21M: Copy that! CI is still failing though :/ I think the culprit is the react-native package that is now installed by the test runner

robwoodgate (approved):
LGTM. Thanks for taking care of this @happylemonprogramming

robwoodgate:
The ReactNative issue was exposed by this PR, but not caused by it.

Upgrading to noble v2 made @noble/* ESM only, but our CJS output was still externalising them, so CJS consumers (Jest, React Native smoke test) ended up doing require() on an ESM file and crashed with “Cannot use import statement outside a module”.

I fixed it by bundling the ESM only deps into the CJS build, so the CJS entrypoint never tries to require ESM packages.

Follow up in 992cf0f4 fixes RN, Jest by bundling ESM only deps and reverting the RN smoke test changes made in this PR

PR #441, opened 2025-12-19
fix: validate proof amounts in token decoding

zk-bits:
## Summary

  - Add validation to ensure proof amounts are finite non-negative integers when decoding tokens
  - Applies to both V3 (cashuA) and V4 (cashuB) token formats
  - Rejects tokens with invalid amount values (non-numeric, negative, non-integer, ornon-finite)

  ## Changes

  - Added `validateProofAmount()` helper function in `src/utils/core.ts`
  - Validation is applied in `tokenFromTemplate()` for V4 tokens
  - Validation is applied in `handleTokens()` for V3 tokens
  - Added test coverage for the validation logic

  ## Test plan

  - [x] All existing tests pass (`npm test`)
  - [x] Lint passes (`npm run lint`)
  - [x] Format passes (`npm run format`)
  - [x] Build passes (`npm run compile`)
  - [x] API unchanged (`npm run api:check`)

### DIFF
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -353,10 +353,32 @@ function templateFromToken(token: Token): TokenV4Template {
 	return tokenTemplate;
 }
 
+/**
+ * Asserts amount is a valid, safe integer.
+ *
+ * @param amount Value to check.
+ * @param allowZero If false, requires amount > 0 (default: false).
+ */
+/**
+ * @internal
+ */
+export function validateAmount(amount: unknown, allowZero = false): asserts amount is number {
+	if (typeof amount !== 'number' || !Number.isFinite(amount) || !Number.isInteger(amount)) {
+		throw new Error(`Invalid amount: ${String(amount)}`);
+	}
+	if (!Number.isSafeInteger(amount)) {
+		throw new Error(`Amount must be a safe integer: ${amount}`);
+	}
+	if (allowZero ? amount < 0 : amount <= 0) {
+		throw new Error(`Amount must be ${allowZero ? 'non-negative' : 'positive'}: ${amount}`);
+	}
+}
+
 function tokenFromTemplate(template: TokenV4Template): Token {
 	const proofs: Proof[] = [];
 	template.t.forEach((t) =>
 		t.p.forEach((p) => {
+			validateAmount(p.a, true);
 			proofs.push({
 				secret: p.s,
 				C: bytesToHex(p.c),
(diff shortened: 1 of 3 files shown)


src/utils/core.ts
```diff
@@ -353,14 +353,28 @@ function templateFromToken(token: Token): TokenV4Template {
 	return tokenTemplate;
 }
 
+/**
+ * Validates that a proof amount is a safe positive integer. Rejects Infinity, NaN, negative values,
+ * and non-integers.
+ */
+function validateProofAmount(amount: unknown): number {
+	if (typeof amount !== 'number' || !Number.isFinite(amount)) {
```
Egge21M: Wallet exposes a private method `assertAmount` that has a lot of overlap with this. Maybe we can unify the two? I think using Number.isSafeInteger() might be even better here, as it checks for NaN, Infinity and > 2^53

Egge21M (changes requested):
LGTM! But there is some overlap with the private Wallet.assertAmount() method. This method should probably be a util as its stateless and then we can probably combine the two into one utility.

CC @robwoodgate

src/utils/core.ts
zk-bits: Good catch also cleaned up the PR a fair bit, much more streamlined now.

src/utils/core.ts
```diff
@@ -353,10 +353,32 @@ function templateFromToken(token: Token): TokenV4Template {
...
+ * Asserts amount is a valid, safe integer.
+ *
+ * @param amount Value to check.
+ * @param allowZero If false, requires amount > 0 (default: false).
+ */
+/**
+ * @internal
+ */
+export function validateAmount(amount: unknown, allowZero = false): asserts amount is number {
+	if (typeof amount !== 'number' || !Number.isFinite(amount) || !Number.isInteger(amount)) {
```
robwoodgate, line 366: Reviewer note: `!Number.isFinite(amount)` is technically redundant, as the infinity check is covered by `!Number.isinteger(amount)`, but does no harm to be super explicit.

src/utils/core.ts
robwoodgate: Great idea making this a generic utility we can use elsewhere

zk-bits:
> LGTM. It's not essential, as the raw proof checks cover it, but would be consistent if Wallet also did the assert check in `prepareSwapToReceive`, right after the zero total check:
> 
> 
> 
> ```ts
> 
> this.failIf(totalAmount === 0, 'Token contains no proofs', { proofs });
> 
> this.assertAmount(totalAmount, 'prepareSwapToReceive'); // add this?
> 
> ```

10/10 catch. Will modify.

robwoodgate (approved):

PR #442, opened 2025-12-19
fix: use string matching for protected endpoint patterns

zk-bits:
## Summary
- Replace regex-based endpoint matching with simple string operations
- Supports the same NUT-22 patterns: exact match, anchored exact, and prefix with `.*`

## Changes
- `^/v1/foo/bar$` → exact match with `===`
- `^/v1/foo/.*` → prefix match with `startsWith()`
- Plain `/v1/foo/bar` → exact match with `===`

### DIFF
--- a/src/model/MintInfo.ts
+++ b/src/model/MintInfo.ts
@@ -11,7 +11,7 @@ type Endpoint = { method: Method; path: string };
 type ProtectedIndex = {
 	cache: Record<string, boolean>; // "METHOD /v1/foo"
 	exact: Array<{ method: Method; path: string }>;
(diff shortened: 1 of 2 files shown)


robwoodgate (approved):

robwoodgate:
The only suggestion I can make to improve would be to add a small test that covers the wildcard case so it hits line 150 of `MintInfo.buildIndex`

```ts
if (p.endsWith('.*')) {
	prefix.push({ method: e.method, path: p.slice(0, -2) });
```

zk-bits:
this can be merged now! @Egge21M

PR #454, opened 2026-01-03
KDF V2: Proper blinding factor reduction per NUT-13 spec

callebtc:
Implements the correct KDF for V2 keysets (01-prefixed) as specified in the updated Keyset V2 version of NUT-13 ([PR](https://github.com/cashubtc/nuts/pull/182)).

Changes:
- Blinding factors are now reduced modulo SECP256K1_N to ensure valid private key scalars
- Added optimization: single subtraction instead of full modulo (only ~2^-128 of HMAC outputs need reduction)
- Added Bytes.toBigInt() and Bytes.fromBigInt() utility methods
- **Deprecated bytesToNumber()** in favor of Bytes.toBigInt() for better performance

Modulo optimization: A 256-bit HMAC output only needs one subtraction to be < SECP256K1_N. Full modulo is ~10-20x more expensive than a single comparison and conditional subtraction.

### DIFF
--- a/src/crypto/core.ts
+++ b/src/crypto/core.ts
@@ -24,7 +24,7 @@ export type RawBlindedMessage = {
 };
 
 /**
- * @deprecated - Use RawBlindedMessage.
+ * @deprecated - Use {@link RawBlindedMessage}.
  */
 export type BlindedMessage = RawBlindedMessage;
 
@@ -109,7 +109,7 @@ export function createBlindSignature(
 }
 
 /**
- * @deprecated - Use createRandomRawBlindedMessage()
+ * @deprecated - Use {@link createRandomRawBlindedMessage}
  */
 export function createRandomBlindedMessage(_deprecated?: PrivKey): RawBlindedMessage {
 	void _deprecated; // intentionally unused
--- a/src/crypto/NUT11.ts
+++ b/src/crypto/NUT11.ts
@@ -681,7 +681,7 @@ export type WellKnownSecret = 'P2PK';
 /**
  * Signs a P2PK secret using Schnorr.
  *
- * @deprecated Use schnorrSignMessage()
+ * @deprecated Use {@link schnorrSignMessage}
  */
 export const signP2PKSecret = (secret: string, privateKey: PrivKey): string => {
 	return schnorrSignMessage(secret, privateKey);
@@ -690,7 +690,7 @@ export const signP2PKSecret = (secret: string, privateKey: PrivKey): string => {
 /**
  * Verifies a Schnorr signature on a P2PK secret.
  *
- * @deprecated Use schnorrVerifyMessage()
+ * @deprecated Use {@link schnorrVerifyMessage}
  */
 export const verifyP2PKSecretSignature = (
 	signature: string,
@@ -701,14 +701,14 @@ export const verifyP2PKSecretSignature = (
 };
 
 /**
- * @deprecated - Typo: use getP2PKExpectedWitnessPubkeys() instead.
+ * @deprecated - Typo: use {@link getP2PKExpectedWitnessPubkeys} instead.
  */
 export function getP2PKExpectedKWitnessPubkeys(secretStr: string | Secret): string[] {
 	return getP2PKExpectedWitnessPubkeys(secretStr);
 }
 
 /**
- * @deprecated Use isP2PKSpendAuthorised or verifyP2PKSpendingConditions instead.
+ * @deprecated Use {@link isP2PKSpendAuthorised} or {@link verifyP2PKSpendingConditions} instead.
  */
 export function verifyP2PKSig(proof: Proof): boolean {
 	return isP2PKSpendAuthorised(proof);
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -175,8 +175,8 @@ export function hasCorrespondingKey(amount: number, keyset: Keys): boolean {
 /**
  * Converts a bytes array to a number.
  *
- * @deprecated Use {@link Bytes.toBigInt} instead for better performance. This function is kept for
- *   backward compatibility.
+ * @deprecated We now use Bytes.toBigInt internally for better performance. This function is kept
+ *   public for backward compatibility.
  * @param bytes To convert to number.
  * @returns Number as bigint.
  */


robwoodgate:
@callebtc  - Can you rebase this to the development branch please

src/index.ts
```diff
@@ -52,7 +52,7 @@ export type { Token, TokenMetadata } from './model/types/token';
 export * from './crypto';
 
 // Core Utils
-export * from './utils/core';
+export * from './utils';
```
Egge21M: Why is this export being widened?

robwoodgate: This is not a great idea as it puts all our internal utils into the public API (as can be seen in the `etc/cashu-ts.api.md` diff). Consumers can still unofficially "borrow" these utils by deep linking the import, but if they go into the public API, we are constrained if we wish to change them.

src/utils/Bytes.ts
```diff
@@ -102,4 +102,25 @@ export class Bytes {
...
+			return new Uint8Array(1);
+		}
+		const bytes: number[] = [];
+		let temp = value;
+		while (temp > 0n) {
+			bytes.unshift(Number(temp & 0xffn));
+			temp >>= 8n;
+		}
+		return new Uint8Array(bytes);
+	}
```
robwoodgate, line 136: Unshift is O(n²) as the array grows, so would recommend we tweak it to "push then reverse" for efficiency. Switching to Uint8Array([0]) for the zero case to make the intent more explicit. Also, since this is unsigned only, we should probably guard `<0n`.

```suggestion
	static fromBigInt(value: bigint): Uint8Array {
		if (value < 0n) {
			throw new RangeError('value must be non-negative');
		}
		if (value === 0n) {
			return new Uint8Array([0]);
		}
		const bytes: number[] = [];
		let temp = value;
		while (temp > 0n) {
			bytes.push(Number(temp & 0xffn));
			temp >>= 8n;
		}
		bytes.reverse();
		return new Uint8Array(bytes);
	}
```

If we want to be even more efficient, we could avoid the intermediate bytes array by calculating the UInt8Array length up front and then filling it from the end. Something like:

```typescript
	static fromBigInt(value: bigint): Uint8Array {
		if (value < 0n) {
			throw new RangeError('value must be non-negative');
		}
		if (value === 0n) {
			return new Uint8Array([0]);
		}
		// Calculate Uint8Array length
		let temp = value;
		let length = 0;
		while (temp > 0n) {
			length++;
			temp >>= 8n;
		}
		// Fill it from the end (big endian)
		const out = new Uint8Array(length);
		temp = value;
		for (let i = length - 1; i >= 0; i--) {
			out[i] = Number(temp & 0xffn);
			temp >>= 8n;
		}
		return out;
	}
```

src/crypto/NUT12.ts
```diff
@@ -19,9 +19,9 @@ export const verifyDLEQProof = (
 	A: WeierstrassPoint<bigint>,
 ) => {
 	const sG = secp256k1.Point.BASE.multiply(secp256k1.Point.Fn.fromBytes(dleq.s));
-	const eA = A.multiply(bytesToNumber(dleq.e));
-	const sB_ = B_.multiply(bytesToNumber(dleq.s));
-	const eC_ = C_.multiply(bytesToNumber(dleq.e));
+	const eA = A.multiply(Bytes.toBigInt(dleq.e));
+	const sB_ = B_.multiply(Bytes.toBigInt(dleq.s));
+	const eC_ = C_.multiply(Bytes.toBigInt(dleq.e));
```
robwoodgate: These actually DO want to be `modulo n` to ensure they are valid for point operations. In the development branch, we have replaced `bytesToNumber` with `secp256k1.Point.Fn.fromBytes`, which applies `modulo n`.

src/crypto/NUT01.ts
```diff
@@ -75,6 +75,6 @@ export function createNewMintKeys(pow2height: IntRange<0, 65>, seed?: Uint8Array
 
 export function verifyProof(proof: RawProof, privKey: Uint8Array): boolean {
 	const Y: WeierstrassPoint<bigint> = hashToCurve(proof.secret);
-	const aY: WeierstrassPoint<bigint> = Y.multiply(bytesToNumber(privKey));
+	const aY: WeierstrassPoint<bigint> = Y.multiply(Bytes.toBigInt(privKey));
```
robwoodgate: Same as NUT12 comment - in development branch, we've used `secp256k1.Point.Fn.fromBytes`, which applies `modulo n` to make the privkey point operation safe.

src/crypto/core.ts
```diff
@@ -94,22 +94,22 @@ export function createBlindSignature(
...
+		Bytes.toBigInt(secp256k1.utils.randomSecretKey()),
 		privateKey,
 	);
 }
 
 export function blindMessage(secret: Uint8Array, r?: bigint, privateKey?: PrivKey): BlindedMessage {
 	const Y = hashToCurve(secret);
 	if (!r) {
-		r = bytesToNumber(secp256k1.utils.randomSecretKey());
+		r = Bytes.toBigInt(secp256k1.utils.randomSecretKey());
```
robwoodgate: Again, in development branch, we've used `secp256k1.Point.Fn.fromBytes`, but we could use `Bytes.toBigInt` too because `secp256k1.utils.randomSecretKey` is guaranteed to be valid.

src/crypto/core.ts
```diff
@@ -94,22 +94,22 @@ export function createBlindSignature(
...
-	const C_: WeierstrassPoint<bigint> = B_.multiply(bytesToNumber(privateKey));
+	const C_: WeierstrassPoint<bigint> = B_.multiply(Bytes.toBigInt(privateKey));
 	return { C_, amount, id };
 }
 
 export function createRandomBlindedMessage(privateKey?: PrivKey): BlindedMessage {
 	return blindMessage(
 		randomBytes(32),
-		bytesToNumber(secp256k1.utils.randomSecretKey()),
+		Bytes.toBigInt(secp256k1.utils.randomSecretKey()),
```
robwoodgate: Again, in development branch, we've used `secp256k1.Point.Fn.fromBytes`, but we could use `Bytes.toBigInt` too because `secp256k1.utils.randomSecretKey` is guaranteed to be valid.

src/crypto/NUT13.ts
```diff
@@ -63,8 +65,22 @@ const derive = (
...
+		// Optimization: single subtraction instead of modulo
+		// Probability of HMAC >= SECP256K1_N is ~2^-128
+		if (x >= SECP256K1_N) {
+			return Bytes.fromBigInt(x - SECP256K1_N);
+		}
+		if (x === 0n) {
+			throw new Error('Derived invalid blinding scalar r == 0');
+		}
+		return hmacDigest;
+	}
```
robwoodgate, line 81: I LOVE the explicit single reduction concept. Definitely a great move for reducing `SHA256` to a value within `N`. And reducing means the only invalid scalar is `r=0`, which is astronomically unlikely.

Noting that, as we've discussed in the comments and elsewhere, this may cause compatibility problems in certain implementations that cannot handle bigint subtraction by default (eg RUST).

src/index.ts
Egge21M: Yes, I agree. I wondered whether I missed something, because nothing inside this PR needs this widening iiuc

robwoodgate: Deprecating the public `bytesToNumber` in favour of the internal `Bytes` class means if/when we remove the `bytesToNumber` function, there will be no alternative in the public API.

I suspect this is why the export was widened, but we should think carefully before making our internal utils public API.

Issue #446, opened 2025-12-20
Export IIFE Build?


robwoodgate:
At the moment, I don’t think we actually ship a standalone browser build (IIFE) as part of our published artefacts.

We can build it locally via Vite (`npm run compile:standalone`), but consumers cannot currently do the simple “drop a script tag in a page and use a global” workflow.

Should we add the IIFE build to our npm package (for example under `dist/`) so it can be consumed via a CDN like jsDelivr or unpkg?

KvngMikey:
please can this PR https://github.com/cashubtc/cashu-ts/pull/444 go in if it's all ready before i treat this iife issue, so i'm working with the fixes in the vite.config and package.json @robwoodgate has done in the existing PR already?

robwoodgate:
> please I noticed this PR [#444](https://github.com/cashubtc/cashu-ts/pull/444) has fixes in the vite.config and package.json [@robwoodgate](https://github.com/robwoodgate) has done, i just wanted to flag the overlap, it would be nice to have it go in first if it's ready to avoid conflicts, but I can base the IIFE-shipping fix either on the current dev branch or on top of this PR. Happy to proceed whichever way you prefer.

Please base it on top, as that PR lays the groundwork for IIFE building, and should be merged this week when @Egge21M gets a chance to look.

PR #457, opened 2026-01-08
Export IIFE Build

KvngMikey:
# Fixes: #446

## Description
This PR ships the existing IIFE (standalone browser) build as part of the published npm package and documents how to consume it via a CDN.

The IIFE build was already supported by Vite `(compile:standalone)` and covered by consumer tests, but it was not included in published artifacts due to `.npmignore`. This change makes the build accessible to CDN users (jsDelivr / unpkg) without altering runtime behavior or introducing new build toggles.

## Changes

- Publish IIFE build artifacts by removing lib/cashu-ts.iife.js and lib/cashu-ts.iife.js.map from .npmignore, which now allows consumption via CDN using a simple <script> tag.

- Added a short “Browser / CDN usage (IIFE)” section to the README, that explicitly shows the global (cashuts) exposed by the IIFE build and basic initialization.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`
- [x] npm run compile:standalone
- [x] npm run test:iife 
- [x] npm pack

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -1,6 +1,6 @@
 name: Publish Package to npmjs
 permissions:
-  contents: read
+  contents: write
   id-token: write
 on:
   release:
@@ -30,3 +30,12 @@ jobs:
       - run: npm publish --provenance --tag ${{ steps.npm-tag.outputs.tag }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - run: npm run build:standalone
+      - name: Upload standalone bundle assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload "${{ github.event.release.tag_name }}" \
+            "./lib/cashu-ts.iife.js" \
+            "./lib/cashu-ts.iife.js.map" \
+            "./lib/types/index.d.ts"


KvngMikey:
this PR builds on top of this PR => https://github.com/cashubtc/cashu-ts/pull/444, so the former would need to be merged first, thank you.

.npmignore
```diff
@@ -10,5 +10,3 @@ vitest.workspace.ts
 tsconfig.json
 src
 examples
-/lib/cashu-ts.iife.js
```
Egge21M, line 13: I think this should stay, as otherwise the iife becomes part of the bundled npm package.

robwoodgate:
Stacked pending #444 merge

.npmignore
robwoodgate: jsdelivr appears to grab the `lib` folder from npm and currently assumes the old `post-process-dts.js` in src folder is the default - yikes!.

https://www.jsdelivr.com/package/npm/@cashu/cashu-ts

UPKG also publishes from npm, but doesn't have a default by looks of it:

https://app.unpkg.com/@cashu/cashu-ts@3.2.2

So I think we probably DO need to publish IIFE as part of the npm bundle to get it hosted... ie remove both these from `.npmignore`.

```
/lib/cashu-ts.iife.js
/lib/cashu-ts.iife.js.map
```

And ensure our `prerelease.yaml` CI builds it AFTER the main compile step then it will always be up-to-date with our other builds, and we won't have to commit it in github (our `.gitignore` excludes `lib`).

```diff
        - run: npm run compile
+       - run: npm run compile:standalone
        - run: npm publish --tag next --provenance
```

Then it looks like we should also signal to unpkg / jsdelivr using the following in `package.json`.

```diff
        "main": "lib/cashu-ts.cjs",
        "module": "lib/cashu-ts.es.js",
        "types": "lib/types/index.d.ts",
+       "jsdelivr": "lib/cashu-ts.iife.js",
+       "unpkg": "lib/cashu-ts.iife.js",
```

Egge21M: Including the standalone build increases the bundle size by about 42%. I don't think that is worth it

robwoodgate: > Including the standalone build increases the bundle size by about 42%. I don't think that is worth it

Agree we should strive to [avoid bloat](https://github.com/cashubtc/cashu-ts/issues/461). The IIFE is 172Kb, its the map at 868Kb that bloats it.

I guess it comes back to whether we WANT to supply a simple single url `<script src="…"></script>` with `window.Cashu` stye option or not.

If an IIFE is desirable, maybe we can post-process away the sourcemap reference in the IIFE (`//# sourceMappingURL=cashu-ts.iife.js.map`) and just publish the smaller IIFE?

Or if we are happy to force a modern `<script type="module"></script>` instead, we can skip IIFE entirely.

Eg

```ts
<script type="module">
  import { Wallet } from "https://unpkg.com/cashu-ts@3.2.2@/lib/cashu-ts.es.js";
</script>

```

Maybe the halfway house is to do what noble does and release the IIFE/standalone as part of the release assets, so people can download an bundle themselves, but NOT publish it to npm for CDN?

Egge21M: > Maybe the halfway house is to do what noble does and release the IIFE/standalone as part of the release assets, so people can download an bundle themselves, but NOT publish it to npm for CDN?

Yeah, this is what I was talking about in dev call. Have the bundle in the Github release so that people can download it ([where is the exe style](https://github.com/anus-dev/ANUS/issues/24#issue-2910497604)).

robwoodgate: > > Maybe the halfway house is to do what noble does and release the IIFE/standalone as part of the release assets, so people can download an bundle themselves, but NOT publish it to npm for CDN?
> 
> Yeah, this is what I was talking about in dev call. Have the bundle in the Github release so that people can download it ([where is the exe style](https://github.com/anus-dev/ANUS/issues/24#issue-2910497604)).

Ah sorry, I didn't grok that in the call... YES, that is a much better compromise.

KvngMikey: > > Maybe the halfway house is to do what noble does and release the IIFE/standalone as part of the release assets, so people can download an bundle themselves, but NOT publish it to npm for CDN?
> 
> Yeah, this is what I was talking about in dev call. Have the bundle in the Github release so that people can download it ([where is the exe style](https://github.com/anus-dev/ANUS/issues/24#issue-2910497604)).

the link attached is quite funny and really drives home the point you were trying to make, I have reverted the change to the npmignore to reduce the bundle size like you pointed out and updated the readMe to point out new intent to users...

PR #458, opened 2026-01-08
feat: update Keyset ID V2 derivation to latest spec

callebtc:
This PR updates the Keyset ID V2 calculation to align with the latest specification in cashubtc/nuts#182.

### Changes:
- Added `input_fee_ppk` to `MintKeys` type.
- Updated `deriveKeysetId` to use the new UTF-8 string-based preimage for v2 keysets (prefix `01`).
- Added ambiguity check to `mapShortKeysetIds`.
- Updated verification logic in `Keyset` and `utils/core`.
- Added latest test vector to `core.test.ts`.

**EDIT:**

- Adds new options signature for `deriveKeysetId` and updates call sites.
- Deprecates old `deriveKeysetId` signature
- Simplifies key sorting and pubkey concat, drops redundant types
- Updates `createNewMintKeys` for v2 keysets.

References: cashubtc/nuts#182

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -187,7 +187,7 @@ export function createNewMintKeys(pow2height: IntRange<0, 65>, seed?: Uint8Array
     expiry?: number;
     input_fee_ppk?: number;
     unit?: string;
-    versionByte: number;
+    versionByte?: number;
 }): KeysetPair;
 
 // @public


src/utils/core.ts
```diff
@@ -511,6 +511,7 @@ export function deriveKeysetId(
 	keys: Keys,
 	unit?: string,
 	expiry?: number,
+	input_fee_ppk?: number,
```
robwoodgate, line 539: This will be a breaking change to the signature if used by an outside consumer.

Should we accept this risk, or move the param to last position (even though that is mentally clunky)?

Egge21M: Needs to come last IMHO. It's not too bad, as other params are effectively optional (all functions with mutliple optional params suck)

test/wallet/wallet.node.test.ts
```diff
@@ -51,12 +51,11 @@ const dummyKeysResp = {
...
-const dummyKeysId = deriveKeysetId(
-	dummyKeysResp.keysets[0].keys,
-	dummyKeysResp.keysets[0].unit,
-	1754296607,
-	1,
-);
+const dummyKeysId = deriveKeysetId(dummyKeysResp.keysets[0].keys, {
+	expiry: 1754296607,
+	unit: dummyKeysResp.keysets[0].unit,
+	versionByte: 0,
```
robwoodgate, line 57: Reviewer note: This is just to keep the test file honest. We don't actually use `dummyKeysId` except to output to console, so it's kind of redundant as it stands. Reverting the versionByte to 0 returns / confirms the hard coded keyset ID used in the dummy keysets.

Unit tests for `deriveKeysetId` are in the `core.test.test` file.

robwoodgate (approved):
LGTM

src/crypto/NUT01.ts
```diff
@@ -45,7 +45,29 @@ export function getPubKeyFromPrivKey(privKey: Uint8Array) {
...
+ * @throws If keyset versionByte is not valid.
+ */
+export function createNewMintKeys(
+	pow2height: IntRange<0, 65>,
+	seed?: Uint8Array,
+	options?: {
+		expiry?: number;
+		input_fee_ppk?: number;
+		unit?: string;
+		versionByte: number;
```
Egge21M: Implementation defaults to `1`. So this is actually not required

src/utils/core.ts
```diff
@@ -497,61 +497,104 @@ export function handleTokens(token: string): Token {
...
 			if (!unit) {
 				throw new Error('Cannot compute keyset ID version 01: unit is required.');
 			}
-			pubkeysConcat = mergeUInt8Arrays(pubkeysConcat, Bytes.fromString('unit:' + unit));
+			const sortedEntries = Object.entries(keys).sort(
+				([amountA], [amountB]) => Number(amountA) - Number(amountB),
+			);
+			let preimage = sortedEntries.map(([amount, pubkey]) => `${amount}:${pubkey}`).join(',');
+			preimage += `|unit:${unit}`;
+			if (input_fee_ppk) {
```
Egge21M, line 588: A fee of `0` will not be encoded. Is that intentional?

Egge21M (changes requested):
Thank you for taking this PR over @robwoodgate. I noticed two things, one nit but one potential bug.

src/utils/core.ts
robwoodgate: This is intentional:

https://github.com/cashubtc/nuts/blob/main/02.md?plain=1#L69
```
4 - If input_fee_ppk is specified and non-zero, add the UTF8-encoded string prefixed with "|input_fee_ppk:" (e.g. "|input_fee_ppk:100"). If input_fee_ppk is omitted, null, or 0, it MUST be omitted from the preimage.
```

src/crypto/NUT01.ts
robwoodgate: Good spot, thank you

```suggestion
		versionByte?: number;
```

PR #459, opened 2026-01-09
feat(build): enable API Extractor dtsRollup for proper TSDoc release tag support

### LINKED ISSUE #416: Handling of alpha and beta TSDOC tags
Our vite build process excludes `alpha` and `beta` tagged items from the public API.

This is not a showstopper, as we can tag items as `experimental` to surface them while flagging them as subject to change.

Should we consider switching from using vite for our rollups in favor of API Extractor's more flexible handling of development tags, or just keep as is and use `experimantal` as needed?

This came up because we marked P2BK features as alpha, and they are not surfaced. See #415

yashbhutwala:
## Summary

Enables API Extractor's dtsRollup to generate trimmed type definitions for different release levels.

## Changes

- **api-extractor.json**: Enable `dtsRollup` with trimmed outputs:
  - `cashu-ts-alpha.d.ts` - excludes `@internal`
  - `cashu-ts-beta.d.ts` - excludes `@alpha`, `@internal`
  - `cashu-ts-public.d.ts` - excludes `@alpha`, `@beta`, `@internal`
- **vite-plugin-dts**: Keeps `rollupTypes: true` for main `index.d.ts`
- **API report**: Unchanged from upstream (no @internal items)

## Note

To use release tag trimming, source files need `@alpha`/`@beta` tags. Currently P2BK features use `@experimental` which API Extractor doesn't filter (it only recognizes standard TSDoc release tags).

Closes #416

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api-extractor.json
+++ b/api-extractor.json
@@ -53,77 +53,41 @@
 
 	/**
 	 * Configures how the .d.ts rollup file will be generated.
+	 *
+	 * API Extractor handles dtsRollup to properly support TSDoc release tags:
+	 * - @alpha: Included in alpha builds (untrimmed), excluded from public builds
+	 * - @beta: Included in alpha and beta builds, excluded from public builds
+	 * - @public: Included in all builds (default if no tag)
+	 * - @internal: Excluded from all builds
 	 */
 	"dtsRollup": {
-		/**
-		 * (REQUIRED) Whether to generate the .d.ts rollup file.
-		 */
-		"enabled": false
+		"enabled": true,
 
 		/**
-		 * Specifies the output path for a .d.ts rollup file to be generated without any trimming.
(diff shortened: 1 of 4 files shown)


yashbhutwala:
> Something seems a little screwy, as the API doc now includes all our internal functions

Fixed - kept `rollupTypes: true` in vite-plugin-dts so the API report stays clean.

> the change to the compile package script sets my spidey senses off

Good catch - removed it. `api:check`/`api:update` handle api-extractor separately now.

One thing I noticed: source uses `@experimental` not `@alpha`/`@beta`, so API Extractor won't actually trim those items (it only knows standard TSDoc release tags).

PR #470, opened 2026-01-21
Fix/preparemelt custom outputtype

TheRealCheebs:
# Fixes: #[issue]

This allows custom outputType to be passed into the prepareMelt functions. Previously it threw an error.

## Description

This commit changes the behavior of the `prepareMelt` function to allow custom
`OutputType` without modification. Previously, custom `OutputType` would
result in a failure. The logic now ensures that custom types are used
as-is, while non-custom types continue to have denominations merged
and counters assigned as needed.

Additionally a test is added to validate nothing changes in the custom provided outputs for prepareMelt.
...

## Changes

- ...
- ...

## PR Tasks

- [x] Open PR
- [ ] run `npm run lint` --> no warnings or errors
- [ ] run `npm run format`

### DIFF
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -2044,9 +2044,12 @@ class Wallet {
 			quoteAmount: meltQuote.amount,
 		});
 
+		if (outputType.type === 'custom') {
+			outputData = outputType.data;
+		}
 		// Create NUT-08 blanks for return of Lightning fee change
 		// Note: zero amount + zero denomination passes splitAmount validation
-		if (feeReserve > 0) {
+		else if (feeReserve > 0) {
 			let count = Math.ceil(Math.log2(feeReserve)) || 1;
 			if (count < 0) count = 0; // Prevents: -Infinity
 			const denominations: number[] = count ? new Array<number>(count).fill(0) : [];
@@ -2055,20 +2058,15 @@ class Wallet {
 				denominations,
 			});
 
-			let meltOT: OutputType = outputType;
-			// When user provides custom outputType use it without modification.
-			if (outputType.type !== 'custom') {
-				meltOT = { ...outputType, denominations };
-				// Assign counter atomically if OutputType is deterministic
-				// and the counter is zero (auto-assign)
-				const autoCounters = await this.addCountersToOutputTypes(keyset.id, meltOT);
-				[meltOT] = autoCounters.outputTypes;
-				if (autoCounters.used) {
-					this.safeCallback(onCountersReserved, autoCounters.used, { op: 'meltProofs' });
-				}
-				this._logger.debug('melt counter', { counter: autoCounters.used, meltOT });
+			let meltOT: OutputType = { ...outputType, denominations };
+			// Assign counter atomically if OutputType is deterministic
+			// and the counter is zero (auto-assign)
+			const autoCounters = await this.addCountersToOutputTypes(keyset.id, meltOT);
+			[meltOT] = autoCounters.outputTypes;
+			if (autoCounters.used) {
+				this.safeCallback(onCountersReserved, autoCounters.used, { op: 'meltProofs' });
 			}
-
+			this._logger.debug('melt counter', { counter: autoCounters.used, meltOT });
 			// Generate the blank outputs (no fees as we are receiving change)
 			// Remember, zero amount + zero denomination passes splitAmount validation
 			outputData = this.createOutputData(0, keyset, meltOT);


src/wallet/Wallet.ts
```diff
@@ -2055,19 +2055,19 @@ class Wallet {
...
-			// Assign counter atomically if OutputType is deterministic
-			// and the counter is zero (auto-assign)
-			const autoCounters = await this.addCountersToOutputTypes(keyset.id, meltOT);
-			[meltOT] = autoCounters.outputTypes;
-			if (autoCounters.used) {
-				this.safeCallback(onCountersReserved, autoCounters.used, { op: 'meltProofs' });
+			let meltOT: OutputType = outputType;
+			// When user provides custom outputType use it without modification.
+			if (outputType.type !== 'custom') {
+				meltOT = { ...outputType, denominations };
```
Egge21M: I don't think this is necessary. We don't need to add the denominations and also we don't need to call addCountersToOutputTypes when passing custom data. @robwoodgate please let me know if I am mistaken here, but I think it would be best to skip all this and simply call createOutputData with the custom OT

src/wallet/Wallet.ts
```diff
@@ -2055,19 +2055,19 @@ class Wallet {
...
-			let meltOT: OutputType = { ...outputType, denominations };
-			// Assign counter atomically if OutputType is deterministic
-			// and the counter is zero (auto-assign)
-			const autoCounters = await this.addCountersToOutputTypes(keyset.id, meltOT);
-			[meltOT] = autoCounters.outputTypes;
-			if (autoCounters.used) {
-				this.safeCallback(onCountersReserved, autoCounters.used, { op: 'meltProofs' });
+			let meltOT: OutputType = outputType;
+			// When user provides custom outputType use it without modification.
+			if (outputType.type !== 'custom') {
```
Egge21M:
```suggestion
			if (outputType.type !== 'custom') {
        outputData = outputType.data;
      } else {
      			let meltOT: OutputType = { ...outputType, denominations };
			// Assign counter atomically if OutputType is deterministic
			// and the counter is zero (auto-assign)
			const autoCounters = await this.addCountersToOutputTypes(keyset.id, meltOT);
			[meltOT] = autoCounters.outputTypes;
			if (autoCounters.used) {
				this.safeCallback(onCountersReserved, autoCounters.used, { op: 'meltProofs' });
			}
			this._logger.debug('melt counter', { counter: autoCounters.used, meltOT });

			// Generate the blank outputs (no fees as we are receiving change)
			// Remember, zero amount + zero denomination passes splitAmount validation
			outputData = this.createOutputData(0, keyset, meltOT);
}
```

Egge21M (changes requested):
When a consumer provides OutputData for this method, we don't need to do anything, but pass this outputData along. Providing custom OutputData always equals the "Trust me bro" escape hatch.

src/wallet/Wallet.ts
robwoodgate: What @TheRealCheebs put is correct, as `createOutputData` will skip denoms for custom type and just do a basic sense check that the amount tallies with the outputs.

Your way is (almost) also correct, just even more 'trust me bro'... I think you have forgotten to flip the equality - should be:

```ts
if (outputType.type === 'custom')
```

If we are happy to be completely 'trust me bro', we could flatten it a bit more with something like:

```ts
		// Custom OT handles change manually, pass it through
		if (outputType.type === 'custom') {
			outputData = outputType.data;
		}

		// Create NUT-08 blanks for return of Lightning fee change
		// Note: zero amount + zero denomination passes splitAmount validation
		else if (feeReserve > 0) {
			let count = Math.ceil(Math.log2(feeReserve)) || 1;
			if (count < 0) count = 0; // Prevents: -Infinity
			const denominations: number[] = count ? new Array<number>(count).fill(0) : [];
			this._logger.debug('Creating NUT-08 blanks for fee reserve', {
				feeReserve,
				denominations,
			});
			let meltOT: OutputType = { ...outputType, denominations };
			// Assign counter atomically if OutputType is deterministic
			// and the counter is zero (auto-assign)
			const autoCounters = await this.addCountersToOutputTypes(keyset.id, meltOT);
			[meltOT] = autoCounters.outputTypes;
			if (autoCounters.used) {
				this.safeCallback(onCountersReserved, autoCounters.used, { op: 'meltProofs' });
			}
			this._logger.debug('melt counter', { counter: autoCounters.used, meltOT });

			// Generate the blank outputs (no fees as we are receiving change)
			// Remember, zero amount + zero denomination passes splitAmount validation
			outputData = this.createOutputData(0, keyset, meltOT);
		}
```

robwoodgate:
> I see, no need to call this.createOutputData again when we already have it in the custom data. The only benefit is there is the additional check inside createOutputData that the sum of the output amounts is 0
> 
> ```
> 			case 'custom': {
> 				outputData = outputType.data;
> 				const customTotal = OutputData.sumOutputAmounts(outputData);
> 				this.failIf(
> 					customTotal !== amount,
> 					`Custom output data total (${customTotal}) does not match amount (${amount})`,
> 				);
> 
> 				break;
> 			}
> ```

Exactly. That's all createOutputData effectively does for custom type.

TheRealCheebs:
I guess the question comes back to, how much guard rails do we want? Do we want to let custom data ride, or make sure that the provided custom data output amount is 0?

Issue #453, opened 2026-01-03
Add CHANGELOG.md file


callebtc:
Might be useful since there are often changes under the hood.

robwoodgate:
We now have a changelog, automated with release-please:
https://github.com/cashubtc/cashu-ts/blob/main/CHANGELOG.md

PR #477, opened 2026-01-26
feat(ci): add release automation and husky

TheRealCheebs:
## Description

This PR introduces automated release management and improves CI resilience. It adds Release Please for automated versioning and changelogs, integrates Husky/Commitlint to ensure semantic commits, and patches the consumer smoke tests to ensure they pass in restricted environments.

You can see the PR opened by the bot with changelog [here](https://github.com/TheRealCheebs/cashu-ts/pull/11).
You can see the release-please workflow in the [actions-run](https://github.com/TheRealCheebs/cashu-ts/actions/runs/21362274721)

And you can see a longer changelog with the older version which I didn't have tagging correct [here](https://github.com/TheRealCheebs/cashu-ts/pull/9)
## Changes

Release Automation: Added release-please configuration and GitHub Actions workflow to automate SemVer-based releases.

Git Hooks: Integrated Husky and Commitlint to enforce [Conventional Commits](https://www.conventionalcommits.org/).

CI Stability: Updated test/consumer/utils.js to use the --ignore-scripts flag during the smoke test build process.

Dependency Management: Added necessary devDependencies to package.json to support the new automation tooling.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ test/__screenshots__
 .DS_Store
 *.tgz
 
+.githooks
 .husky/_


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: googleapis/release-please-action@v4
+        id: release
+        with:
+          release-type: node
+
+      - name: Publish/Tag Actions
+        if: ${{ steps.release.outputs.release_created }}
+        run: |
+          echo "The release PR was merged, we need to figure what else to do here... npm publish, docs..."
```
KvngMikey: Is this intentionally left as TODO for a follow-up PR, or should we complete it now ?

also i think the general consensus from the meeting yesterday was to rename `main` to `stable` just to have it as  the most stable version of the library...

so should the steps here in turn start pointing to `stable` instead of `main` or do we just point it to `development` since it would be the most active branch that does the npm publishing, docs update, etc ?

package.json
```diff
@@ -40,7 +40,11 @@
 		"api:update": "npm run compile && api-extractor run --local",
 		"setup-hooks": "chmod +x scripts/install-git-hooks.sh && ./scripts/install-git-hooks.sh",
 		"uninstall-hooks": "git config --local --unset core.hooksPath && [ \"${REMOVE_GITHOOKS:-0}\" = \"1\" ] && rm -rf .githooks || echo 'To remove the installed hooks set REMOVE_GITHOOKS=1'",
-		"prtasks": "npm run lint && npm run format && npm run api:update && npm run test && git status"
+		"prtasks": "npm run lint && npm run format && npm run api:update && npm run test && git status",
+		"prepare": "husky",
+		"setup-git": "git config commit.template .gitmessage",
+		"postinstall": "npm run setup-git",
```
KvngMikey: i might be wrong but the `postinstall` looks like it runs setup-git on every npm install including in CI environments or when users install the library. Should we add || true to prevent CI failures if git config fails ?

so we try it, but even if it fails, don’t stop the install.
`"postinstall": "npm run setup-git || true"`
or we can do it here: `"setup-git": "git config commit.template .gitmessage || true"` so we make it gracefully skip in non-dev environments.

package.json
```diff
@@ -40,7 +40,11 @@
...
+		"migrate-hooks": "REMOVE_GITHOOKS=1 npm run uninstall-hooks && npm install"
```
KvngMikey, line 46: is this still needed? It looks like a one-time migration from the old `.githooks/` setup to Husky. If the migration is complete, we could remove both `migrate-hooks` and `uninstall-hooks` to clean up package.json

KvngMikey:
just a few comments for us to clarify on, otherwise LGTM !

.github/workflows/release-please.yml
TheRealCheebs: After thinking about this and some discussion, I am going to remove this section. I had been thinking we could add in the versions and typedoc workflows here. But they are better in their own files. I am going to add to this PR an update in the typedoc workflow to trigger on release so we can move to the one main branch flow that was discussed yesterday.

package.json
TheRealCheebs: Perfect catch. I will remove the post install and add setup-git into the prepare call.

package.json
TheRealCheebs: I don't know how many folks installed the old hooks to start, but I would keep it as is, and mark it for removal in a month or two. It will only be run once, and its more of a helper task for folks while we migrate. I do have to update docs in here so I could also not include this migrate and just call it out in the docs?

KvngMikey: that makes sense, keeping it for a migration period is reasonable and we can document in the readMe `why it exists`, `how to use it` and `when it would be removed.` but then we gotta remember to come clean it all out in a month or two.

TheRealCheebs:
This PR should also address #453

.gitignore
```diff
@@ -7,4 +7,4 @@ test/__screenshots__
 .DS_Store
 *.tgz
 
-.githooks
+.husky/_
```
robwoodgate, line 11: I think, maybe let's keep the `.githooks` ignore - it does no harm and prevents an accidental push if a contributor forgets to run the remove script.

```suggestion
.githooks
.husky/_
```

Issue #468, opened 2026-01-15
Keep main up to date with latest release


robwoodgate:
We discussed looking at automating as part of the release process.

PR #475, opened 2026-01-22
ci: create automatic release PR (dev->main) on release

Egge21M:
# Fixes: #468 

## Description

This ensures back-merges from development to main are not forgotten, by opening an automated PR from dev to main on release

## Changes

- Have release workflow create a temp branch from the release tag and open a PR to main

PR #481, opened 2026-01-30
chore: remove development branch and adopt main-only model

TheRealCheebs:
This should follow on top of the release automation and husky hooks #477 

chore: remove development branch and adopt main-only model
The `development` branch has been removed to simplify the branching
and release workflow. All new development should now be based
on and target the `main` branch.

- Updated documentation to reflect the `main`-only development model.
- Aligned CI/CD workflows and repository configuration to use `main` as the base branch.

This should help with the direction discussed in #468

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -68,7 +68,7 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           TAG="${{ inputs.tag || github.event.release.tag_name }}"
-          gh release upload "$TAG" \
+          gh release upload "$TAG" --clobber \
             "./lib/cashu-ts.iife.js" \
             "./lib/cashu-ts.iife.js.map" \
             "./lib/types/index.d.ts"
--- a/docs-src/DEVELOPER.md
+++ b/docs-src/DEVELOPER.md
@@ -123,11 +123,12 @@ Caution: `api:update` can modify generated files (e.g. API reports). Inspect and
 
 Many maintainers prefer to run the full PR checks locally before pushing. A common, reliable workflow:
 
-1. Start a local mint (for integration tests). An example docker-compose is available at `examples/auth_mint/docker-compose.yml`:
+1. Start a local mint (for integration tests). We have make targets for both CDK's mintd, and Nutshell:
 
 ```bash
 # from the repo root
-docker compose -f examples/auth_mint/docker-compose.yml up -d
+DEV=1 make cdk-up
+# or DEV=1 make nutshell-up
 ```
 
 1. Run the full PR tasks (lint, format, api:update, tests):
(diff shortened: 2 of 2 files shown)


TheRealCheebs:
@robwoodgate take another look at the version.yml. I pulled in the changes you mentioned in #475. Keeping in this file was the "smart" way, otherwise you can get into race conditions as there isn't a nice way to gate. But i added a manual kickoff on a specified tag so we can rebuild things if needed. 

I also had been looking at the wrong file for that duplicate permission issue which had been failing the format, so that's cool.

.github/workflows/version.yml
```diff
@@ -27,15 +41,34 @@ jobs:
...
       - run: npm run compile:standalone
+
       - name: Upload standalone bundle assets
+        if: ${{ inputs.tag || github.event.release.tag_name }}
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          gh release upload "${{ github.event.release.tag_name }}" \
+          TAG="${{ inputs.tag || github.event.release.tag_name }}"
+          gh release upload "$TAG" \
```
robwoodgate: Wondering if we should clobber (overwrite) the old files if re-run?

```suggestion
          gh release upload "$TAG" --clobber \
```

docs-src/DEVELOPER.md
```diff
@@ -134,13 +130,13 @@ Many maintainers prefer to run the full PR checks locally before pushing. A comm
 docker compose -f examples/auth_mint/docker-compose.yml up -d
```
robwoodgate: The auth mint is a bit technical for day to day - maybe just use the `make` mint.

```suggestion
make cdk-up
# or make nutshell-up
```

docs-src/DEVELOPER.md
```diff
@@ -165,7 +161,7 @@ The `test:consumer` aggregator runs the following scripts (you can run them indi
 
 Run the individual script if you want to isolate failures or speed up debugging.
 
-4. When finished, stop the local mint:
+1. When finished, stop the local mint:
 
 ```bash
 docker compose -f examples/auth_mint/docker-compose.yml down
```
robwoodgate: Ditto above:
```suggestion
make cdk-down
# or make nutshell-down
```

robwoodgate (commented):
Can't wait to see this workflow in action - seriously well done, and thank you @TheRealCheebs .

I've made a couple of suggestions covering mint up/down, and a question about whether we need to clobber. Otherwise LGTM.

docs-src/DEVELOPER.md
TheRealCheebs: I updated that section. We might want to rework it again soon as the integration tests bit is followed immediately by the makefile section and it flows wrong.

TheRealCheebs:
@robwoodgate Can you kick that node.js ci test again? It failed in the test prepare 
```
Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist [144 B]
Err:6 https://packages.microsoft.com/repos/azure-cli noble InRelease
  403  Forbidden [IP: 13.107.246.69 443]
```
hopefully that doesn't require an actual fix...

Also when this PR gets merged we should make sure to sqaush the commits.  That is the ideal flow, when possible, for release-please.

Issue #464, opened 2026-01-09
SIG_ALL - Message in P2PKWitness


robwoodgate:
https://github.com/cashubtc/nuts/pull/318

Experimental - can we usefully use this to facilitate multi-party signing?

This would be a wallet-led feature, so does not rely on mint changes, and can be flexible.

KvngMikey:
hi @robwoodgate , can we talk about this issue when you have the time ?

robwoodgate:
> * my question is what exactly should mts contain ? and won't it's data leak transaction data ? should we prioritize using the digest here if its available?

mts contains the concatenation of inputs (proofs) and outputs (amounts + blindedMessages) [as per NUT-11](https://github.com/cashubtc/nuts/blob/main/11.md#message-aggregation-for-sig_all)

> * when mts and digest exist must wallets confirm that SHA256(mts) == digest or is there a default we should pick at this point?

That was in the PR. But in new scheme, we would pass around a stripped down preview object with just the MTS materials.

> * are we enforcing the convention of embedding the `mts/digest` in the first proof alone to avoid ambiguity? that means if multiple proofs contain `mts/digest`, we reject and throw an error ?

In the new scheme, this would be a design decision. I suspect we might want to collate signatures in the object then inject them into first proof when merging back into the original Preview object, as that avoids complications.

> * i also noticed signP2PKProofs currently signs against three different message formats due to SIG_ALL being in flux any idea how we should handle this for or multi-party coordination ?

We can continue to sign all three - and remove the deprecated ones in v4. A mint doesn't care how many signatures, only that enough are valid for each signer expected.

PR #485, opened 2026-02-02
Sig_all signing package

KvngMikey:
# Fixes: #464 

## Description
Implements secure multi-party signing for P2PK SIG_ALL transactions (swap and melt)

## Changes

- Sanitized signing package from full `Proof[]` to minimal `{secret, C}[]` 
- **Multi-format digests**: Added `digests` field with legacy/current SHA256 hashes
  for backward compatibility with different mint versions
- Adds (de)serialilization to `sigallA` string format, and transforms between [Melt|Swap]Preview types.

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/model/SigAll.ts
+++ b/src/model/SigAll.ts
@@ -187,7 +187,10 @@ function deserializePackage(
 	if (options?.validateDigest) {
 		const recomputed = computeDigests(pkg.inputs, pkg.outputs, pkg.quote);
 		if (recomputed.current !== digests.current) {
-			throw new Error('Digest validation failed');
+			throw new Error('Digest validation failed: current digest mismatch');
+		}
+		if (digests.legacy && recomputed.legacy !== digests.legacy) {
+			throw new Error('Digest validation failed: legacy digest mismatch');
 		}
 	}
 


KvngMikey:
> @KvngMikey - I've only been able to go a high level look over so far, but it seems to be shaping up well, some really nice looking code here. Thank you!
> 
> The only major missing piece I can see is the "payload transport" format - we could use some explicit `serializeSigningPackage()` / `deserializeSigningPackage()` style helpers.
> 
> Also, perhaps we need to consider whether this should live in Wallet, or whether a separate utility class (like `OutputData`) would be more appropriate, as it's seems kind of self-contained, and ideally I would like to minimise the number of low level (noble) crypto utils that wallet imports.
> 
> Also, just FYI, as this is experimental, we won't be rushing it into production - but your work here will form the basis of the NUT specification.

@robwoodgate - thanks a lot for the review and the kind words, i really appreciate it.

I agree on the missing payload transport piece, with the first few commits of this PR i was mostly trying to get out a pow ensuring i'm on the right path and take it up from there.

also moving it to it's own utility class makes sense, i also have the type in the proof.ts of which i know it doesn't belong there so i'll have to move it too.

roger that on it being experimental, glad to be making solid contributions towards a NUT specification, that's why i'm here.

src/model/SigAll.ts
```diff
@@ -0,0 +1,439 @@
...
+ *
+ * @experimental
+ */
+export type SigAllSigningPackage = {
+	/**
+	 * Signing package version.
+	 */
+	version: 'cashu-sigall-v1';
+	/**
+	 * Typr of signing package.
```
robwoodgate:
```suggestion
	 * Type of signing package.
```

src/model/SigAll.ts
```diff
@@ -0,0 +1,439 @@
...
+	}
+
+	const digests = pkg.digests as Record<string, string> | undefined;
+	if (!digests || typeof digests.current !== 'string' || digests.current.length === 0) {
+		throw new Error('Signing package digests.current is required');
+	}
+
+	// Optional digest validation
+	if (options?.validateDigest) {
+		const proofLike = pkg.inputs.map((i) => ({ ...i, secret: '' })) as Proof[];
```
robwoodgate: Validate won't work, because the signing package doesn't have secret, and secret is required for both `buildP2PKSigAllMessage` and `buildLegacyP2PKSigAllMessage`

src/model/SigAll.ts
```diff
@@ -0,0 +1,439 @@
...
+	 */
+	quote?: string;
+	/**
+	 * Minimal inputs required for signing transport to prevent leaking sensitive data.
+	 */
+	inputs: Array<{
+		id: string;
+		amount: number;
+		C: string;
+	}>;
```
robwoodgate: `id` and `amount` is not required for SIG_ALL signing, but `secret` is.

Might also be better as a Pick

```suggestion
    inputs: Array<Pick<Proof, 'secret' | 'C'>>;
```

Will also change some of the logic below.

src/model/SigAll.ts
```diff
@@ -0,0 +1,439 @@
...
+	/**
+	 * Minimal outputs required for signing transport to prevent leaking sensitive data.
+	 */
+	outputs: Array<{ amount: number; blindedMessage: SerializedBlindedMessage }>;
```
robwoodgate: This is kind of duplicated - SerializedBlindedMessage contains `amount`.

Maybe:

```suggestion
    outputs: Array<Pick<OutputDataLike, 'blindedMessage'>>;
```

Will obviously have knock-on effects below.

robwoodgate (commented):
This is starting to come together, thank you @KvngMikey .

You’ve gone “privacy‑first” by stripping everything non‑essential, which I like. But without proof.secret the signing package can’t be verified — the SIG_ALL message includes the secret, so digest validation is impossible.

So we need to choose:  
 
- include proof.secret or full proof (verifiable package, more leakage), or  
- remove inputs (privacy, but package message can’t be self‑verified).

Do signers actually need to see the input proofs they’re spending, or only the output blinded messages and a digest? And does including secrets leak more privacy than the benefit of verifiability?

The coordinator keeps the original SwapPreview/MeltPreview and we rehydrate at the end, so that path is covered either way.

****EDIT:** I think we should include everything needed to verify for now, and strip it back later if there are privacy concerns. A signing package should probably be kept private between signers in any case, and input proofs are unspendable unless fully signed (and accompanied by full outputs, which we do NOT include in the signing package) so there is not early / race spend risk.**

@Egge21M - what do you think?

src/model/SigAll.ts
robwoodgate: If we resolve to keep input secrets in the package (to allow message_to_sign verification), then it's probably simpler to `Pick<>` the inputs/outputs in `buildP2PKSigAllMessage` and `buildLegacyP2PKSigAllMessage` rather than doing gymnsatics to make ProofLike objects

robwoodgate: This PR allows you to pass just the data required, rather than building full inputs/outputs.

https://github.com/cashubtc/cashu-ts/pull/518

robwoodgate (dismissed):

robwoodgate:
@KvngMikey - thinking ahead to the NUT proposal, I tweaked `SigAllSigningPackage.outputs` to `SerializedBlindedMessage[]`, which is the NUT-00 `BlindedMessages` array. 

So now we store (partial) `Proof`s for inputs and full NUT-00 `BlindMessage` for outputs. Both of which are NUT-00 compatible objects.

robwoodgate (approved):

KvngMikey:
> @KvngMikey - for the NUT proposal, have a think about how best to present this CONCISELY. My thoughts are:
> 
> At a protocol level, the NUTS need to define :
> 
> * The `SigAllSigningPackage` shape
> * A `sigallA,,,` example (to enable implementators to check conversion)
> 
> The rest is implementation detail that can be left out.
> 
> Think about where this will live. Either in NUT-00 as a core type, or in NUT-11 as a section. I don't have a strong opinion, both could be a natural fit. Though might lean towards NUT-11 to keep all the SigAll stuff together.
> 
> So if you are happy with the shape of this PR overall now, please can you draft the NUT PR and then we can tweak it together as needed.

okay, i'll come up with something for this...

als for point B `sigallA example`, i have this test that looks like when it outputs:
stdout | test/model/SigAll.test.ts > SigAll — serializePackage / deserializePackage > round-trip preserves all fields
[serialize output] sigallAeyJ2ZXJzaW9uIjoiY2FzaHUtc2lnYWxsLXYxIiwidHlwZSI6InN3YXAiLCJpbnB1dHMiOlt7InNlY3JldCI6InRlc3RzZWNyZXQiLCJDIjoiMDIxMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExIn1dLCJvdXRwdXRzIjpbeyJhbW91bnQiOjMyLCJpZCI6ImJtMSIsIkJfIjoiZHVtbXlCIn1dLCJkaWdlc3RzIjp7ImxlZ2FjeSI6ImQyYjBhODkxOTY0ZTIwM2VjZjI5YzBhYzQ4M2QzNzI3OWVjNzFiZmJiNzZhYjRjNDhiYTdkNDY0MDMxNDQ4ZGEiLCJjdXJyZW50IjoiYjJjMDkwYzgzNzdmYTE4MjljMjZhOTlkNjhkNmIzNmQwYTUxMzc0OWQzMDdlYzVkZDcwYTk0YzU0ZGI2OTdlMiJ9LCJ3aXRuZXNzIjp7InNpZ25hdHVyZXMiOlsic2lnMSJdfX0
stdout | test/model/SigAll.test.ts > SigAll — serializePackage / deserializePackage > round-trip preserves all fields
[deserialize output] {
  "version": "cashu-sigall-v1",
  "type": "swap",
  "inputs": [
    {
      "secret": "testsecret",
      "C": "021111111111111111111111111111111111111111111111111111111111111111"
    }
  ],
  "outputs": [
    {
      "amount": 32,
      "id": "bm1",
      "B_": "dummyB"
    }
  ],
  "digests": {
    "legacy": "d2b0a891964e203ecf29c0ac483d37279ec71bfbb76ab4c48ba7d464031448da",
    "current": "b2c090c8377fa1829c26a99d68d6b36d0a513749d307ec5dd70a94c54db697e2"
  },
  "witness": {
    "signatures": [
      "sig1"
    ]
  }
}

does this work ?

robwoodgate:
> i love these tweaks, also shortening the version string is actually best to match the prefix, although i don't think you've pushed that change yet ?

ooops. done now!

> als for point B sigallA example:

I'd probebly reuse the existing SIG_ALL example `SwapRequest` data, so it is based on something "real":

```json
{
  "inputs": [
    {
      "amount": 2,
      "id": "00bfa73302d12ffd",
      "secret": "[\"P2PK\",{\"nonce\":\"c7f280eb55c1e8564e03db06973e94bc9b666d9e1ca42ad278408fe625950303\",\"data\":\"030d8acedfe072c9fa449a1efe0817157403fbec460d8e79f957966056e5dd76c1\",\"tags\":[[\"sigflag\",\"SIG_ALL\"]]}]",
      "C": "02c97ee3d1db41cf0a3ddb601724be8711a032950811bf326f8219c50c4808d3cd",
      "witness": "{\"signatures\":[\"ce017ca25b1b97df2f72e4b49f69ac26a240ce14b3690a8fe619d41ccc42d3c1282e073f85acd36dc50011638906f35b56615f24e4d03e8effe8257f6a808538\"]}"
    }
  ],
  "outputs": [
    {
      "amount": 2,
      "id": "00bfa73302d12ffd",
      "B_": "038ec853d65ae1b79b5cdbc2774150b2cb288d6d26e12958a16fb33c32d9a86c39"
    }
  ]
}
```

turn that data into a `SwapPackage`... example:

The Example `SwapRequest` above would be represented as the following `SigAllSigningPackage`:

{....blah...}

The serialized base64url representation of this is:

sigallA....

robwoodgate (approved):

src/model/SigAll.ts
```diff
@@ -0,0 +1,435 @@
...
+	const digests = pkg.digests as Record<string, string> | undefined;
+	if (!digests || typeof digests.current !== 'string' || digests.current.length === 0) {
+		throw new Error('Signing package digests.current is required');
+	}
+
+	// Optional digest validation
+	if (options?.validateDigest) {
+		const recomputed = computeDigests(pkg.inputs, pkg.outputs, pkg.quote);
+		if (recomputed.current !== digests.current) {
+			throw new Error('Digest validation failed');
```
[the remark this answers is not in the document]
robwoodgate: @KvngMikey - can you add an if `digests.legacy`, check against `recomputed.legacy` guard please

Issue #482, opened 2026-01-30
[ BUG ] - Can not restore from inactive keysets


minibits-cash:
**Describe the bug**
Attempt to restore from inactive keyset fails with "Wallet keyset has no keys"

**To Reproduce**
Create Mint and Wallet instance, call wallet.loadMint and call wallet.restore with inactive keysetId.

**Expected behavior**
User can restore from inactive keyset.

**Environment (please complete the following information):**

- React native, but should not make diff
- cashu-ts 3.2.2 and 3.3.0

**Additional context**
Issue is that loadMint method filters out keys from inactive keysets when calling buildKeychain. Error is then thrown from finishInit.

robwoodgate:
Thanks for raising this @minibits-cash . I've prepared a fix #486 - if you are able to test it locally and confirm, that would be great. Will look to make a release later this week.

PR #486, opened 2026-02-02
fix: restore from inactive keyset

robwoodgate:
# Fixes: #482 

## Description

Wallet and Keychain were too strict, allowing only active keysets, which broke restore. This PR fixes restore and makes them more forgiving, including supporting "closing" mints which have no active keysets (restore / swap / melt only).

## Changes

- `KeyChain.ensureKeysetKeys()` allows lazy load of inactive keyset keys, with concurrent fetch protection
- KeyChain now loads all valid supplied keys, regardless of keyset status / version
- Removes smoke tests in KeyChain to allow inactive keychains
- Removes invalid keys instead of throwing
- `verifyKeysetId` returns false instead of throwing on invalid hex
- `Wallet.finishInit` no longer throws if keys not loaded or no active keysets

## PR Tasks

- [x] Open PR
- [x] run `npm run format`

### DIFF
--- a/src/wallet/KeyChain.ts
+++ b/src/wallet/KeyChain.ts
@@ -184,7 +184,7 @@ export class KeyChain {
 			const keyset = mk ? Keyset.fromMintApi(meta, mk) : Keyset.fromMintApi(meta);
 
 			// Discard unverifed keys
-			if (keyset.hasKeys && !keyset.verify()) {
+			if (!keyset.verify()) {
 				keyset.keys = {};
 			}
 


src/wallet/KeyChain.ts
```diff
@@ -188,19 +179,13 @@ export class KeyChain {
...
-			} else {
-				keyset = Keyset.fromMintApi(meta);
-			}
+			// Add keys if we have them
+			const mk = keysMap.get(meta.id);
+			const keyset = mk ? Keyset.fromMintApi(meta, mk) : Keyset.fromMintApi(meta);
 
-			// Validate active keysets with keys
+			// Discard unverifed keys
 			if (keyset.hasKeys && !keyset.verify()) {
```
TheRealCheebs: The first thing keyset.verify() does is check if this.hasKeys. We could drop that bit of double checking.

Issue #437, opened 2025-12-18
All amounts and key amounts need to be BigInt


Egge21M:
Right now cashu-ts assumes all amounts to be regular numbers. In JS numbers above 2^53-1 loose precision. As Nutshell and CDK both define their keyset key amounts as 0 - 2^64 - 1, we need to upgrade our amount handling to use BigInt in the future.

This might require quite a bit of change, as BigInt does not have great compatibility with other parts of the API:

- BigInt and regular numbers can not be mixed in operations ( `1 + 1n` -> throws)
- JSON.stringify does not support bigint conversion

I think the risk right now is limited as everything up to 2^53-1 works just fine and amounts larger than that are very unlikely, but just to be safe we could strip key amount larger than this and add assertions to enfore any amounts to be below that threshold

robwoodgate:
I've added a safeInteger check to the `assertAmount` method in Wallet until we figure out a longer term solution to this.

PR #514, opened 2026-02-18
feat(amount): add Amount value object (bigint support), AmountLike input and refactor internals

robwoodgate:
# Related: https://github.com/cashubtc/cashu-ts/issues/437

## Summary
We discussed introducing an Amount value object to make amount handling safer and to prepare for eventual u64/bigint support.

This PR adds and plumbs in `Amount` support through core wallet flows in a non-breaking way:

- public inputs move from number to `AmountLike` where applicable
- values are normalized early via `Amount.from(...)`
- internals use `Amount` in key paths
- public/wire outputs remain `number` for v3 api compatibility

## Detailed Changes

- Adds `Amount` VO, `AmountLike`, and `AmountError` (`src/model/Amount.ts`)
- Exports these to public API (`src/index.ts`)
- Replaces wallet `assertAmount`/`validateAmount` methods with `Amount`-based normalization
- Migrates wallet/core amount-taking APIs to `AmountLike` input while keeping output types stable
- Uses `Amount` internally across key wallet logic:
  - output configuration/splitting prep
  - totals/fee calculations
  - send/select paths
- Migrates `OutputData`/`BlindedMessage` to use `Amount` (converts to `number` at serialization edge)
- Migrates `WalletOps` amount entrypoints and denomination config to `AmountLike`
- Migrates `SelectProofs` target input to `AmountLike` with normalized target handling
- Widens `createBlindSignature` input to `AmountLike` (still emits numeric wire shape)
- Updates tests accordingly

## For v4, we should complete the migration by:
1. Switching public amount-bearing return types from `number` to `Amount` where appropriate.
2. Making domain models consistently hold `Amount` (not numeric primitives).
3. Re-evaluating JSON (wire serialization) strategy:
   - if [NUT PR#343](https://github.com/cashubtc/nuts/pull/343) is approved, move wire amounts to string/Amount-safe encoding.
   - if not, keep numeric JSON fields, but enforce explicit overflow/error policy at boundaries (fail, not precision loss)
4. Removing remaining legacy number-first assumptions once API break is allowed.

### DIFF
--- a/src/model/Amount.ts
+++ b/src/model/Amount.ts
@@ -107,14 +107,6 @@ export class Amount {
 	}
 
 	/**
-	 * Whether this Amount can be safely converted to a number.
-	 */
-	isSafeNumber(): boolean {
-		const maxSafe = BigInt(Number.MAX_SAFE_INTEGER);
-		return this.value <= maxSafe;
-	}
-
-	/**
 	 * Unsafe conversion to number. Precision can be lost above MAX_SAFE_INTEGER.
 	 */
 	toNumberUnsafe(): number {
@@ -176,6 +168,14 @@ export class Amount {
 	// -----------------------------------------------------------------
 	// Section: Comparison
 	// -----------------------------------------------------------------
+
+	/**
+	 * Whether this Amount can be safely converted to a number.
+	 */
+	isSafeNumber(): boolean {
+		const maxSafe = BigInt(Number.MAX_SAFE_INTEGER);
+		return this.value <= maxSafe;
+	}
 
 	isZero(): boolean {
 		return this.value === 0n;


src/wallet/Wallet.ts
```diff
@@ -1146,12 +1166,12 @@ class Wallet {
 		// Calculate our expected change from the swap (and sanity check!)
 		const selectedSum = sumProofs(selectedProofs);
 		const swapFee = this.getFeesForProofs(selectedProofs);
-		const changeAmount = selectedSum - swapFee - sendAmount;
-		this.failIf(changeAmount < 0, 'Not enough funds available for swap', {
+		const changeAmount = Amount.from(selectedSum).sub(swapFee).sub(sendAmount);
+		this.failIf(changeAmount.lt(0), 'Not enough funds available for swap', {
```
Egge21M: This guard is unreachable as `Amount.sub()` will throw if < 0

etc/cashu-ts.api.md
```diff
@@ -6,6 +6,57 @@
 
 import { WeierstrassPoint } from '@noble/curves/abstract/weierstrass.js';
 
+// @public
```
Egge21M, line 9: Should this be public or an internal class? As far as I understood the changes we do convert back to number for now at all consumer edges, so we could keep this internal, right?

Right now we do export it via AmountLike, so we would need to remove Amount from AmountLike and have an internal type that is `Amountlike | Amount`.

Regardless the comparison methods should probably either be documented or written out in full (e.g. lt -> lowerThan)

robwoodgate: Sorry, my original comment thought you were talking about the import.

I think this needs to be public - if not now then eventually - as consuming apps will need a way to process unsafe amounts, and this hands them the ability on a plate.

Plus, if we return `Amount` in v4, which we will have to for some lower level utility functions we export for convenience, it will need to be public API.

robwoodgate: Happy to adjust the comparisons naming - I went with math format for brevity, but `lessThan` etc makes the code self documenting, which is a win in my book!

src/wallet/Wallet.ts
robwoodgate: Great catch! Will tweak it.

src/model/OutputData.ts
```diff
@@ -49,7 +50,7 @@ export interface OutputDataLike<TKeyset extends HasKeysetKeys = HasKeysetKeys> {
  * removing the generic.
  */
 export type OutputDataFactory<TKeyset extends HasKeysetKeys = HasKeysetKeys> = (
-	amount: number,
+	amount: AmountLike,
```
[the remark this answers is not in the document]
robwoodgate: reverted. Will break in v4

src/wallet/SelectProofs.ts
```diff
@@ -3,10 +3,11 @@ import type { Proof } from '../model/types/proof';
 import { fail, failIf, failIfNullish, type Logger, NULL_LOGGER, measureTime } from '../logger';
 import { type SendResponse } from './types';
 import { type KeyChain } from './KeyChain';
+import { Amount, type AmountLike } from '../model/Amount';
 
 export type SelectProofs = (
 	proofs: Proof[],
-	amountToSend: number,
+	amountToSelect: AmountLike,
```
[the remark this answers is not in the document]
robwoodgate: Added overloads to provide dual support, with deprecation notice on number signature. 
This gives consumers a heads up and clean migration. We can then remove in v4.

PR #607, opened 2026-04-09
feat(nut29): wire NUT-06 max_batch_size into prepareBatchMint enforcement

KvngMikey:
## Description

Follow-up to #478 (NUT-29 batch minting). That PR implemented the batch mint flow but did not read or enforce the `max_batch_size` and `methods` fields that mints can advertise via NUT-06 mint info. This PR closes that gap.

---

## Changes

- `Nut29Info` added to `NUT29.ts`
- `GetInfoResponse.nuts['29']` added to `NUT06.ts`
- `isSupported(29)` overload returning `{ supported: boolean; params?: Nut29Info }`
- `normalizeNut29()` normalizes `max_batch_size`
- `prepareBatchMint` enforces `max_batch_size` before building outputs — fails fast with a message that includes both the attempted size and the advertised limit
- `prepareBatchMint` warns when the requested method is absent from the mint's advertised NUT-29 methods (string match only, no unit filtering — matches spec)
- `completeBatchMint` is unchanged — errors from the mint propagate with their original type and message intact
- All new checks are gated on `this._mintInfo?.isSupported(29)`. Mints that do not advertise NUT-29 info are completely unaffected.

---

## Alignment & Discussion

- Was an issue opened prior to this PR? [No]
- Does this PR align with the current project direction and scope? [Yes]

---

## Reviewer Notes

- `BATCH_SIZE_EXCEEDED` has no registered numeric error code in the NUT spec yet. The `prepareBatchMint` pre-flight check covers the common case. Once a formal error code is registered in the spec, a targeted catch on `e.code` can be added to `completeBatchMint` to handle mints that enforce the limit server-side without advertising it via NUT-06.

---

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -840,7 +840,7 @@ export type MintContactInfo = {
 
 // @public (undocumented)
 export class MintInfo {
-    constructor(info: GetInfoResponse);
+    constructor(info: GetInfoResponse, logger?: Logger);
     // (undocumented)
     get cache(): GetInfoResponse;
     // (undocumented)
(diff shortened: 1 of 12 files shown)


robwoodgate (approved):
@KvngMikey Awesome work, thank you! LGTM!

PR #608, opened 2026-04-10
refactor(wallet): inject output data creators while preserving OutputData helpers

Egge21M:
## Description

  This refactor extracts output-data construction into an OutputDataCreator abstraction and
  allows Wallet to accept a custom creator implementation at instantiation time.

  The existing OutputData.create*() static helpers are preserved for compatibility and now
  delegate to the default internal creator, so existing consumers do not need to change.

  ## What Changed

  - introduced OutputDataCreator as the interface for output-data construction
  - extracted the default implementation into DefaultOutputDataCreator
  - added outputDataCreator?: OutputDataCreator to the Wallet constructor
  - made wallet internals use the configured creator instead of hard-coded OutputData.create*()
    calls
  - preserved OutputData.create*() static methods as delegation-based compatibility helpers
  - ensured withKeyset() carries the configured creator into cloned wallets
  - exported OutputDataCreator from the public entrypoint

  ## Why

  This enables wallet-level customization of blinded output generation without breaking
  existing OutputData helper usage.

  Use cases include:

  - custom deterministic secret generation
  - alternative P2PK output construction strategies
  - injecting specialized output builders in advanced integrations

  ## Compatibility

  This keeps current OutputData static helper behavior intact.
  The public API expands with the new OutputDataCreator type and Wallet constructor option, but
  existing call sites should continue to work unchanged.

---

@robwoodgate I did keep the static mathod on OutputData intact on purpose, even though we are approaching a major release. I also decided not to export DefaultOutputDataCreator and defaultOutputDataCreator so that OutputData statics remain the idomatic way of OutputData creation. Please let me know if you want me to change that.


robwoodgate:
@codex review

robwoodgate (approved):
LGTM

PR #638, opened 2026-05-04
feat(nut12): deterministic DLEQ nonce derivation

robwoodgate:
Implements the [proposed change to NUT-12 spec](https://github.com/cashubtc/nuts/pull/368), switching createDLEQProof to derive `r` deterministically via HMAC-SHA256, keyed on mint's private key (`a`) and bound to the proof context (A, B', C'):

```
r = HMAC-SHA256(key=a, data="Cashu_DLEQ_R_v1" || A || B' || C' || ctr) mod n
```

This is not a breaking change. Verifiers reconstruct R1 and R2 from (e, s) — how the prover derived r is opaque to the verification equation. Existing proofs continue to verify.

## Changes

- `src/crypto/NUT12.ts` replaces `createRandomSecretKey()` with `deriveDLEQNonce()`. Function signature of createDLEQProof is unchanged.
- `test/crypto/NUT12.test.ts` — adds spec test vectors plus unit tests

### DIFF
--- a/src/crypto/NUT12.ts
+++ b/src/crypto/NUT12.ts
@@ -25,6 +25,7 @@ function deriveDLEQNonce(
     const r = bytesToNumberBE(h) % n;
     if (r !== 0n) return r;
   }
+  /* c8 ignore next */
   throw new Error('DLEQ nonce derivation failed');
 }
 
--- a/src/crypto/NUT12.ts
+++ b/src/crypto/NUT12.ts
@@ -1,10 +1,32 @@
 import { type WeierstrassPoint } from '@noble/curves/abstract/weierstrass.js';
 import { secp256k1 } from '@noble/curves/secp256k1.js';
-import { numberToBytesBE } from '@noble/curves/utils.js';
+import { bytesToNumberBE, numberToBytesBE } from '@noble/curves/utils.js';
+import { hmac } from '@noble/hashes/hmac.js';
+import { sha256 } from '@noble/hashes/sha2.js';
+import { concatBytes, utf8ToBytes } from '@noble/hashes/utils.js';
 
 import { Bytes } from '../utils';
 
-import { type DLEQ, hash_e, hashToCurve, createRandomSecretKey } from './core';
+import { type DLEQ, hash_e, hashToCurve } from './core';
+
+const DST_R = utf8ToBytes('Cashu_DLEQ_R_v1');
+
+function deriveDLEQNonce(
+  a: Uint8Array,
+  A: WeierstrassPoint<bigint>,
+  B_: WeierstrassPoint<bigint>,
+  C_: WeierstrassPoint<bigint>,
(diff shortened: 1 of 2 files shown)


robwoodgate:
@codex review

KvngMikey (approved):
ACK... 
although there's no test verifying a previous weakness: that reusing a nonce across two different messages would leak the key. Since this is the whole motivation for the PR, an explicit test documenting the threat model would be a nice addition.

robwoodgate:
> ACK... although there's no test verifying a previous weakness: that reusing a nonce across two different messages would leak the key. Since this is the whole motivation for the PR, an explicit test documenting the threat model would be a nice addition.

I have added the math to the NUT PR. There is no real point adding a permanent test to CTS as a) it means artificially regressing the fix in tests and b) just proves known math.

Here is a simplified TS proof of the issue of nonce reuse:

```typescript
import { invert, mod } from "@noble/curves/abstract/modular.js";
import { expect, test } from "vitest";

const n = BigInt(
  "0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141",
);

const a = BigInt(
  "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
);

const reusedNonce = BigInt(
  "0x4242424242424242424242424242424242424242424242424242424242424242",
);

const proof1 = {
  e: BigInt("0x1111111111111111111111111111111111111111111111111111111111111111"),
  s: BigInt("0xc49a2d11da2194415dfd54aa2a948adf95f5b1168422bb6386df3170690f28ca"),
};

const proof2 = {
  e: BigInt("0x3333333333333333333333333333333333333333333333333333333333333333"),
  s: BigInt("0xc94a02b109e0383f95737979fb391c1b82adb1d8589b0d6a5046b13fe672b499"),
};

test("reused DLEQ nonce reveals the mint private key", () => {
  // Prove s values were calculated using same private key
  expect(proof1.s).toBe(mod(reusedNonce + proof1.e * a, n));
  expect(proof2.s).toBe(mod(reusedNonce + proof2.e * a, n));

  const sDiff = mod(proof1.s - proof2.s, n);
  const eDiff = mod(proof1.e - proof2.e, n);
  const recoveredA = mod(sDiff * invert(eDiff, n), n);
  // Prove private key was recovered
  expect(recoveredA).toBe(a);
});
```

This is why the nonce must not repeat: the verifier only sees different `(e, s)` values, but if the prover reused `r`, the mint private scalar is recovered as `(s1 - s2) / (e1 - e2) mod n`.

a1denvalu3 (approved):
lgtm

robwoodgate:
Updated to latest version of spec - the Mint secret key is no longer hashed before being used as the HMAC-SHA256 key.

PR #655, opened 2026-05-09
feat(types): align nullable wire fields with spec (T | null)

robwoodgate:
## Summary

Harmonizes typing and runtime normalization for `<X|null>` wire fields that were previously typed as `T?` (undefined-flavored) or as required despite the spec admitting `null`. After this PR, the affected fields are spec-faithful `T | null` and consistently normalized at the response boundary.

**Affected fields:**
| NUT | Field | Before | After |
|---|---|---|---|
| 25 | `MintQuoteBolt12Response.amount` | `Amount?` | `Amount \| null` |
| 04/05/25/XX | `SwapMethod.min_amount` | `AmountLike` (required) | `AmountLike \| null` |
| 04/05/25/XX | `SwapMethod.max_amount` | `AmountLike` (required) | `AmountLike \| null` |

Each is normalized via `nullIfUndefined` at the response boundary, so mints that omit the fields entirely get coerced to spec-faithful `null` rather than leaking `undefined`. Also removes the now-unused internal `undefinedIfNull` helper — its sole call site was the old bolt12 amount path.

## Why now

Two recent PRs (#653 for `payment_preimage` / `witness`, and the `maxSpendableAfterFees` work in #654) leaned on the existing `nullIfUndefined` helper. After a NUTs-spec audit, these three fields were the remaining outliers — they're the only public types that lie about nullability relative to their spec definitions. The wider keyset metadata fields (`MintKeys.input_fee_ppk`, `MintKeyset.final_expiry`, etc.) stay typed `T?` intentionally — consumers truthy-check / `??` them and there's no observed breakage, so harmonizing here has a real cost (Keyset wrapper, KeyChainCache shape, persistence) and no protocol-meaning payoff, unlike the bolt12 case.

## Compatibility

Shipping as a minor (no `!` / `BREAKING CHANGE:` marker) — the actual blast radius is small and conventional consumer patterns are unaffected.

**Untouched** (still works after this PR):
- `if (x)` truthy checks
- `x ?? fallback`
- `x?.foo` optional chains
- `x != null` checks (covers both null and undefined)

**Will need a one-line consumer update**:
- TS code comparing `q.amount === undefined` on a bolt12 mint quote response — switch to `q.amount === null` or `q.amount == null`
- Destructuring `min_amount` / `max_amount` with a default value (`const { min_amount = 0n } = m`) — defaults no longer trigger on null, switch to `m.min_amount ?? 0n`
- Variables typed as `T?` reassigned from these fields — TS will require widening

**Silent runtime change**:
- `JSON.stringify(quote)` now emits `"amount": null` instead of omitting the field for amountless bolt12 quotes. Could matter for round-trip storage; won't matter for typical wallet UX.

Bolt12 adoption is still early (Nutshell + CDK on the mint side; only a handful of wallets ship bolt12 melts), so the practical downstream impact is limited even for `q.amount` — the highest-risk field by far. `SwapMethod.min_amount`/`max_amount` are read by no internal cashu-ts code; only consumers feeding them into `Amount.from(...)` would have noticed the type lie.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] Two `MintQuoteBolt12` normalization tests in `test/mint/Mint.node.test.ts` updated (`toBeUndefined()` → `toBeNull()`); names tightened to describe the new contract
- [x] Wallet-init test now compares `info.cache` to `MintInfo.normalizeInfo(mintInfoResp)` rather than the raw wire payload — that assertion was passing by accident before normalization landed for these fields
- [x] Wallet-init `isSupported(5)` assertion explicitly lists `min_amount: null, max_amount: null` per method, surfacing the new normalized shape
- [x] Full unit/MSW suite (`npx vitest run test/mint/ test/wallet/ test/model/ test/utils/`) — 2569/2569 pass
- [x] API report (`etc/cashu-ts.api.md`) regenerated

## Related
- #653 added `nullIfUndefined` for `payment_preimage` / `witness` — same Postel-style shape
- #654 added `maxSpendableAfterFees`; touches the same fee-primitive area but no overlap

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1046,7 +1046,7 @@ export type MintQuoteBolt12Request = MintQuoteBaseRequest & {
 
 // @public
 export type MintQuoteBolt12Response = MintQuoteBaseResponse & {
-    amount?: Amount;
+    amount: Amount | null;
     expiry: number | null;
     pubkey: string;
     amount_paid: Amount;
@@ -1764,8 +1764,8 @@ export function sumProofs(proofs: Array<Pick<ProofLike, 'amount'>>): Amount;
 export type SwapMethod = {
     method: string;
     unit: string;
-    min_amount: AmountLike;
-    max_amount: AmountLike;
+    min_amount: AmountLike | null;
+    max_amount: AmountLike | null;
     description?: boolean;
     options?: {
         description?: boolean;
--- a/src/mint/Mint.ts
+++ b/src/mint/Mint.ts
@@ -47,7 +47,7 @@ import {
   normalizeMintKeyset,
   normalizeSafeIntegerMetadata,
   normalizeUrl,
-  undefinedIfNull,
+  nullIfUndefined,
 } from '../utils';
 
 import type {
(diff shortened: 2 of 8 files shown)


robwoodgate:
@codex review

PR #658, opened 2026-05-11
refactor: optimize deterministic derivation

robwoodgate:
## Summary

- add `deriveSecretAndBlindingFactor` - a paired NUT-13 derivation API for deriving deterministic secrets and blinding factors together.
- deprecated the single derive functions: `deriveSecret`, `deriveBlindingFactor`
- cache BIP32 master derivation for deterministic batch output creation, speeding up legacy recovery paths
- preserve `DefaultOutputDataCreator` subclass behavior when `createSingleDeterministicData` is overridden
- add focused coverage for the extender fallback and OutputData helper branches

## Notes

Local benchmark for 1000 legacy counters showed roughly 2x speedup for the cached batch path versus repeated BIP32 derivation of previous implementation.

### DIFF
--- a/src/crypto/NUT13.ts
+++ b/src/crypto/NUT13.ts
@@ -133,12 +133,12 @@ function deriveHmacSecretAndBlindingFactor(
   counter: number,
 ): DerivedSecretAndBlindingFactor {
   return {
-    secret: derive(seed, keysetId, counter, DerivationType.SECRET),
-    blindingFactor: derive(seed, keysetId, counter, DerivationType.BLINDING_FACTOR),
+    secret: deriveHmac(seed, keysetId, counter, DerivationType.SECRET),
+    blindingFactor: deriveHmac(seed, keysetId, counter, DerivationType.BLINDING_FACTOR),
   };
 }
 
-function derive(
+function deriveHmac(
   seed: Uint8Array,
   keysetId: string,
   counter: number,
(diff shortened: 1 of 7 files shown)


robwoodgate:
@codex review

robwoodgate:
@Egge21M - I took your excellent initial optimization (caching the expensive BIP32 setup per restore batch) and expanded it to cover all deprecated BIP32 keyset IDs. I also updated DefaultOutputDataCreator to take advantage of the batch optimization, and bumped default restore batching to reduce network round trips. The rest was mostly API cleanup, tests, etc.

src/crypto/NUT13.ts
```diff
@@ -17,44 +17,133 @@ enum DerivationType {
...
+
+/**
+ * Creates a deterministic deriver function for a seed/keyset pair.
+ *
+ * @remarks
+ * For deprecated BIP-32 derivation this caches the master key once, so callers that derive many
+ * counters can reuse the expensive seed setup.
+ * @internal
+ */
+export function createSecretAndBlindingFactorDeriver(
```
Egge21M, line 85: nit: This is marked as internal, but because the whole NUT13 barrel is exported it will be part of the public API

Egge21M (approved):
LGTM! Thank you for picking this up. Only thing I noticed was an export of an internal helper

src/crypto/NUT13.ts
robwoodgate: Thank you sir!

It's exported because it's used in `OutputData`. It's marked internal so it doesn't appear in the official public API (not in API Extractor report). Could technically be deep linked, but that's "buyer beware"

PR #662, opened 2026-05-15
feat(model): add AmountWithUnit value object

robwoodgate:
## Summary

Adds `AmountWithUnit`, a unit-aware sibling to `Amount`, so multi-unit consumers (wallets aggregating across `sat`/`usd` or multiple mints) can do arithmetic and comparisons without silently mixing units. Binary ops require matching units and throw `AmountWithUnitError`; scalar ops preserve the unit.

`AmountWithUnit` and `Amount` are deliberately distinct types with no implicit coercion: lift with `Amount.withUnit()` or `AmountWithUnit.from()`, drop back with `.toAmount()`.

**Design Note:** Extending `AmountLike` to accept `AmountWithUnit` was rejected: that would silently strip the unit at every coercion point. CTS APIs continue to take `AmountLike ` and return `Amount`; `AmountWithUnit` is opt-in at the consumer boundary.

## Detailed Changes

- new `AmountWithUnit` class in `src/model/Amount.ts`: passes through Amount methods with unit checks as needed
- new `AmountWithUnitError extends CTSError`
- new `Amount.withUnit(unit)` lifting method
- updated documentation and migration guides
- tests bring `src/model/Amount.ts` to 100%

## Notes

No internal call sites migrated: `Wallet` / `Mint` / `SelectProofs` / `utils/core` continue using unit-agnostic `Amount`.

Public API additions:

- new exports: `AmountWithUnit`, `AmountWithUnitError`
- new method on `Amount`: `withUnit(unit: string): AmountWithUnit`

### DIFF
--- a/docs-src/usage/amounts.md
+++ b/docs-src/usage/amounts.md
@@ -56,8 +56,8 @@ AmountWithUnit.sum([a, b]); // 150 sat (unit inferred)
 AmountWithUnit.sum([], 'sat'); // 0 sat (empty + explicit hint)
 
 // Display: string coercion is unit-bearing (and parseInt-safe)
-`balance: ${a}`; // 'balance: sat: 100'
-String(a); // 'sat: 100'
+`balance: ${a}`; // 'balance: [sat]: 100'
+String(a); // '[sat]: 100'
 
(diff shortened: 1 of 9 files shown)


robwoodgate:
@codex review please

PR #665, opened 2026-05-21
fix: use generic NUT-17 subscription kinds

callebtc:
## Summary
- switch wallet websocket subscriptions to the new generic `mint_quote` and `melt_quote` kinds
- keep the old bolt11-specific kinds in the public RPC type as deprecated aliases for backwards compatibility
- update NUT-17 test fixtures and websocket tests to advertise and expect the new command strings

## Testing
- npx vitest run --project node test/transport/WSConnection.test.ts test/wallet/WalletEvents.test.ts test/wallet/wallet.node.test.ts
- npx vitest run --project node test/wallet/bolt12.test.ts test/transport/request.node.test.ts

Refs: cashubtc/nuts#372


robwoodgate:
Closing in favour of #668, which is exactly the same patch, just applied cleanly to current state of main.

The PR above was based on a very old state of main, making rebase more difficult than a new PR.

PR #682, opened 2026-05-24
Add conditional token primitives for NUT-CTF

joemphilips:
## Summary

Draft implementation of client-side primitives for the conditional token / prediction-market NUT proposal in https://github.com/cashubtc/nuts/pull/337.

This is intentionally opened as a draft because the NUT text is still under review and the API shape may need to change with maintainer feedback. The branch is based on current upstream `main` and excludes generated `lib/` output.

## What Is Included

- CTF condition-id and conditional-keyset helpers
- typed mint payloads/responses for conditional keyset discovery, condition registration, split/merge, and outcome redemption
- `Mint` methods for the proposed CTF endpoints
- optional `wallet.ctf` facade behind `new Wallet(mint, { enableCtf: true })`
- wallet/keychain validation that keeps conditional swaps within one conditional keyset and rejects mixed regular/conditional inputs
- node tests for CTF crypto helpers, mint payload handling, conditional keychain lookup, conditional swap, and redemption flows

## Review Notes

The implementation is meant to exercise the shape proposed in cashubtc/nuts#337 and provide concrete feedback from a TypeScript wallet implementation. In particular, it keeps conditional keysets out of regular NUT-02 keyset selection and treats condition-derived keysets as an explicit registry loaded through CTF endpoints.

This PR does not make CTF behavior default wallet behavior. Callers must opt into the facade with `enableCtf: true`.

Related reference implementation context from the NUT PR: https://github.com/cashubtc/cdk/pull/1666

## Validation

- `npm run compile`
- `npx vitest run --project node test/crypto/CTF.test.ts test/mint/Mint.node.test.ts test/wallet/keychain.node.test.ts test/wallet/wallet-ctf.node.test.ts`

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -278,6 +278,75 @@ export function computeMessageDigest(message: string, asHex: false): Uint8Array;
 // @public (undocumented)
 export function computeMessageDigest(message: string, asHex: true): string;
 
+// @public (undocumented)
+export interface ConditionalKeysetInfo {
+    // (undocumented)
+    active: boolean;
+    // (undocumented)
+    condition_id: string;
+    // (undocumented)
+    final_expiry?: number;
+    // (undocumented)
+    id: string;
+    // (undocumented)
+    input_fee_ppk?: number;
+    // (undocumented)
+    outcome_collection: string;
+    // (undocumented)
+    outcome_collection_id: string;
+    // (undocumented)
+    registered_at?: number;
+    // (undocumented)
+    unit: string;
+}
+
+// @public (undocumented)
+export interface ConditionalKeysetMetadata {
+    conditionId: string;
+    outcomeCollection: string;
+    outcomeCollectionId: string;
+    registeredAt?: number;
+}
+
+// @public (undocumented)
+export interface ConditionalKeysetsResponse {
+    // (undocumented)
+    keysets: ConditionalKeysetInfo[];
+}
+
+// @public (undocumented)
+export interface ConditionalSwapOptions {
+    // (undocumented)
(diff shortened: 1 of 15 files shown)


robwoodgate:
Thanks for opening this as a draft. It's an interesting concept.

As requested, I’m treating this draft as a prototype/reference implementation, not something ready for full maintainer review.

Even if the NUT progresses to acceptance, there are a few high-level blockers before this could become reviewable - flagging now to make it easier for you later.

## Scope

This PR adds a large amount of public Mint/Wallet/KeyChain surface for a very niche application area.

Even with the `wallet.ctf` facade gated behind `enableCtf`, the PR still adds public `Wallet` methods, `Mint` methods, keyset metadata, and keychain behavior that every other consumer would have to carry.

Please keep CTF behind a much narrower experimental boundary, preferably without expanding core `Wallet`/`KeyChain` public surface.

See the `auth` module for the kind of encapsulation I mean.


## Compat overloads in Wallet

The compatibility overloads for the old v3-style methods are also unacceptable. We moved to the generic method-based APIs in v4, and overloading those same names to also behave like the old BOLT11 convenience methods is a step backwards.

It makes the API ambiguous, especially because `AmountLike` includes `string`, and it blurs the distinction between generic payment-method flows and the explicit first-class helpers like `createMintQuoteBolt11()` / `mintProofsBolt11()`.

Please remove the v3 compatibility overloads entirely.

## API

The public API needs looking at - the API Extractor report shows a bunch of forgotten exports and an explosion of changes in the public API.

## Docs & Tests

Finally, would need to add docs and complete test coverage when the API shape is intentional and stable enough to expose.

PR #685, opened 2026-06-01
fix(nut-17): replace Math.random() subId with CSPRNG-backed UUID v7

KvngMikey:
## Description

Implements [cashubtc/nuts#383](https://github.com/cashubtc/nuts/pull/383).

The NUT spec now recommends UUID v7 for quote IDs (NUT-04, NUT-05) and WebSocket subscription IDs (NUT-17).  ([robwoodgate](https://github.com/robwoodgate)) clarified that all 74 variable bits must be CSPRNG-generated, not filled with a monotonic counter or predictable suffix.

---

## Changes

- no changes needed on NUT-04 / NUT-05, the quote IDs are assigned by the mint and returned to the client. The wallet never generates them.
- fixed, NUT-17, the `WSConnection.createSubscription()` was generating `subId` with a weak idiom:
```ts
// before: ~30 bits of entropy, Math.random() is not cryptographically secure
const subId = (Math.random() + 1).toString(36).substring(7);
```
Replaced with a new `generateUuidV7()` helper:
```ts
// after
const subId = generateUuidV7();
```

---

## Alignment & Discussion

- Was an issue opened prior to this PR? [Spec]

---

### DIFF
--- a/src/utils/uuid.ts
+++ b/src/utils/uuid.ts
@@ -28,10 +28,10 @@ export function generateUuidV7(): string {
   buf[5] = Number(tsMs & 0xffn);
 
   // Byte 6: version nibble (0x7) in the high nibble, keep low nibble random (rand_a[0:4])
-  buf[6] = (0x70 | (buf[6] & 0x0f)) >>> 0;
+  buf[6] = 0x70 | (buf[6] & 0x0f);
 
   // Byte 8: variant bits 0b10 in the two high bits, keep lower 6 bits random (rand_b[0:6])
-  buf[8] = (0x80 | (buf[8] & 0x3f)) >>> 0;
+  buf[8] = 0x80 | (buf[8] & 0x3f);
 
   const hex = Array.from(buf, (b) => b.toString(16).padStart(2, '0')).join('');
   return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
--- a/src/utils/uuid.ts
+++ b/src/utils/uuid.ts
@@ -1,4 +1,4 @@
-import { randomBytes } from '@noble/hashes/utils.js';
+import { bytesToHex, randomBytes } from '@noble/hashes/utils.js';
 
 /**
  * Generates a UUID v7 string.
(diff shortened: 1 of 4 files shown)


robwoodgate:
Great work, tidy implementation - thanks @KvngMikey 

Please can you add a `test/utils/uuid.test.ts` file with the following tests to assert the version/variant bit and ensure the timestamp encodes into bytes 0–5 properly. That ensures the UUIDv7 invariants:

```ts
import { test, describe, expect } from 'vitest';
import { generateUuidV7 } from '../../src/utils/uuid';

describe('generateUuidV7', () => {
  test('has version 7 and variant 10 bits', () => {
    const re = /^[0-9a-f]{8}-[0-9a-f]{4}-7[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
    for (let i = 0; i < 100; i++) {
      expect(generateUuidV7()).toMatch(re);
    }
  });

  test('encodes Date.now() into bytes 0-5 in big-endian order', () => {
    const before = Date.now();
    const u = generateUuidV7();
    const after = Date.now();
    const tsHex = u.slice(0, 8) + u.slice(9, 13);
    const ts = Number('0x' + tsHex);
    expect(ts).toBeGreaterThanOrEqual(before);
    expect(ts).toBeLessThanOrEqual(after);
  });
});
```

Aside from that, LGTM when the NUT is signed off. Thanks!

KvngMikey:
you got it ! @robwoodgate

src/utils/uuid.ts
```diff
@@ -0,0 +1,38 @@
...
+  buf[4] = Number((tsMs >> 8n) & 0xffn);
+  buf[5] = Number(tsMs & 0xffn);
+
+  // Byte 6: version nibble (0x7) in the high nibble, keep low nibble random (rand_a[0:4])
+  buf[6] = 0x70 | (buf[6] & 0x0f);
+
+  // Byte 8: variant bits 0b10 in the two high bits, keep lower 6 bits random (rand_b[0:6])
+  buf[8] = 0x80 | (buf[8] & 0x3f);
+
+  const hex = Array.from(buf, (b) => b.toString(16).padStart(2, '0')).join('');
```
robwoodgate: Nit: we should use noble's `bytesToHex` instead of rolling our own hex conversion here. It's also about 1.5x faster as it uses a precomputed hex lookup table internally.

robwoodgate (approved):
LGTM

PR #742, opened 2026-07-02
fix(utils): throw CTSError for malformed token templates

robwoodgate:
## Problem

`tokenFromTemplate` casts decoded CBOR to `TokenV4Template` with an unchecked cast, then dereferences `template.t.forEach(...)` with no shape validation. CBOR can encode any structure, so a token whose body is well-formed CBOR but not a v4 template (e.g. a map `{ m, u }` with no `t` array) makes `template.t` undefined, and `undefined.forEach` throws a raw `TypeError` instead of the library's `CTSError`.

Every other failure on this path (unsupported version, multi-entry token, short-keyset resolution) throws `CTSError`, so callers that decode attacker-supplied tokens and branch on `e instanceof CTSError` get an unexpected exception type. Reachable from the public `getDecodedToken`, `getTokenMetadata`, and `getDecodedTokenBinary`.

This is a robustness/input-validation issue only (CWE-20). No privilege escalation.

## Fix

Guard the template shape in the shared `tokenFromTemplate`, so both the `cashuB` and `crawB` decode paths are covered by a single change. A malformed template now throws `CTSError('Invalid token template')`.

## Tests

Added regression tests for both entry points (`getDecodedToken` / `getDecodedTokenBinary`) asserting `CTSError` rather than a raw `TypeError`. `npm run prtasks` passes.

---

This finding was discovered by Project Loupe.

### DIFF
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -302,8 +302,14 @@ function templateFromToken(token: Token): TokenV4Template {
 }
 
 function tokenFromTemplate(template: TokenV4Template): Token {
+  if (!template || !Array.isArray(template.t)) {
+    throw new CTSError('Invalid token template');
+  }
   const proofs: Proof[] = [];
-  template.t.forEach((t) =>
+  template.t.forEach((t) => {
+    if (!t || !Array.isArray(t.p)) {
+      throw new CTSError('Invalid token template');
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #743

PR #783, opened 2026-07-13
fix(wallet): widen config type on bolt12/onchain mint helpers

robwoodgate:
## Summary

`mintProofsBolt12()` and `mintProofsOnchain()` typed their config as an inline `{ keysetId?: string }`. That literal was written when `MintProofsConfig` was just `{ keysetId, privkey }` (privkey moved to a positional param, since bolt12/onchain quotes are always locked), so at the time it was the full config minus privkey. When `proofsWeHave` and `onCountersReserved` were later added to `MintProofsConfig`, the inline type fell behind: `prepareMint()` honours both at runtime, but the typed helpers rejected them, forcing callers to cast or drop to the builder API.

Both helpers now take `config?: Omit<MintProofsConfig, 'privkey'>`, so they track the shared config type from now on. Pure widening, no behaviour change.

## Changes

- `src/wallet/Wallet.ts`: widen the config parameter on both helpers.
- `test/wallet/wallet-mint.node.test.ts`: the onchain mint test now uses a seeded wallet and asserts `onCountersReserved` fires through the typed helper.
- `etc/cashu-ts.api.md`: regenerated.

## Notes

Reported by a consumer integrating v4.7, who hit the cast on `mintProofsOnchain`. Labelled for v4-dev backport. 

`npm run prtasks` passes.


robwoodgate:
- #784

PR #794, opened 2026-07-15
feat(wallet): batched probe search for NUT-342 recovery

robwoodgate:
## NUT

- https://github.com/cashubtc/nuts/pull/342

Draft spec experiment, stacked on #788

## Description

Speeds up NUT-342 recovery by changing how the wallet searches for its last used counter. The previous search (#788) asked the mint one question at a time, roughly 30 questions in a row using binary search.

This PR asks the same questions in bundles, using an exponential ladder, followed by a grid and tile, so a full recovery takes 4-8 requests at any wallet age.

## Why a ladder rather than pure binary search

Counters live in a vast space (2^32 positions), but real wallets cluster at the very bottom of it: every wallet starts at counter 0 and climbs only as it is used, so the last counter is typically in the hundreds or low thousands, and even a million would be remarkable.

Pure binary search ignores that shape. It treats every position as equally likely and spends the same ~31 questions whether the wallet is a day or a decade old. That purity has a real virtue: most of its probes land above T in never-used territory, where they reveal nothing, making it the most private search (see the Linked column below).

The **exponential *ladder*** instead bets on the actual distribution: it starts where wallets really are and doubles outward, so the cost of the search follows the wallet's true age rather than the size of the theoretical space. The bet trades a little privacy for a large win in round trips: its probes land in used territory more often, so the mint can link a few more old signatures. `probeBudget` tunes where between those two extremes the search sits.

## How it works, step by step:

1. **Bracket the wallet's high point.** The wallet asks the mint about a 25-counter window (`probeWindow`) at each of a spread of positions, all in one request: [0-24], [25-49], [50-74], [100-124], [200-224] and so on, the start doubling each time (in the code: the **exponential *ladder***). Positions the wallet once used come back with signatures; positions it never reached come back empty. The highest position with a signature and the lowest empty one above it fence in the wallet's true last counter, T. (An optional `ladderSkip` drops the lowest rungs, which aged wallets have always used; a keyset living below the first remaining rung triggers one extra request that fires the skipped rungs after all. The default is no skip.)
2. **Narrow the fence.** That fence can still be wide. The next request spreads twelve small probes evenly across it, shrinking the unknown range by roughly 13x (a ***grid* round**; width set by `probeBudget`). One or two of these rounds is enough for any realistic wallet.
3. **Find T exactly.** Once the remaining range is small, one request checks every counter in it (**the *tile* round**). That lands on T precisely, and the mint's reply for T already carries the encrypted recovery gap (`d_gap`), so nothing extra is fetched.
4. **Restore the window.** All unspent proofs are guaranteed to sit within `d_gap` counters below T. The wallet fetches that window in chunks of 500 (matching `batchRestore`), several chunks in flight at once (via the request pool from #789), then runs the usual NUT-07 state check.

Nothing changes on the wire or at the mint: NUT-09 already accepts any set of blinded messages per request, so this is compatible with the draft spec and with the Nutshell reference implementation as they stand. The aim is to prove the approach in code and then propose it as the recommended search in the cashubtc/nuts#342 discussion.

## Benchmark

`scripts/bench-nut342-recovery.ts` mirrors the Nutshell reference benchmark setup (probe batch 25, `d_gap`=100, one skipped counter every 137 derivations) against an in-process mock mint with 50 ms simulated network latency. Calls include the NUT-07 state check; "this PR" rows use the defaults (`probeBudget: 12`, `ladderSkip: 0`). "Linked" counts the previously issued signatures the mint can associate with the recovery session - the privacy cost; the final `d_gap + 1` recovery window is unavoidable for any strategy.

| Method            | T         | Calls | Messages | Linked | Proofs checked | Wall  |
| ----------------- | --------- | ----- | -------- | ------ | -------------- | ----- |
| NUT-09 scan       | 100       | 5     | 2,000    | 101    | 101            | 0.8s  |
| bisection (#788)  | 100       | 32    | 851      | 101    | 101            | 2.4s  |
| this PR           | 100       | 4     | 925      | 101    | 101            | 0.7s  |
| NUT-09 scan       | 1,000     | 6     | 2,000    | 994    | 994            | 2.3s  |
| bisection (#788)  | 1,000     | 31    | 826      | 125    | 100            | 2.4s  |
| this PR           | 1,000     | 5     | 1,160    | 289    | 100            | 0.8s  |
| NUT-09 scan       | 10,000    | 44    | 12,000   | 9,928  | 9,928          | 20.7s |
| bisection (#788)  | 10,000    | 31    | 826      | 201    | 101            | 2.4s  |
| this PR           | 10,000    | 6     | 1,451    | 523    | 101            | 0.9s  |
| NUT-09 scan       | 100,000   | 403   | 102,000  | 99,272 | 99,272         | 210.5s |
| bisection (#788)  | 100,000   | 32    | 851      | 226    | 101            | 2.5s  |
| this PR           | 100,000   | 6     | 1,718    | 821    | 101            | 1.0s  |
| this PR           | 1,000,000 | 8     | 2,028    | 1,048  | 100            | 1.3s  |

Scan and "this PR" rows are measured on this branch as it stands (the pooled 500-counter `batchRestore` from #795, 500-counter window chunks): the scan's call count benefits from #795's waves, at the price of some speculative overshoot in messages. The bisection rows predate that rebase and are kept as measured: they reproduce the Nutshell reference benchmark exactly (32/851, 31/826, 31/826, 32/851), independently validating the harness and confirming the two implementations behave identically.

Long-run trends: this PR's search adds one narrowing round per ~13x of wallet age, so requests cap at ~10 even at the counter-space maximum (the one-at-a-time search stays at 31-32). Linked signatures grow slowly with age and cap around ~2,000, versus the scan linking everything.

## d_gap is wallet policy

The search costs ~4 requests at any wallet age; the recovery window costs `d_gap`, and the wallet controls it. Three live runs of the bundled demo against the dockerized Nutshell mint (`input_fee_ppk` 100; each regime is an env knob on the demo) differ only in wallet policy:

| Wallet history | Method | Restore requests | Messages | Linked | Wall |
| --- | --- | --- | --- | --- | --- |
| unmanaged, small spends (T=4,409, d_gap = T) | this PR | 13 | 6,155 | 4,409 | 32s |
| | NUT-09 scan | 12 | 6,000 | 4,409 | 33s |
| pending token sent mid-history, refresh on (T=16,176, d_gap=8,489) | this PR | 22 | 10,377 | 8,739 | 60s |
| | NUT-09 scan | 36 | 18,000 | 16,176 | 115s |
| oldest-refresh policy (T=16,285, d_gap=135) | this PR | 5 | 1,510 | 586 | 1.8s |
| | NUT-09 scan | 36 | 18,000 | 16,285 | 114s |

(The 13 and 22 each include one extra request from a retried dropped connection.)

**Lump-sum funding plus small spends maximizes `d_gap`.** The window is set by the wallet's single oldest live proof, and nothing about normal usage moves it. Small spends never break a lump-funded wallet's large reserve proofs: in the first row the counter-0 reserve survived every round, the window is the whole history, and the search sits at its floor: scan cost plus three search requests.

**Unclaimed sends pin the window** at their own counter even while consolidation runs (second row: the refresh policy was active throughout), and dust that never clears its own fee pins it the same way.

**Randomized coin selection keeps old proofs in play indefinitely.** Only heavy turnover consolidates by accident (a run swapping ~half the balance per round landed at d_gap 172 with no policy at all).

**A cheap periodic refresh bounds `d_gap`** by the refresh cadence (the demo rolls its oldest proofs forward every 20 swaps): one extra swap per cycle, a few counters (third row).

Wallets should actively manage the window by periodically refreshing their oldest live proofs (folding dust in), rather than relying on spending patterns; long-unclaimed sends set a floor no wallet policy can cross. Proposed as wallet guidance in the cashubtc/nuts#342 discussion.

## Privacy comparison on a real wallet

The "binary search is more private" assertion, measured. Recovery runs per keyset, so a wallet pays the search cost once for every keyset it has ever used. This scenario replays the nine keysets of an actual daily-use wallet (last counters: 2024, 728, 69, 53, 19, 4, 4, 2, 2) through the same harness.

| Strategy | Requests | Linked signatures | Wall @50ms |
| --- | --- | --- | --- |
| bisection (#788) | 283 | 538 | 21.7s |
| this PR, defaults | 45 | 909 | 6.2s |
| this PR, privacy profile (`probeBudget: 4, ladderSkip: 3`) | 53 | 558 | 6.6s |

Three facts the totals hide:

- On seven of the nine keysets, every strategy links the identical signatures: those keysets' whole history sits inside the `d_gap` recovery window, which any strategy must fetch. There is no privacy to win on young keysets; binary search still pays ~31 requests each to tie.
- On the most-used keyset (T=2024) the privacy profile links fewer signatures than binary search (199 vs 225): skipping the low rungs (`ladderSkip`) removes exactly the probes that always land in an aged wallet's spent history.
- The wallet-wide difference comes to 20 signatures (558 vs 538), concentrated on one mid-size keyset. That is what binary search's 230 extra requests and 15 extra seconds buy.

`ladderSkip` is two-stage: a keyset living below the first rung triggers one extra request that fires the skipped rungs, so young keysets are never missed and cost a single extra round trip.

## Changes

- `restoreFromGapBackup` rewritten as the bundled search described above, with a 16-round safety cap and the existing linear NUT-09 scan fallback untouched.
- New `restoreSignaturesBatch` packs several counter windows into a single NUT-09 request; `restoreSignatures` becomes a thin wrapper over it.
- New config knobs on `RestoreEfficientConfig`: `batchSize` (recovery chunk size, 1-1000, default 300), `probeBudget` (probes per narrowing round, 2-40, default 12) and `ladderSkip` (low rungs to drop, 0-20, default 0). Defaults came from sweeping the benchmark: budget 12 keeps the minimum request count through T=100k while halving linked signatures versus a 30-probe round; `probeBudget: 4, ladderSkip: 3` is the measured privacy profile above.
- Tests: multi-window search, exact chunked-recovery call counts, search efficiency at T=600, out-of-range config rejection; all prior fallback and gap-tolerance tests pass unchanged.
- Benchmark script in `scripts/`, following the existing bench-* conventions.

## Reviewer Notes

- Same working assumption as #788 and Nutshell: a probe window covers 25 consecutive counters, so if a wallet ever skipped more than 25 counters in a row, an empty window could end the search early. The bundled search's early probes sit in the wallet's dense early history, which is if anything safer than bisection's arbitrary midpoints.
- The benchmark makes the underlying trade explicit: the one-at-a-time bisection sends the fewest messages and lets the mint link the fewest old signatures (its probes mostly land above T and reveal nothing); the bundled search wins on requests and wall-clock. `probeBudget` and `ladderSkip` span the space between them, and the spec discussion should present that axis rather than pick a single point.
- The live demo (`examples/nut342_recovery`) runs unchanged against the Nutshell reference branch mint and recovers a churned wallet in 3 requests.
- Draft, experimental, and stacked on #788: merge that first (or together); the `342` identifier remains provisional.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1881,6 +1881,8 @@ export type RestoreConfig = {
 export type RestoreEfficientConfig = {
     keysetId?: string;
     probeWindow?: number;
+    batchSize?: number;
+    probeBudget?: number;
     filterSpent?: boolean;
 };
 
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1883,6 +1883,7 @@ export type RestoreEfficientConfig = {
     probeWindow?: number;
     batchSize?: number;
     probeBudget?: number;
+    ladderSkip?: number;
     filterSpent?: boolean;
 };
 
--- a/examples/nut342_recovery/Makefile
+++ b/examples/nut342_recovery/Makefile
@@ -9,7 +9,7 @@
 DOCKER ?= docker
 BIND_ADDR ?= 127.0.0.1
 PORT ?= 3338
-RATE_LIMIT_PM ?= 200
+RATE_LIMIT_PM ?= 5000
 
 NUT_342_CTX ?= https://github.com/cashubtc/nutshell.git\#feat/efficient-recovery
 NUT_342_IMAGE ?= cashu-dev-nutshell-342:local
--- a/examples/nut342_recovery/nut342_recovery_demo.ts
+++ b/examples/nut342_recovery/nut342_recovery_demo.ts
@@ -154,15 +154,10 @@ async function refreshOldest(wallet: Wallet, minNet = 10) {
     if (k >= byAge.length) return; // all old value is dust; nothing sensible to refresh
   }
   const rest = proofs.filter((p) => !oldest.includes(p));
-  // net of the input fee, so the swap consumes every old proof
-  const fee = wallet.getFeesForProofs(oldest);
-  const { keep, send } = await wallet.send(sumProofs(oldest).subtract(fee), oldest, {
-    onCountersReserved,
-  });
-  proofs = [...rest, ...keep];
-  track([...keep, ...send]);
-  const received = await wallet.receive(send, { onCountersReserved });
-  proofs.push(...received);
+  // One swap: receiving our own proofs re-issues them as fresh deterministic outputs,
+  // netting the input fee itself.
+  const received = await wallet.receive(oldest, { onCountersReserved });
+  proofs = [...rest, ...received];
   track(received);
   console.log(
     `Refreshed ${oldest.length} oldest proofs (first unspent counter now ${await firstUnspentCounter()})`,
--- a/examples/nut342_recovery/nut342_recovery_demo.ts
+++ b/examples/nut342_recovery/nut342_recovery_demo.ts
@@ -129,7 +129,7 @@ async function selfSwap(wallet: Wallet, amount: number) {
   });
   proofs = keep;
   track([...keep, ...send]);
-  const received = await wallet.receive(send, { onCountersReserved, proofsWeHave: keep });
+  const received = await wallet.receive(send, { onCountersReserved });
   proofs.push(...received);
   track(received);
   console.log(`Swapped ${amount} sats through the mint (balance ${sumProofs(proofs)})`);
@@ -218,10 +218,17 @@ async function main() {
   }
   console.log(`Mint advertises NUT-342 support (${wallet.keysetId})\n`);
 
-  // A thorough workout: every operation writes fresh encrypted gaps.
-  // Random-value swap churn approximates real wallet wear; override rounds
-  // with CHURN_ROUNDS=n for a longer history.
+  // A thorough workout: every operation writes fresh encrypted gaps. Knobs, so the
+  // d_gap regimes in the PR table are one-liners:
+  //   CHURN_ROUNDS=n     swap rounds per churn block (default 4)
+  //   REFRESH_EVERY=n    oldest-refresh cadence; 0 disables it (default 20)
+  //   SPEND_FRAC=f       max spend as a fraction of balance; small values model a
+  //                      lump-funded wallet making everyday payments (default 0.9)
+  //   PIN_MID=1          send the unclaimed token mid-history, pinning d_gap at ~T/2
   const churnRounds = Number(process.env.CHURN_ROUNDS ?? 4);
+  const refreshEvery = Number(process.env.REFRESH_EVERY ?? 20);
+  const spendFrac = Number(process.env.SPEND_FRAC ?? 0.9);
+  const pinMid = process.env.PIN_MID === '1';
   const churn = async (rounds: number) => {
     for (let i = 0; i < rounds; i++) {
       const balance = Number(sumProofs(proofs).toBigInt());
@@ -229,15 +236,16 @@ async function main() {
         console.log(`Balance ${balance} too low to keep churning; stopping early`);
         return;
       }
-      await selfSwap(wallet, 1 + Math.floor(Math.random() * Math.max(1, balance * 0.9)));
-      // every 20 rounds, roll the oldest proofs forward (see refreshOldest)
-      if (i % 20 === 19) await refreshOldest(wallet);
+      await selfSwap(wallet, 1 + Math.floor(Math.random() * Math.max(1, balance * spendFrac)));
+      // periodically roll the oldest proofs forward (see refreshOldest)
+      if (refreshEvery > 0 && i % refreshEvery === refreshEvery - 1) await refreshOldest(wallet);
     }
   };
 
   await mintSats(wallet, 10000);
   await churn(churnRounds);
   await mintSats(wallet, 15000);
+  if (pinMid) await sendUnclaimed(wallet, 21);
   try {
     await meltSats(wallet, externalInvoice);
   } catch {
@@ -249,10 +257,10 @@ async function main() {
     await meltSats(wallet, target.request);
   }
   await churn(churnRounds);
-  // Sent late in history: a pending token pins first-unspent at its own counter (the
-  // provider must include it), so an unclaimed token sent mid-history would cap d_gap
-  // at ~half of T no matter how well the wallet consolidates.
-  await sendUnclaimed(wallet, 21);
+  // Sent late in history by default: a pending token pins first-unspent at its own
+  // counter (the provider must include it), so an unclaimed token sent mid-history
+  // (PIN_MID=1) caps d_gap at ~T/2 no matter how well the wallet consolidates.
+  if (!pinMid) await sendUnclaimed(wallet, 21);
 
   const balance = sumProofs(proofs);
   const pending = sumProofs(sentPending);
--- a/examples/nut342_recovery/nut342_recovery_demo.ts
+++ b/examples/nut342_recovery/nut342_recovery_demo.ts
@@ -129,10 +129,44 @@ async function selfSwap(wallet: Wallet, amount: number) {
   });
   proofs = keep;
   track([...keep, ...send]);
-  const received = await wallet.receive(send, { onCountersReserved });
+  const received = await wallet.receive(send, { onCountersReserved, proofsWeHave: keep });
   proofs.push(...received);
   track(received);
   console.log(`Swapped ${amount} sats through the mint (balance ${sumProofs(proofs)})`);
+}
+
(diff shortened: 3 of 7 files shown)


robwoodgate:
Potential NUT spec wording for laddered search (per suggested change 1 in the [nuts#342 discussion](https://github.com/cashubtc/nuts/pull/342#issuecomment-4982629858)): one sentence for the invariant, one worked trace, one pseudocode block in the draft's existing style. Every round is exactly one NUT-09 request.

**Invariant**

> A probe window containing any signature proves T is at or beyond its start; an empty window proves T is below its start, provided derivation gaps stay shorter than the window.

**Worked trace** (w = 25, wallet's unknown T = 5000; `##` = signatures returned, `--` = empty):

```
Round 1 - ladder (one request): windows at 0 and w*2^k, doubling across the space
 start:  0   25   50  100  200  400  800 1600 3200 6400 12800 ... 2^31
         ##  ##   ##  ##   ##   ##   ##   ##   ##   --    --       --
                                       last hit^    ^first empty
 => T is in (3224, 6400)

Round 2 - grid (one request): 12 windows spread evenly across the bracket
 start:  3469 3713 3957 4202 4446 4690 4934 5179 5423 5667 5911 6155
         ##   ##   ##   ##   ##   ##   ##   --   --   --   --   --
 => T is in (4958, 5179)

Repeat round 2 until grid width is < batch size (in this case no repeat).

Round 3 - tile (one request): the remaining 220 counters, wall to wall
         4959 ............... 5178
         ############---------
 => T = 5000, and its BlindSignature already carries the encrypted d_gap

Round 4 - fetch [T - d_gap, T] in chunks, filter spent per NUT-07
```

**Pseudocode**

```
Probe window (w, default 25). The number of consecutive counters checked at each probe position.
A window containing any signature proves T lies at or beyond its start; an empty window proves T
lies below it. NOTE: w must exceed the wallet's longest run of never-used counters.

Probe budget (budget, default 12). The maximum number of probe windows sent in one search
request. It sets the narrowing rate: each grid round splits the unknown zone into budget + 1
segments, so the search needs one request per ~13x of zone width at the default.

MAX = top of counter space (2^31 for legacy v0 keysets, else 2^32)

 function find_t(keyset_id) {
      // one request: a w-wide window at 0 and at each doubling position
      hits = restore([window(0)] + [window(w * 2^k) for k = 0 .. log2(MAX/w)])
      if (hits is empty) return undefined                  // caller falls back to linear scan
      floor   = highest counter with a signature           // T >= floor
      ceiling = lowest rung start above floor, else MAX+1  // T <  ceiling
      while (floor + 1 < ceiling) {
          zone = (floor + 1, ceiling)                      // width = ceiling - floor - 1
          if (width <= budget * w) {                       // default: width < 300 (12 * 25)
              hits = restore(tile(zone))                   // wall to wall
              floor = highest counter with a signature
              break                                        // zone fully probed: floor is T
          }
          hits = restore(spread(budget, zone))             // 12 probes, ~13x narrowing
          update floor and ceiling from hits
      }
      return floor    // = T; its BlindSignature carries d_gap
  }

```

The trace is generated from the implementation in this PR, not illustrative: rungs at w*2^k, a 12-window grid over the 3176-wide bracket (step 244), and a 220-counter tile.

robwoodgate:
**Worked example: T = 30,000,000** (defaults: `probeWindow` 25, `probeBudget` 12, `ladderSkip` 0)

*Round 1 (ladder, 1 request).* Windows at 0, 25, 50, 100, ... (25 * 2^k). The rung at 25 * 2^20 = 26,214,400 returns signatures; the next, at 52,428,800, comes back empty.
That brackets T inside a zone ~26.2M counters wide.

*Rounds 2-6 (grid, 1 request each).* Twelve probes spread evenly across the zone divide it into 13 segments, so each round shrinks it by ~13x:

  | Round | Zone width in | Zone width out |
  | ----- | ------------- | -------------- |
  | 2     | 26,214,375    | ~2,016,500     |
  | 3     | ~2,016,500    | ~155,100       |
  | 4     | ~155,100      | ~11,930        |
  | 5     | ~11,930       | ~918           |
  | 6     | ~918          | ~71            |

*Round 7 (tile, 1 request).* The zone is now below the tile threshold (`probeBudget * probeWindow` = 300), so one request probes every counter in it (3 windows, ~71 messages). The highest signature is T exactly, and that reply already carries the encrypted `d_gap`.

*Round 8 (restore, 1 request).* Decrypt `d_gap` (100 in the benchmark) and fetch `[T - 100, T]` as a single 300-counter chunk.

Total: 8 requests for a wallet thirty million counters old, plus the usual NUT-07 state check (= 9 total for 100 `d_gap`). In general the grid needs one round per ~13x of zone width, so even at the u32 ceiling (~4.3e9) it takes 7 grid rounds and ~10 requests overall.

PR #800, opened 2026-07-16
feat(wallet)!: remove legacy base64 keyset support

robwoodgate:
## Summary

Removes the last compatibility paths for legacy (pre-v1, base64-id) keysets, which were removed from the Cashu spec in December 2023 and have never been supported by CDK.

- `KeyChain` skips non-hex keyset ids on load (CDK-style): mints that still list legacy keysets (eg Minibits) load fine, the legacy keysets just never enter the chain
- Token decode, NUT-13 derivation and `getKeysetIdInt` throw for base64 keyset ids with an actionable error
- `Keyset.verifyKeysetId` returns false for non-hex ids
- `Mint.getKeys` drops the URL-safe keyset id encoding
- Public API: removes `DeriveKeysetIdOptions.isDeprecatedBase64` and `Keyset.hasHexId`
- Migration guide section added, including the sweep-before-upgrading advice

## Breaking change

Proofs on legacy base64 keysets can no longer be decoded, spent or restored. See `migration-5.0.0.md.`


robwoodgate:
superseded by #813, #835

PR #801, opened 2026-07-16
feat(wallet): warn when a mint has deprecated legacy keysets

robwoodgate:
## Summary

Legacy (pre-v1, base64-id) keyset support is planned for removal in v6 (#800 carries the removal, parked for the v6 window). This PR gives consumers a full major cycle of notice:

- `loadMint()` / `loadMintFromCache()` now log a warning naming any legacy keysets the mint lists, so apps can prompt users to sweep old proofs during v5
- `DeriveKeysetIdOptions.isDeprecatedBase64` is marked `@deprecated`
- the v5 migration guide documents the deprecation and the v6 removal plan

No behaviour change otherwise: legacy keysets still load and remain spend/restore-only, exactly as before.

## Context

Base64 keyset ids left the spec in December 2023 and CDK has never supported them, but some nutshell mints (notably Minibits) still serve them and may keep a redemption path alive long-term. Removing support at v5 GA would strand redeemable proofs and give affected wallets a reason not to upgrade; warn-now-remove-in-v6 gives everyone a clean runway.

### DIFF
--- a/migration-5.0.0.md
+++ b/migration-5.0.0.md
@@ -12,6 +12,14 @@ Existing v0 (legacy base64), v1 (`00…`), and v2 (`01…`) keysets continue to 
 
 ---
 
+## Legacy (base64-id) keysets are deprecated, removal planned for v6
+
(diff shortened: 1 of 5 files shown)


robwoodgate:
superseded by #813, #835

PR #806, opened 2026-07-16
fix(crypto): validate counter range in BIP-32 secret derivation

robwoodgate:
Adds a range check to the deprecated BIP-32 secret/blinding-factor deriver so a non-integer or out-of-range `counter` (outside `[0, 2^31)`) throws instead of deriving an unintended child index.

## Why

#796 replaced the string-path derivation with `HARDENED_OFFSET + counter` arithmetic. The string path implicitly rejected out-of-range counters via the path regex; the arithmetic form does not, so a value below zero resolves to a valid child index and derives silently. This restores the fail-fast behavior and matches the guard already present on `createKeyPairDeriver`.

## Test

Adds a case asserting a negative counter throws on the deprecated BIP-32 path. `npm run prtasks` passes locally.

### DIFF
--- a/src/crypto/NUT13.ts
+++ b/src/crypto/NUT13.ts
@@ -179,6 +179,11 @@ function deriveBip32SecretAndBlindingFactor(
   parentKey: HDKey,
   counter: number,
 ): DerivedSecretAndBlindingFactor {
+  // HARDENED_OFFSET + counter must stay in the hardened range; a counter outside [0, 2^31) would
+  // wrap to a different index and derive the wrong key rather than fail.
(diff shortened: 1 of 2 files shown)


robwoodgate:
Backport failed for `v4-dev`, because it was unable to cherry-pick the commit(s).

```bash
git fetch origin v4-dev
git worktree add -d .worktree/backport-806-to-v4-dev origin/v4-dev
cd .worktree/backport-806-to-v4-dev
git switch --create backport-806-to-v4-dev
git cherry-pick -x e61c0648fec9c62dc1b999ad18ff813b7fea3887
```

PR #813, opened 2026-07-20
feat(wallet): prefer stale keysets in default proof selection

robwoodgate:
## Summary

Adds `selectProofsRotating`, a wrapper around `selectProofsRGLI` that preferentially spends proofs on stale keysets, and makes it the Wallet default. Balances migrate onto the mint's current keyset during normal sends, swaps and melts.

- Buckets proofs by staleness: legacy base64 ids first (whatever their status), then older version bytes, inactive before active within a version. An old version never queues behind a newer version's rotation churn.
- Force-includes whole stale buckets, uneconomical dust included. The set-level fee ceil makes aggregation cost at most about a sat per dust proof, and flushing beats stranding the value when legacy formats are eventually dropped.
- Delegates the residual to RGLI: boundary bucket first, then boundary plus fresher buckets, then plain full-set RGLI as the fallback. Merged results are verified under a single fee ceil, so split accounting can never overpay an exact match.
- Wrapper arithmetic is bigint, so u64-scale proofs on stale keysets flush without hitting RGLI's number guard.

## Behavior changes

- The default selector now prefers stale keysets. Opt out with `selectProofs: selectProofsRGLI` in the Wallet options. `selectProofsRGLI` is unchanged, and the fallback path reproduces current behavior exactly.
- Wallets holding uneconomical dust (fee at or above value) now deliberately spend it to drain old keysets, a one-off cost of up to about a sat per dust proof.

## Why

Base64 keyset ids left the spec long ago but some remain in use, and NUT-02 says wallets SHOULD swap proofs off inactive keysets. Draining old keysets as a side effect of normal spending also keeps unspent proofs clustered on recent keysets, which benefits wallet restore schemes.

## Testing

18 unit tests cover bucket ordering, forced-walk invariants, exact-match fallbacks, unified fee verification (including rejection of a split-computed off-by-one exact match), u64-scale flushing, and a 50k-proof scale case. Full suite and prtasks green.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1903,6 +1903,9 @@ export type SelectProofs = (proofs: ProofLike[], amountToSelect: AmountLike, key
 export function selectProofsRGLI(proofs: ProofLike[], amountToSelect: AmountLike, keyChain: KeyChain, includeFees?: boolean, exactMatch?: boolean, _logger?: Logger): SendResponse;
 
 // @public
+export function selectProofsRotating(proofs: ProofLike[], amountToSelect: AmountLike, keyChain: KeyChain, includeFees?: boolean, exactMatch?: boolean, _logger?: Logger): SendResponse;
+
+// @public
 export class SendBuilder {
     constructor(wallet: Wallet, amount: AmountLike, proofs: ProofLike[]);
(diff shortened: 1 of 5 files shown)


robwoodgate:
- #818

PR #814, opened 2026-07-20
fix(wallet): widen exact-match trim bound for selection-level fee rounding

robwoodgate:
## Summary

`selectProofsRGLI`'s exact-match pre-trim discarded candidates at `exFee <= target`, but fees round up once per selection rather than per proof, so the real net can sit up to a sat below the fractional `exFee` sum. A proof whose `exFee` is just above the target can therefore still net the target exactly, and the old bound made such solutions unreachable: a wallet holding a single 6 sat proof at 500ppk could never exact-match a 5 sat send (`exFee` 5.5, net `6 - ceil(500/1000) = 5`).

The trim bound is now `target + 1`. Solution acceptance (`sumExFees`, true fee ceil) is unchanged, so nothing new can be wrongly accepted; the only behavior change is that previously missed exact matches are found, meaning fewer needless swaps where an offline exact send was possible.

## Scope note

Close match has a milder analog (a candidate just past the `nextBiggerExFee` cutoff can carry a marginally lower-ppk delta), but close match always finds a valid solution and the realized fee after the ceil is essentially identical, so that trim is deliberately left alone.

## Testing

One regression test proving the previously missed exact match is found; full suite and prtasks green.

### DIFF
--- a/src/wallet/SelectProofs.ts
+++ b/src/wallet/SelectProofs.ts
@@ -158,12 +158,15 @@ export function selectProofsRGLI(
   spendableProofs.sort((a, b) => a.exFee - b.exFee);
 
   // Remove proofs too large to be useful and adjust totals
-  // Exact Match: Keep proofs where exFee <= amountToSend
+  // Exact Match: Keep proofs where exFee <= amountToSend + 1. Fees round up
+  // once per selection, not per proof, so the real net can be up to a sat
+  // below the exFee sum: a 6 sat proof at 500ppk has exFee 5.5 yet nets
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #819 with remaining conflicts!

```bash
git fetch origin backport-814-to-v4-dev
git worktree add --checkout .worktree/backport-814-to-v4-dev backport-814-to-v4-dev
cd .worktree/backport-814-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 78e1180ffbbf44c45d25aefccd238b222f02ec6d
```

PR #815, opened 2026-07-20
refactor(wallet): make RGLI exFee an integer sort key

robwoodgate:
This is a stepping stone to Bigint aware RGLI.

## Summary

`exFee` is only ever a sort key: it drives the ordering, binary searches, trims and the economical filter, never the fee accounting. This change floors each proof's own fee in the `exFee` calculation so the key is an integer instead of a fine-grained double.

- The economical filter decides identically: fractional keeps a proof iff `1000*amount > ppk`, floored keeps iff `amount > floor(ppk/1000)`, and for integers those are the same predicate.
- Set-level fee accounting (`sumExFees`, one ceil per selection) is untouched, so nothing about acceptance changes.
- Ordering coarsens by sub-sat distinctions only, which can shift which of two delta-equal solutions a trial finds, never validity. The full suite passes unchanged.
- The `target + 1` exact-match trim bound from #814 composes: members of an exact solution have integer `exFee <= target + 1`, and the trim comment's example gets cleaner (a 6 sat proof at 500ppk has `exFee` 6 yet nets 5).

This removes the last fractional layer from the selection hot path, which is what makes a future bigint port of RGLI mechanical rather than a rewrite.

## Testing

Two new boundary tests: one thousand 1 sat proofs at 999ppk aggregate into exactly one spendable sat (fails under any per-proof fee rounding, protecting the aggregation rebate), and a 1 sat proof at 1000ppk remains uneconomical. Full suite and prtasks green.

### DIFF
--- a/src/wallet/SelectProofs.ts
+++ b/src/wallet/SelectProofs.ts
@@ -141,7 +141,10 @@ export function selectProofsRGLI(
     }
     const ppkfee = feeForProof(p);
     const amountNum = p.amount.toNumber(); // safe: guarded above
-    const exFee = includeFees ? amountNum - ppkfee / 1000 : amountNum;
+    // Floor the proof's own fee: keeps exFee an integer sort key, and the economical
+    // filter decides identically (amount > floor(ppk/1000) iff 1000 * amount > ppk);
+    // set-level fee accounting (sumExFees) is untouched
+    const exFee = includeFees ? amountNum - Math.floor(ppkfee / 1000) : amountNum;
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #821 with remaining conflicts!

```bash
git fetch origin backport-815-to-v4-dev
git worktree add --checkout .worktree/backport-815-to-v4-dev backport-815-to-v4-dev
cd .worktree/backport-815-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x bd0ab89662829d897177e12266ff6e4862d82740
```

PR #822, opened 2026-07-20
feat(wallet): support u64 amounts in proof selection

robwoodgate:
## Summary

Ports `selectProofsRGLI`'s internals from `number` to `bigint` arithmetic, making proof selection exact at any u64 scale and removing the `MAX_SAFE_INTEGER` guard. Selection was the last number-arithmetic path in the library; msat-denominated mints have been representable since the Amount VO landed, and this closes the one API that could still throw on them.

- Amounts, fee sums, `exFee` and the trim bounds are all bigint; the set-level fee ceil becomes `(ppk + 999n) / 1000n`, matching `getFeesForProofs`.
- The delta tiebreaker is scaled by 1000 into integers (`1000n*amount + feePPK - 1000n*target`), with a null sentinel replacing `Infinity`.
- Shuffles, indexes, trial counters and the time budget stay plain numbers; only value arithmetic changed.
- The dead close-match overage knobs (`MAX_OVRPCT`/`MAX_OVRAMT`, always zero, kept for tuning that never proved necessary) are removed; `maxOverAmount` reduces to `min(target, totalNetSum)` with identical behavior.
- Single implementation: no number/bigint dispatch, nothing extra to maintain.

## Benchmarks (bigint port vs pre-port number baseline)

Mixed-fee close match and solvable exact match, median of 5 runs, power-of-2 denominations:

| Scenario | Proofs | bigint | number |
|---|---|---|---|
| close match | 1k | 0.5ms | 0.7ms |
| close match | 10k | 2.1ms | 2.1ms |
| close match | 50k | 12.5ms | 12.3ms |
| exact match | 1k | 0.2ms | 0.2ms |
| exact match | 10k | 1.8ms | 2.0ms |
| exact match | 50k | 11.7ms | 12.5ms |

Overage quality identical, zero exact-match timeouts on either side. The per-op bigint cost disappears into RGLI's real hot paths (shuffles, array and Set churn), which are index-and-object work.

## Behavior changes

- Proofs and targets above `Number.MAX_SAFE_INTEGER` now select correctly instead of throwing; the "provide a custom SelectProofs implementation" carve-out for msat-scale wallets is gone.
- No other observable change: the full suite passes unchanged, and existing sat-scale selection behavior is identical.

## Testing

Four new u64 tests (huge-proof close and exact match, mixed huge and small exact change, and a 1024-proof sum past 2^53 that double arithmetic would drift on), the obsolete guard test removed, full suite and prtasks green (6504 passing).

### DIFF
--- a/src/wallet/SelectProofs.ts
+++ b/src/wallet/SelectProofs.ts
@@ -1,5 +1,5 @@
 // Minimal types to avoid importing the whole wallet, keeps this module independent
-import { fail, failIf, failIfNullish, type Logger, NULL_LOGGER, measureTime } from '../logger';
+import { failIf, failIfNullish, type Logger, NULL_LOGGER, measureTime } from '../logger';
 import { Amount, type AmountLike } from '../model/Amount';
 import { CTSError } from '../model/Errors';
 import type { Proof, ProofLike } from '../model/types/proof';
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #823 with remaining conflicts!

```bash
git fetch origin backport-822-to-v4-dev
git worktree add --checkout .worktree/backport-822-to-v4-dev backport-822-to-v4-dev
cd .worktree/backport-822-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 42c97e3cebc74408d74499a11af7a879885828d1
```

PR #824, opened 2026-07-20
refactor(model): return snapshots from MintInfo accessors

robwoodgate:
## Summary

`MintInfo` accessors returned references into the normalized internal response, so an in-place edit by a consumer (or a plugin sharing the instance) could alter later reads and the persisted `cache`. Because `AmountLike` fields legally accept `bigint`, such an edit is even type-legal, and the damage only surfaces later, for example when serializing the cache. The class now has a one-line contract, stated in its TSDoc: accessors return snapshots, never internal references.

- A private structural `snapshot()` helper deep-copies plain data, preserving `bigint` values exactly. A JSON round-trip was considered and rejected: `JSON.stringify` cannot serialize bigints, and `JSONInt` canonicalizes by magnitude (small bigints come back as numbers), which is right for wire serialization but wrong for a faithful copy.
- Applied at every reference-returning surface: the `cache`, `nuts` and `contact` getters, `checkMintMelt` (covering `isSupported(4|5)`, `supportedMethods` and `getMintMeltMethod`), the NUT-15/17 params arrays, NUT-19's embedded `cached_endpoints`, and NUT-29's params object.
- String getters (`name`, `description`, `version`, ...) are deliberately unchanged: primitives cannot leak references, so the presence of `snapshot()` marks exactly which returns used to be shared.

## Behavior notes

- Returned objects are now safe to mutate or persist; edits no longer feed back into the instance. Object identity across calls is no longer stable (it was never contractual).
- Perf: the cloned values are small (mint info sized); the only internal consumer of an affected getter is a one-shot auth lookup.

## Testing

Four new tests: a type-legal `min_amount = 10n` write-back followed by cache serialization, copy verification for `supportedMethods`/`isSupported(4).params`, mutation isolation for `cache`/`nuts`/`contact`, and the NUT-17 params array. The existing suite's bigint-preservation tests validated the snapshot fidelity choice. Full suite and prtasks green (6520 passing).

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1107,7 +1107,6 @@ export type MintContactInfo = {
 // @public
 export class MintInfo {
     constructor(info: GetInfoResponse, logger?: Logger);
-    // (undocumented)
     get cache(): GetInfoResponse;
     // (undocumented)
     get contact(): MintContactInfo[];
--- a/src/model/MintInfo.ts
+++ b/src/model/MintInfo.ts
@@ -37,7 +37,8 @@ type ProtectedIndex = {
  * Wraps a mint's `/v1/info` response with capability helpers.
  *
  * Constructable from any `GetInfoResponse`, including one persisted from an earlier session; no
- * live connection is needed. Persist `cache` and rehydrate with `new MintInfo(stored)`.
+ * live connection is needed. Persist `cache` and rehydrate with `new MintInfo(stored)`. Accessors
+ * return snapshots, never internal references.
  */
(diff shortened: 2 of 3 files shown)


robwoodgate:
- #825 with remaining conflicts!

```bash
git fetch origin backport-824-to-v4-dev
git worktree add --checkout .worktree/backport-824-to-v4-dev backport-824-to-v4-dev
cd .worktree/backport-824-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 26d639a12981dd74c48a968a6db06cd343bd2e64
```

PR #835, opened 2026-07-21
feat(wallet): prefer newest keyset version in getCheapestKeyset

robwoodgate:
Rotation-aware proof selection drains old-keyset proofs, but a fee-only default keyset choice can mint change straight back onto them. This orders active keysets by ID version first, then fee, then `final_expiry` (latest wins; none sorts as never expiring), so new outputs land on the mint's newest keyset generation. Tie-breaks are now deterministic (previously insertion order).

- Adds `Keyset.version` (ID version byte, `-1` for deprecated base64 IDs)
- `getCheapestKeyset()` keeps its name for v4 backportability; TSDoc now documents the "cheapest modern" ordering
- Tests cover version-over-fee, fee-over-expiry, latest-expiry and no-expiry ties

Note for merge: keep the `Release-As: 5.0.0-rc.4` footer in the squash commit body (repins the pending release after the NUT-18 major bump).

PR #837, opened 2026-07-21
fix(wallet): return -1 from Keyset.version for unparseable ids

robwoodgate:
Follow-up to #835.

`isValidHex` accepts odd-length strings, but `hexToBytes` rejects them, so `Keyset.version` threw on such ids. Treat them as versionless (-1), matching base64 ids.

Not reachable from mint data (keysets that fail verification are dropped before the version sort); only manually constructed Keysets were affected.

### DIFF
--- a/src/wallet/Keyset.ts
+++ b/src/wallet/Keyset.ts
@@ -56,10 +56,10 @@ export class Keyset {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #838

PR #839, opened 2026-07-22
fix(wallet): classify odd-length hex keyset ids as legacy

robwoodgate:
Follow-up to #837.

Internal `isValidHex` accepted odd-length strings, so consumers disagreed on how to classify such ids: `Keyset.version` returned -1 while proof-selection ranking parsed a version byte from them, the output-keyset guard let them through, and token encoding failed deep inside hex decoding with a raw error. Requiring byte-decodable (even-length) hex in `isValidHex` makes every consumer treat them as legacy consistently: rotation-aware selection ranks them in the legacy bucket, new outputs refuse them, and `getEncodedToken` raises the proper legacy-keyset CTSError.

The NUT-13 derivation paths keep their own local hex checks unchanged, preserving historical derivation semantics. No behavior change for well-formed ids.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -773,7 +773,6 @@ export class Keyset {
     // (undocumented)
     get fee(): number;
     static fromMintApi(meta: MintKeyset, keys?: MintKeys): Keyset;
-    // (undocumented)
     get hasHexId(): boolean;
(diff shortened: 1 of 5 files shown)


robwoodgate:
@codex review

robwoodgate:
- #840

PR #851, opened 2026-07-22
fix(wallet): validate pubkey in locked mint quote responses

robwoodgate:
`createLockedMintQuote` accepted any pubkey in the mint's response as long as one was present; `createMintQuoteBolt12` and `createMintQuoteOnchain` did not inspect the echoed pubkey at all. All three now verify the response pubkey matches the requested one (case-insensitive) and fail fast with an actionable error at quote creation. Adds mutant tests for the three paths.


robwoodgate:
- #852

PR #854, opened 2026-07-22
fix(wallet): bound fee convergence loop

robwoodgate:
`receiveFeeAmounts` converges on the fee that covers its own fee outputs by advancing one unit at a time. The number of steps scales with `input_fee_ppk`, and the loop does not converge at all when the per-input fee exceeds the keyset's largest denomination, so a keyset advertising an extreme `input_fee_ppk` could make fee calculation (any `includeFees` path, or `getFeesToInclude`) run for a very long time.

Bounds the loop and throws a clear error past the cap. Any realistic keyset converges in a few dozen steps (even 10 sat/input is ~23), so the bound only trips on degenerate fee metadata; the minimal-fee result is unchanged for every converging case.

Adds regression tests: minimal (fewest-output) convergence at 3 sat/input, and a fail-fast throw for an extreme fee.

### DIFF
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -784,11 +784,19 @@ class Wallet {
    * Iterates until the fee is stable, since fee outputs also incur fees.
    */
   private receiveFeeAmounts(nOutputs: number, keyset: Keyset): Amount[] {
+    // Bound the +1 convergence; any realistic keyset converges in a few dozen steps.
+    const maxIters = 1 << 12;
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #855 with remaining conflicts!

```bash
git fetch origin backport-854-to-v4-dev
git worktree add --checkout .worktree/backport-854-to-v4-dev backport-854-to-v4-dev
cd .worktree/backport-854-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 308b3cd8ca310c0d954d30b57ebc3b179ae39023
```

PR #856, opened 2026-07-22
fix(wallet): reject a missing pubkey in locked mint quotes

robwoodgate:
The locked mint quote methods (`createLockedMintQuote`, `createMintQuoteBolt12`, `createMintQuoteOnchain`) type `pubkey` as a required string, but a loosely-typed JavaScript caller can omit it. When the mint then returns a string pubkey, the response echo check called `pubkey.toLowerCase()` on `undefined` and threw a raw `TypeError` instead of a clear error.

Validate the argument at method entry (`typeof pubkey !== 'string' || empty`) and fail fast with an actionable message, before any request is sent. Adds a test per method.


robwoodgate:
- #857

PR #859, opened 2026-07-23
fix(utils): bound splitAmount output count

robwoodgate:
`splitAmount` fills a value from a keyset's denominations. When a keyset offers only small denominations relative to the value, the denomination fill could build a very large output array, previously via a spread push that also overflows the call stack at high counts.

Bound the fill to a sane maximum (8192, 8x cdk's default `max_outputs`) and reject before allocating, comparing the required count against the remaining budget so an oversized count never reaches `toNumber()`. Normal splits over a power-of-two keyset are unaffected (a few dozen outputs). Adds an inclusive-boundary test.

Also swaps a bit-shift literal in the fee-convergence cap (`1 << 12` -> `4_096`) for readability, no behaviour change.

### DIFF
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -34,6 +34,12 @@ import { Bytes } from './Bytes';
 import { decodeCBOR, encodeCBOR } from './cbor';
 import { JSONInt } from './JSONInt';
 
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #868

PR #864, opened 2026-07-23
fix(wallet): bound keyset denomination count at ingest

robwoodgate:
## Description

Caps the number of keys a keyset may carry at 256, checked in `Keyset.verifyKeysetId` before id derivation. All keyset ingest funnels through `verify()`, so the single check covers every path. Mint responses are untrusted input; per the contributing guidelines, any work they drive should be bounded. Real keysets carry ~64 denominations (powers of two up to 2^63), so the bound leaves ample headroom.

An oversized keyset now fails verification the same way a wrong-id keyset does: `verifyKeysetId` returns false and the keyset is rejected at load.

## Testing

New unit test: a 257-key keyset whose id genuinely derives is rejected, proving rejection is due to the count bound rather than an id mismatch. `npm run prtasks` is green; no public API change.


robwoodgate:
- #865

PR #869, opened 2026-07-23
fix(wallet): compute keyset fees with integer arithmetic

robwoodgate:
`getFeesForKeyset` computed the fee in JS `number` space (`Math.floor((nInputs * feePPK + 999) / 1000)`). For a large `input_fee_ppk` the `nInputs * feePPK + 999` intermediate crosses `Number.MAX_SAFE_INTEGER` and the result can be one off. It now uses `Amount.ceilPercent`, which keeps the product in bigint, so the fee is exact for any value the ingest bound admits.

Alongside, `ceilPercent`/`floorPercent` now accept a zero numerator (ceil/floor of 0% is 0), which the fee path needs for zero-fee keysets rather than special-casing them. Negative and non-integer numerators still throw.

## Testing

- Unit: zero-numerator now returns 0 for both helpers; negative/zero-denominator still throw.
- Wallet: a large `input_fee_ppk` yields the exact integer fee (the `number` path was off by one sat), and a zero-fee keyset returns zero.
- `npm run prtasks` green; no public API change.

### DIFF
--- a/src/model/Amount.ts
+++ b/src/model/Amount.ts
@@ -207,11 +207,13 @@ export class Amount {
    *     amount.ceilPercent(1, 200); // ceil(0.5% of amount)
    *     amount.ceilPercent(15, 10); // ceil(1.5% of amount)
    *
-   * @throws If numerator or denominator are not positive integers.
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #870

PR #874, opened 2026-07-23
fix(crypto): bound untrusted P2PK witness and CBOR decode input

robwoodgate:
Bounds a few untrusted inputs on the P2PK/HTLC and CBOR paths so work stays proportional to a sane maximum rather than to input size:

- Cap pubkey and signature counts on the P2PK/HTLC verify, sign, and P2BK-derive paths (the two key extractors are the shared chokepoint). Signatures get more headroom than keys because SIG_ALL carries a signature per message variant per signer.
- Treat a non-array witness `signatures` value as an empty array.
- Bound CBOR decode recursion to a fixed depth, returning a `CTSError` for input nested beyond it.

## Testing

Unit: over-cap pubkeys/signatures throw a CTSError, a SIG_ALL-sized (33-signature) witness is accepted, a non-array `signatures` verifies as false, and deeply nested CBOR throws a CTSError instead of overflowing. `npm run prtasks` green; no public API signature change.

### DIFF
--- a/src/crypto/NUT11.ts
+++ b/src/crypto/NUT11.ts
@@ -30,6 +30,14 @@ export const SigFlags = {
 export type SigFlag = (typeof SigFlags)[keyof typeof SigFlags];
 const VALID_SIG_FLAGS: ReadonlySet<SigFlag> = new Set(Object.values(SigFlags));
 
+// Upper bounds on untrusted P2PK/HTLC secret and witness sizes, applied on the verify/sign path so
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #876 with remaining conflicts!

```bash
git fetch origin backport-874-to-v4-dev
git worktree add --checkout .worktree/backport-874-to-v4-dev backport-874-to-v4-dev
cd .worktree/backport-874-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x e8b5c1e826411f20b39a917f3708112af821c88c
```

PR #877, opened 2026-07-23
fix(crypto): harden P2PK witness edge cases

robwoodgate:
Follows up #874 with two edge-case behaviours on the P2PK witness path:

- `parseWitnessData` treats a JSON-primitive witness (eg `"null"`) as
  absent instead of throwing a `TypeError` when reading its fields.
- `signP2PKProof` refuses to append a signature to a witness already at
  the signature limit, so signing cannot produce a proof the verify path
  then rejects.

## Testing

Regression tests beside the #874 witness bound tests: a `"null"` witness
verifies as false, and signing an at-cap witness throws a `CTSError`.
`npm run prtasks` green; no public API change.


robwoodgate:
- #878

PR #892, opened 2026-08-06
fix(crypto): harden HTLC witness parsing edge cases

robwoodgate:
`getHTLCWitnessPreimage` parses an untrusted witness, but the try/catch only traps JSON syntax errors. A witness string that parses to a JSON primitive (eg `"null"`) got past the catch and then had a field read off it, so a malformed witness surfaced a raw error at the verification boundary instead of being treated as absent.

This adds the same non-object guard the P2PK witness parser (`parseWitnessData`) already carries, so both NUT-11 and NUT-14 witness parsing behave consistently: a parsed primitive is treated as no witness and returns `undefined`.

Regression tests cover `"null"`, a numeric primitive and a bare string. `npm run prtasks` green.

### DIFF
--- a/src/crypto/NUT14.ts
+++ b/src/crypto/NUT14.ts
@@ -192,6 +192,8 @@ export function getHTLCWitnessPreimage(witness: Proof['witness']): string | unde
     console.error('Failed to parse HTLC witness string:', e);
     return undefined;
   }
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #898

PR #893, opened 2026-08-06
fix(crypto): tighten NUT-10 secret and tag-integer parsing

robwoodgate:
Two input-hygiene fixes at the NUT-10 parse boundary.

**`parseSecret` tags check.** `tags` was validated behind an `if (data.tags)` truthiness check. A present but falsy non-array value (`0`, `false`, `""`) skipped the array validation and passed straight through, where a downstream tag reader (`.some()`/`for..of`) then choked on it. The check now gates on presence (`!= null`), so any defined non-array value is rejected here with a controlled `CTSError`. Per NUT-10, `tags` is optional and, when present, an array; `null`/absent are tolerated as "no tags" (matching `getTags`' `?? []`) and a tolerated `null` is normalised to `undefined` on output so the field is never a runtime `null`.

**`getTagInt` parsing.** It used `Number.parseInt`, which partial-parses (`"1700000000abc"` -> `1700000000`) and accepts hex/exponent/float forms. It now requires a strict base-10 integer within the safe-integer range, returning `undefined` otherwise, so a malformed tag reads as absent. The sign is deliberately retained: callers apply their own semantic checks (`assertPositiveInteger` for `n_sigs`, `getLocktime`'s `<= 0` guard) and those stay fail-closed on an out-of-range value rather than seeing it silently dropped.

Regression tests cover the falsy-non-array tags cases, the null/absent tolerance (asserting normalised output), and the strict-integer rejections. `npm run prtasks` green.

### DIFF
--- a/src/crypto/NUT10.ts
+++ b/src/crypto/NUT10.ts
@@ -87,8 +87,8 @@ export function parseSecret(secret: string | Secret): Secret {
   if (typeof data.nonce !== 'string' || typeof data.data !== 'string') {
     throw new CTSError('Invalid NUT-10 secret nonce / data');
   }
-  if (data.tags) {
-    // Check data.tags is an array
+  if (data.tags != null) {
+    // Check data.tags is an array (a falsy non-array must be rejected here)
     if (!Array.isArray(data.tags)) {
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #899

PR #894, opened 2026-08-06
fix(wallet): harden P2PK spending-condition locktime and tag validation

robwoodgate:
`P2PKBuilder` (and the shared write path beneath it) accepted inputs that produced NUT-10/11 secrets the library later refuses to parse or verify. This tightens both.

**Builder (`P2PKBuilder`)**
- `lockUntil` validates the resolved Unix time via `assertUnixSeconds`, rejecting `NaN`/`Infinity`/negative values and invalid `Date`s at the setter, instead of letting a bad value flow through and be silently dropped downstream.
- `addTag` rejects empty tag values, matching `parseSecret`'s read-side rule (NUT-10 tag values must be non-empty strings) and the file's existing "validate at the setter" convention.
- `fromOptions` assigns the canonical seconds `locktime` directly rather than replaying it through `lockUntil`'s millisecond heuristic. A locktime already in seconds and `>= 1e12` was otherwise divided by 1000 and expired; the heuristic only belongs at the ergonomic `lockUntil` entry, not when re-reading an already-normalised options object.

**Write chokepoint (`buildP2PKTags`)**
Both the builder and a direct `OutputData.createSingleP2PKData(options)` call flow through `buildP2PKTags`, so the same rules are enforced there:
- A present-but-invalid `locktime` now throws instead of being silently dropped. A dropped locktime alongside refund keys is a structural violation the verifier rejects outright (`refund keys require a locktime`), leaving a secret that is unspendable via any path, so it must fail at construction.
- Empty `additionalTags` values are rejected, matching the read side.

Regression tests cover the builder setters, the `fromOptions` round-trip of a `>= 1e12` second locktime, and the direct-build-path rejections. A prior test that pinned the silent-locktime-drop is updated to expect the throw, and the `MAX_SECRET_LENGTH` boundary test is rebased off a single-character pad (it had used an empty tag value as a measuring stick). `npm run prtasks` green.

### DIFF
--- a/src/crypto/NUT11.ts
+++ b/src/crypto/NUT11.ts
@@ -341,9 +341,14 @@ export function buildP2PKTags(lock: LockConditions): string[][] {
   const pubkeys = lock.pubkeys ?? [];
   const refund = lock.refundKeys ?? [];
 
-  const ts = lock.locktime ?? NaN;
-  if (Number.isSafeInteger(ts) && ts >= 0) {
-    tags.push(['locktime', String(ts)]);
+  // Reject a present-but-invalid locktime, don't drop it: refund keys without a locktime
+  // is a structural violation the verifier rejects outright ("refund keys require a
+  // locktime"), so a silent drop makes the whole proof unspendable, main keys included.
+  if (lock.locktime !== undefined) {
+    if (!Number.isSafeInteger(lock.locktime) || lock.locktime < 0) {
+      throw new CTSError(`locktime must be a non-negative integer, got ${lock.locktime}`);
+    }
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #900 with remaining conflicts!

```bash
git fetch origin backport-894-to-v4-dev
git worktree add --checkout .worktree/backport-894-to-v4-dev backport-894-to-v4-dev
cd .worktree/backport-894-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 6985fa1a67eabb83dc23fd6317db75722cb26930
```

PR #895, opened 2026-08-06
fix: guard CBOR and keyset map lookups against prototype-chain keys

robwoodgate:
Two internal maps resolved untrusted string keys through the object prototype chain. This finishes applying the null-prototype / explicit-define pattern the codebase already uses in `JSONInt` and `WalletEvents`.

**`decodeMap` (`utils/cbor.ts`)**
Map entries were assigned with `map[key] = value`. A `"__proto__"` key (reachable when decoding an untrusted V4 token or `creqA` payment request) hit the prototype setter and reparented the decoded object, dropping the key as data and pulling in inherited members. Entries are now written with `Object.defineProperty`, so any key, `__proto__` included, becomes an own data property, exactly as `JSONInt.parseObject` already does.

**`KeyChain` (`wallet/KeyChain.ts`)**
The `keysets` map was a bare object literal, so `getKeyset('constructor')` (or `'__proto__'`, `'toString'`, etc.) returned an inherited `Object.prototype` member. That value is truthy, so it slipped past the `if (!keyset)` not-found guard and was handed back as a keyset. The map is now created with `Object.create(null)` at both the field declaration and the rebuild reset, so a non-own id resolves to `undefined` and `getKeyset` throws as expected. `WalletEvents` already uses this pattern for the same reason.

Tests assert the decoded CBOR object keeps `Object.prototype` (no reparent, no leaked inherited property, `__proto__` preserved as an own key) and that `getKeyset` rejects the four prototype-chain ids. `npm run prtasks` green.

### DIFF
--- a/src/utils/cbor.ts
+++ b/src/utils/cbor.ts
@@ -479,7 +479,14 @@ function decodeMap(
       throw new CTSError('Invalid key type');
     }
     const valueResult = decodeItem(view, keyResult.offset, depth + 1);
-    map[String(keyResult.value)] = valueResult.value;
+    // Define explicitly so a "__proto__" key becomes an own data property instead of
+    // hitting the prototype setter and reparenting the decoded map (see JSONInt.parseObject).
+    Object.defineProperty(map, String(keyResult.value), {
+      value: valueResult.value,
+      writable: true,
+      enumerable: true,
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #901

PR #896, opened 2026-08-06
fix(auth): validate OIDC provider URLs use an http(s) scheme

robwoodgate:
OIDC providers supply URLs the client then navigates to or opens: `authorization_endpoint` (the auth-code redirect) and the device flow's `verification_uri` / `verification_uri_complete`. These came straight from the discovery document and device response and were used without a scheme check, so a `javascript:` (or `data:`) value would run when opened.

A shared `assertHttpUrl` helper parses the value with `new URL` and requires an `http:`/`https:` protocol, throwing a `CTSError` otherwise. It is applied to:
- `authorization_endpoint` in `buildAuthCodeUrl`, before the redirect URL is built.
- `verification_uri` and `verification_uri_complete` in `deviceStart`, which also covers `startDeviceAuth` (it spreads the result).

`http` is allowed alongside `https` because local-dev providers use it. `token_endpoint` and `device_authorization_endpoint` are only ever POSTed to via `fetch` (a non-http scheme there cannot execute), so they are intentionally out of scope here to avoid coupling the flows. `npm run prtasks` green.

### DIFF
--- a/src/auth/OIDCAuth.ts
+++ b/src/auth/OIDCAuth.ts
@@ -52,6 +52,24 @@ export type OIDCAuthOptions = {
   onTokens?: (t: TokenResponse) => void | Promise<void>;
 };
 
+/**
+ * Assert a provider-supplied URL uses an http(s) scheme.
+ *
+ * @remarks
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #902

PR #904, opened 2026-08-07
fix(crypto): count unique signers by x-only pubkey identity

robwoodgate:
The exported signer helpers disagreed on what makes a pubkey unique: `getValidSigners` deduped candidates by exact string, while `schnorrVerifyMessage`/`schnorrVerifyDigest` verify against the x-only key (BIP-340 ignores the parity prefix). One key written in different encodings (X-only, 02/03 prefix, letter case) was therefore counted as several signers. The NUT-11 P2PK path in this repo is unaffected because keys arrive pre-normalized via `normalizeSecpPubkey`; this tightens the helpers for direct callers.

- A single normalization helper (lowercase, strip a 02/03 prefix) now defines key identity for both dedupe and verification, so the two can never diverge again.
- `getValidSigners` dedupes by the x-only form, keeping the first occurrence.
- `schnorrVerifyDigest` only strips a recognized 02/03 prefix; previously any 66-char input had its first byte dropped, so e.g. an `ff`-prefixed key verified as its x-coordinate. Such inputs now fail verification.

No public API signature changes. New unit tests cover multi-encoding dedupe, signer-threshold counting, and the non-02/03 prefix reject. `npm run prtasks` green.

### DIFF
--- a/src/crypto/core.ts
+++ b/src/crypto/core.ts
@@ -127,9 +127,7 @@ export const schnorrVerifyDigest = (
 ): boolean => {
   try {
     const digestBytes = typeof digest === 'string' ? hexToBytes(digest) : digest;
-    // Use X-only pubkey: strip 02/03 prefix if pubkey is 66 hex chars (33 bytes)
-    const pubkeyX = pubkey.length === 66 ? pubkey.slice(2) : pubkey;
-    return schnorr.verify(hexToBytes(signature), digestBytes, hexToBytes(pubkeyX));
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #910 with remaining conflicts!

```bash
git fetch origin backport-904-to-v4-dev
git worktree add --checkout .worktree/backport-904-to-v4-dev backport-904-to-v4-dev
cd .worktree/backport-904-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 673468f4f849855f28a7251734ece72969b1b189
```

PR #905, opened 2026-08-07
fix(utils): align base64 fallback chunk size to a 3-byte boundary

robwoodgate:
The no-Buffer fallback in `Bytes.toBase64` (the path browsers take) chunks input through `btoa` to avoid a `String.fromCharCode` arg-spread stack overflow. The chunk size was 32768, which is not a multiple of 3, so every chunk's output ended in `=` padding and the concatenated result contained mid-string padding for inputs over 32 KiB. That is not valid base64: strict decoders (`atob`, our own `fromBase64` on the same path) throw on it, and lenient decoders (Node `Buffer`) strip the stray `=` and decode to different bytes than were encoded.

Chunk by 32766 (a multiple of 3) instead: every non-final chunk encodes to a clean 4-char block and only the final chunk may pad, which is valid. The spread arg count stays within the previously safe bound.

The new test forces the fallback path by stubbing `Buffer` (a no-op in the browser projects, where it is already absent) and asserts both no mid-string `=` and a strict round-trip back to the input bytes; the existing large-array test only checked `length % 4`, which the invalid output happened to satisfy. `npm run prtasks` green.

### DIFF
--- a/src/utils/Bytes.ts
+++ b/src/utils/Bytes.ts
@@ -76,11 +76,14 @@ export class Bytes {
     if (bufferConstructor) {
       return bufferConstructor.from(bytes).toString('base64');
     }
-    // preventing stack overflow by chunking
-    if (bytes.length > 32768) {
+    // Chunk to avoid a String.fromCharCode arg-spread stack overflow. Chunk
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #911

PR #906, opened 2026-08-07
fix(logger): escape control characters in log messages

robwoodgate:
ConsoleLogger passed the message argument to `console.*` verbatim, so embedded control characters (newlines, ANSI escape sequences) were emitted raw into the log stream. Some in-tree call sites interpolate remote-derived text (eg error messages) into the message position.

Escape control characters at the `header()` chokepoint, which every message passes through: C0 controls, DEL, and the C1 range (some terminals honour 8-bit CSI) render as visible escapes — `\n`, `\r`, `\t` by name, anything else as `\xNN` — so one log call renders as one line and escape sequences arrive defanged but readable. Context objects are unchanged: they already render through the console's own inspection, which quotes control characters in strings.

Behaviour note: a message containing an intentional tab or newline now shows the escape literally; no in-tree call site does that. `npm run prtasks` green.

### DIFF
--- a/src/logger/ConsoleLogger.ts
+++ b/src/logger/ConsoleLogger.ts
@@ -8,6 +8,21 @@ const LEVEL_ORDER: Record<LogLevel, number> = {
   trace: 4,
 };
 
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #912

PR #908, opened 2026-08-07
fix(wallet): reject extraPayload keys reserved by the melt request

robwoodgate:
`CompleteMeltOptions.extraPayload` exists so callers can add method-specific fields to a melt request, but `completeMelt` spread it *after* the prepared `quote`, `inputs`, `outputs` and `prefer_async`. A colliding key therefore silently replaced data taken from the `MeltPreview`, breaking the boundary between the prepared transaction and the request actually sent.

Reject those four keys before the payload is built, using the wallet's usual `failIf` so the caller gets a `CTSError` naming the offending keys and no request is made. Rejecting is preferable to letting the prepared fields win silently: a caller supplying `quote` here either has a bug or is forwarding untrusted input, and both deserve to surface.

The key check uses `Object.keys`, which is exactly what the spread below copies, so the guard cannot drift from the thing it protects. Method-specific fields are unaffected — the in-tree onchain path passes `fee_index` and is covered by a test asserting it still reaches the mint. `npm run prtasks` green.

### DIFF
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -3196,13 +3196,22 @@ class Wallet {
     // Prepare proofs for mint
     inputs = this._prepareInputsForMint(inputs);
 
-    // Construct melt payload
+    // Construct melt payload. Extension fields must not clobber the prepared request.
+    const extra = completeOptions.extraPayload;
+    if (extra) {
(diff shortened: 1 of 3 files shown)


robwoodgate:
- #914

PR #909, opened 2026-08-07
perf(wallet): look up keyset unit by map instead of scanning ids

robwoodgate:
The receive path checked proof keysets by building a keyset id array on every call and running `includes()` inside a `find()`, so cost grew with both proof count and keyset count. Since keyset count only grows as a mint rotates, that shape degrades over time on a wallet doing nothing unusual.

`KeyChain.isUnitKeyset(id)` reads the keyset map that KeyChain already holds, so the check is O(1) per proof, flat in keyset count, and allocates nothing. Measured over 10 passes:

| proofs | keysets | before | after |
| --- | --- | --- | --- |
| 50,000 | 50 | 1.79 ms | 0.37 ms |
| 50,000 | 200 | 5.71 ms | 0.38 ms |

The same check now also runs where proof amounts are totalled or selected:

- `isPaymentRequestSatisfied` — a mint's KeyChain holds keysets for every unit it offers and `getKeyset(id)` is deliberately cross-unit, so amounts from another unit previously counted toward a NUT-18 request total.
- `selectProofsRGLI` and `selectProofsRotating` — a proof from another unit could be selected against a target in the wallet's unit, so `sendOffline` could hand out value in the wrong asset.

The receive-side check keeps its behaviour and now shares one private helper with the settlement path, so the wallet diff removes more than it adds. In the selectors the check sits inside the loops that already run per proof (the fee cache and the bucketing walk), adding no extra pass or allocation to the hot path, and it follows the keyset lookup so an unknown id still reports as unknown rather than as a unit mismatch. Selector test stubs gained the new method rather than the guard being made optional. `npm run prtasks` green.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -752,6 +752,7 @@ export class KeyChain {
     getKeyset(id?: string): Keyset;
     getKeysets(): Keyset[];
     init(forceRefresh?: boolean): Promise<void>;
+    isUnitKeyset(id?: string): boolean;
     loadFromCache(cache: KeyChainCache): void;
     static mintToCacheDTO(mintUrl: string, allKeysets: MintKeyset[], allKeys: MintKeys[]): KeyChainCache;
 }
--- a/src/wallet/KeyChain.ts
+++ b/src/wallet/KeyChain.ts
@@ -298,6 +298,19 @@ export class KeyChain {
   }
 
   /**
+   * True if `id` is a keyset belonging to this KeyChain's unit.
+   *
+   * @remarks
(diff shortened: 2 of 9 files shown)


robwoodgate:
- #915 with remaining conflicts!

```bash
git fetch origin backport-909-to-v4-dev
git worktree add --checkout .worktree/backport-909-to-v4-dev backport-909-to-v4-dev
cd .worktree/backport-909-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x ec6fc8b1d13384d81e36f28db6052e6a2aa05b94
```

PR #907, opened 2026-08-07
fix(utils): cap JSON nesting depth in JSONInt.parse

robwoodgate:
The recursive-descent parser in `JSONInt` had no depth bound: `parseValue` -> `parseObject`/`parseArray` recurse per nesting level, so pathologically nested input (reachable through v3 token decoding and proof-tag deserialization, which call `parse` without a try/catch) exhausted the call stack at around 5k levels instead of failing as a parse error.

Thread a depth counter through the value dispatch and reject past `MAX_PARSE_DEPTH = 64` with a normal `SyntaxError`, mirroring the existing `MAX_CBOR_DEPTH` pattern in the CBOR decoder (same cap, same entry-point check). Real Cashu payloads nest about 5 levels, so the cap is generous. The reviver walk is bounded for free since it only traverses what parse produced; `stringify` operates on the caller's own data and is unchanged.

Tests cover the exact boundary (65 container levels parse, 66 reject) and the pathological case (100k levels now throws `SyntaxError`, not `RangeError`). `npm run prtasks` green.

PR #913, opened 2026-08-07
fix(utils): cap JSON nesting depth in JSONInt.parse [backport v4-dev]

robwoodgate:
# Description
Backport of #907 to `v4-dev`.

### DIFF
--- a/src/utils/JSONInt.ts
+++ b/src/utils/JSONInt.ts
@@ -77,6 +77,10 @@ type ReviverFn = (this: unknown, key: string, value: unknown) => unknown;
 type ReplacerFn = (this: unknown, key: string, value: unknown) => unknown;
 type ReplacerList = ReadonlyArray<string | number>;
 
+// Legitimate Cashu payloads nest a handful of levels; the recursive descent
+// overflows the call stack near ~5k. Mirrors MAX_CBOR_DEPTH in cbor.ts.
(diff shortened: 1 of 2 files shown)


robwoodgate:
```bash
git fetch origin backport-907-to-v4-dev
git worktree add --checkout .worktree/backport-907-to-v4-dev backport-907-to-v4-dev
cd .worktree/backport-907-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x c7b358927ce5dcef17c2f2de2d953840a309732b
```

robwoodgate:
Conflict resolved and force-pushed; `npm run prtasks` green on the v4 lockfile.

Both conflicts came from `MAX_INT_TOKEN_LENGTH`, which sits directly above `MAX_PARSE_DEPTH` on main and so got pulled in as context. v4 has never had that cap and should not gain it here: per #887 it changes the default behaviour of a public utility, which is breaking on the LTS line, and v4 has no `decodePayload` so its only exposure is parsing a semi-trusted mint response.

This backport therefore takes only the depth cap: `MAX_PARSE_DEPTH`, the threading through `parseValue`/`parseObject`/`parseArray`, and the two nesting-depth tests. The `rejects oversized integer tokens` test was dropped with the constant (it would fail on v4).

PR #915, opened 2026-08-07
perf(wallet): look up keyset unit by map instead of scanning ids [backport v4-dev]

robwoodgate:
# Description
Backport of #909 to `v4-dev`.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -663,6 +663,7 @@ export class KeyChain {
     getKeyset(id?: string): Keyset;
     getKeysets(): Keyset[];
     init(forceRefresh?: boolean): Promise<void>;
+    isUnitKeyset(id?: string): boolean;
     loadFromCache(cache: KeyChainCache): void;
     static mintToCacheDTO(mintUrl: string, allKeysets: MintKeyset[], allKeys: MintKeys[]): KeyChainCache;
 }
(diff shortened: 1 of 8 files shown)


robwoodgate:
```bash
git fetch origin backport-909-to-v4-dev
git worktree add --checkout .worktree/backport-909-to-v4-dev backport-909-to-v4-dev
cd .worktree/backport-909-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x ec6fc8b1d13384d81e36f28db6052e6a2aa05b94
```

robwoodgate:
Conflict resolved and force-pushed; `npm run prtasks` green on the v4 lockfile.

All three conflicts were main-only code arriving as diff context. v4 has no `isPaymentRequestSatisfied`, so the settlement guard and its test suite are not carried over: adding that method here would be a v5 feature rather than a fix. An unrelated pre-existing offline-selection test came along with the send-test hunk and was left out too, to keep this to one concern.

What v4 gets is the fix and the speedup:
- `KeyChain.isUnitKeyset(id)`, an O(1) map lookup replacing the per-proof `includes()` scan on the receive path.
- The same check in `selectProofsRGLI` and `selectProofsRotating`, so a proof from another unit of the mint cannot be selected against a target in the wallet's unit.
- `prepareSwapToReceive` keeps its existing behaviour, now via the shared helper.

The direct map lookup is safe here because v4 already has the `Object.create(null)` keyset map from #901; without that a prototype-chain id could slip through.

PR #919, opened 2026-08-07
fix(crypto): bound the deterministic counter to the safe integer range

robwoodgate:
The HMAC branch of the NUT-13 derivation fed the counter straight into a uint64 encoding. `DataView.setBigUint64` applies modulo 2^64 rather than rejecting out-of-range input, so an out-of-range counter aliased onto a valid one and re-derived its secret: `2 ** 64` serialised exactly like `0`, and negatives became valid encodings too. The BIP-32 branch has had a range guard for this reason; the HMAC branch had none.

The cap is `MAX_SAFE_INTEGER`, not 2^64. Above it `counter + 1` no longer produces a distinct value, so a batch of outputs would derive one counter twice and issue duplicate blinded messages rather than independent coins. Rejecting the unsafe range closes both the wrap and the precision case with one bound.

Guarding the deriver covers every caller: batch creation adds `i` to the base counter and funnels through here (`OutputData.ts`), so the creator needs no separate check. Tests cover the wrap value, negatives, non-integers, the precision boundary, and a batch that crosses it, plus the full accepted range at both ends.

The second commit removes the underlying footgun. `Bytes.writeBigUint64BE` existed only to wrap `setBigUint64`, so it silently produced a different value than asked for. noble's `numberToBytesBE` throws on both overflow and negatives, is already imported in this file, and emits identical bytes across the valid u64 range. It had a single caller and was not part of the public API, so it goes with its tests rather than being reimplemented. The explicit guard stays as the contract check with an actionable message; the encoder is now the backstop that makes silent aliasing impossible even if that guard is ever moved.

`npm run prtasks` green.

### DIFF
--- a/src/crypto/NUT13.ts
+++ b/src/crypto/NUT13.ts
@@ -199,16 +199,17 @@ function deriveHmacSecretAndBlindingFactor(
   keysetId: string,
   counter: number,
 ): DerivedSecretAndBlindingFactor {
-  // NUT-13 encodes the counter as u64, but setBigUint64 wraps, so an out-of-range value would
-  // alias onto a valid counter and re-derive its secret. Cap at MAX_SAFE_INTEGER rather than
-  // 2^64: past it `counter + 1 === counter`, so a batch would derive one counter twice.
+  // NUT-13 encodes the counter as u64, but cap at MAX_SAFE_INTEGER: past it
+  // `counter + 1 === counter`, so a batch would derive one counter twice.
   if (!Number.isSafeInteger(counter) || counter < 0) {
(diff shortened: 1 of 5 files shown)


robwoodgate:
- #921 with remaining conflicts!

```bash
git fetch origin backport-919-to-v4-dev
git worktree add --checkout .worktree/backport-919-to-v4-dev backport-919-to-v4-dev
cd .worktree/backport-919-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 54871f6324815b9586e89a8976aa42fb622de2cd
```

PR #920, opened 2026-08-07
fix(crypto): require a 64-byte seed and slim the seed error context

robwoodgate:
Two places where the seed input was looser than the contract.

**`deriveMintBackupKeys` accepted any non-empty seed.** NUT-27 derivation is defined over the 64-byte BIP-39 seed, but only emptiness was rejected. TypeScript types are erased for JavaScript callers, so the length is now checked at runtime like the other public-boundary guards in this file.

**The Wallet constructor kept the rejected value in its error context.** `failIf` forwards that context to the configured logger, so the guard now passes the received type rather than the value. Construction still throws exactly as before; only what reaches the log changes.

Tests cover the accepted length at both edges and the rejected cases, plus an assertion that the constructor's log calls do not contain the supplied value.

Note for the backport: only the constructor half applies to v4-dev, which has no NUT-27 backup helper. The bot's PR will need the `NUT27.ts` hunks dropped.

`npm run prtasks` green.

### DIFF
--- a/src/crypto/NUT27.ts
+++ b/src/crypto/NUT27.ts
@@ -18,6 +18,11 @@ export const MINT_BACKUP_D_TAG = 'mint-list';
 const BACKUP_DOMAIN_SEPARATOR = utf8ToBytes('cashu-mint-backup');
 
 /**
+ * Length of a BIP-39 seed, the only input NUT-27 derivation accepts.
+ */
+const BIP39_SEED_BYTES = 64;
+
+/**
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #922 with remaining conflicts!

```bash
git fetch origin backport-920-to-v4-dev
git worktree add --checkout .worktree/backport-920-to-v4-dev backport-920-to-v4-dev
cd .worktree/backport-920-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 640b8fe03c6be24754ce19805133bebd1151f954
```

PR #921, opened 2026-08-07
fix(crypto): bound the deterministic counter to the safe integer range [backport v4-dev]

robwoodgate:
# Description
Backport of #919 to `v4-dev`.

### DIFF
--- a/src/crypto/NUT13.ts
+++ b/src/crypto/NUT13.ts
@@ -138,6 +138,11 @@ function deriveHmacSecretAndBlindingFactor(
   keysetId: string,
   counter: number,
 ): DerivedSecretAndBlindingFactor {
+  // NUT-13 encodes the counter as u64, but cap at MAX_SAFE_INTEGER: past it
+  // `counter + 1 === counter`, so a batch would derive one counter twice.
+  if (!Number.isSafeInteger(counter) || counter < 0) {
+    throw new CTSError('Counter must be an integer in the range 0 <= counter <= 2^53 - 1');
+  }
   return {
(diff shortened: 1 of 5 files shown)


robwoodgate:
```bash
git fetch origin backport-919-to-v4-dev
git worktree add --checkout .worktree/backport-919-to-v4-dev backport-919-to-v4-dev
cd .worktree/backport-919-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 54871f6324815b9586e89a8976aa42fb622de2cd
```

robwoodgate:
Conflict resolved and force-pushed; `npm run prtasks` green on the v4 lockfile. Squashed the bot's conflict commit into a single clean one.

The conflict was structural, not context-drag: v4 splits the HMAC path into `deriveHmacSecretAndBlindingFactor` (which calls `deriveHmac` twice, once for the secret and once for the blinding factor), where main derives both in one function. The guard goes at the entry point rather than inside `deriveHmac`, so it runs once per derivation instead of once per HMAC, matching main's placement. `deriveHmac` is private and reached only from there, so nothing bypasses it — including the deprecated `deriveSecret` / `deriveBlindingFactor` exports, which route through the same entry.

Everything else applied cleanly: the `Bytes.writeBigUint64BE` removal, the switch to `numberToBytesBE`, and all three test files. v4 was already importing `numberToBytesBE` in this file and is on the same `@noble/curves` major, so the encoder swap needed no other change.

One deliberate difference from main: the comment on the encoder line says "the caller's guard" rather than "the guard above", because here the guard genuinely lives in the calling function.

PR #922, opened 2026-08-07
fix(wallet): slim the seed error context [backport v4-dev]

robwoodgate:
Backport of the constructor half of #920.

The `bip39seed` guard put the received value in its error context, which `failIf` forwards to the configured logger. It now passes the received type instead. Construction still throws exactly as before; only what reaches the log changes.

The NUT-27 seed-length check is not carried over. v4 has no backup helper, so the cherry-pick brought `src/crypto/NUT27.ts` and its tests across as new files, unexported and unreachable on this line. Adding a v5 module is not a backport, so both are dropped and only the constructor change remains.

### DIFF
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -232,12 +232,11 @@ class Wallet {
     this._unit = options?.unit ?? this._unit;
     this._boundKeysetId = options?.keysetId ?? this._boundKeysetId;
     if (options?.bip39seed) {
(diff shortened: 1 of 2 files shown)


robwoodgate:
```bash
git fetch origin backport-920-to-v4-dev
git worktree add --checkout .worktree/backport-920-to-v4-dev backport-920-to-v4-dev
cd .worktree/backport-920-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 640b8fe03c6be24754ce19805133bebd1151f954
```

PR #926, opened 2026-08-08
fix(auth): stop following redirects on token-bearing requests

robwoodgate:
## Description

Requests that carry a `Clear-auth` (CAT) or `Blind-auth` (BAT) header were sent through the ordinary pipeline with the default `redirect: 'follow'`. A trusted endpoint that emits a redirect (an open redirect on the mint frontend, a compromised redirect rule) causes the runtime to replay the same POST, tokens intact, at the redirect target. Token-bearing requests now set `redirect: 'error'`, so a 3xx fails the request instead of forwarding the token.

---

## Changes

- `Mint.requestWithAuth`: sets `redirect: 'error'` only when a BAT or CAT is actually attached; unauthenticated requests are untouched.
- `AuthManager.topUp`: same on the `/v1/auth/blind/mint` call, only when a CAT accompanies it.

## Reviewer Notes

- Reject-all over strip-and-follow or origin-check: auth endpoints have no legitimate reason to redirect, and rejecting also covers same-origin downgrade tricks. `redirect: 'error'` is inert on any non-3xx response, so normal auth requests are unchanged.
- Behaviour change: a mint that legitimately 3xx-redirects a token-bearing endpoint (eg trailing-slash normalisation on a reverse proxy) will now throw. This is engine-uniform (not an FF/WebKit quirk) and the right trade for a token-bearing request.
- Tests assert the option is passed on protected requests and absent on unprotected ones; the default transport honours `redirect`, custom transports (mobile/OHTTP/Tor) must implement it themselves, as with the other anti-fingerprinting options.

### DIFF
--- a/src/auth/AuthManager.ts
+++ b/src/auth/AuthManager.ts
@@ -421,6 +421,8 @@ export class AuthManager implements AuthProvider {
       endpoint: joinUrls(this.mintUrl, '/v1/auth/blind/mint'),
       method: 'POST',
       headers,
+      // A CAT must never be replayed to a redirect target: fail rather than follow.
+      ...(cat ? { redirect: 'error' as const } : {}),
       requestBody: payload,
     });
     if (!Array.isArray(res?.signatures) || res.signatures.length !== outputs.length) {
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #933

PR #927, opened 2026-08-08
fix(auth): drop a CAT refresh that completes after the session changes

robwoodgate:
## Description

`ensureCAT` launches at most one refresh at a time but never bound the in-flight refresh to the token state that started it. If a refresh was already awaiting the OIDC provider when the caller cleared the session via `setCAT(undefined)`, the late response still ran `updateFromOIDC` and repopulated the access/refresh tokens after logout, so the next protected request could authenticate the cleared session. Reachable in apps that keep the `AuthManager` across logout or account switch.

## Changes

- `AuthManager` tracks a `tokenGeneration` counter, bumped by `setCAT` (set or clear).
- `ensureCAT` captures the generation when it starts a refresh and applies `updateFromOIDC` only if the generation still matches; a refresh that finishes after a set/clear is discarded.

## Reviewer Notes

- Bumping on every `setCAT` (not just the clear path) means an explicit new token also wins over a racing refresh, which is the caller's intent.
- A post-clear `ensureCAT` that awaits the stale in-flight promise gets `undefined`: the refresh no longer writes back, and `refreshToken` was cleared, so there is nothing to retry.

### DIFF
--- a/src/auth/AuthManager.ts
+++ b/src/auth/AuthManager.ts
@@ -78,6 +78,9 @@ export class AuthManager implements AuthProvider {
   private info?: MintInfo;
   private lockChain?: Promise<void>;
   private inflightRefresh?: Promise<void>;
+  // Bumped whenever the caller sets or clears the CAT. An in-flight refresh that started under an
+  // older generation must not write its result back (eg a refresh landing after logout).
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #934

PR #929, opened 2026-08-08
fix(auth): coerce the device-flow polling interval to a finite value

robwoodgate:
## Description

`startDeviceAuth` used the device authorization response's `interval` field directly: `Math.max(start.interval ?? 1, intervalSec)`. Because the response is only cast to its type, a provider returning a non-numeric `interval` (e.g. the string `"not-a-number"`) makes `Math.max` yield `NaN`, which flows through to `sleep(NaN)` and schedules an immediate timer. A provider that keeps returning `authorization_pending` then drives a tight polling loop from a single user action.

## Changes

- Coerce `start.interval` with `Number(...)` and fall back to 1 unless it is a finite positive number, before combining it with the caller's `intervalSec`.

## Reviewer Notes

- Only the network-sourced `start.interval` is sanitised; `intervalSec` is our own typed API parameter with a default, not a trust boundary.

### DIFF
--- a/src/auth/OIDCAuth.ts
+++ b/src/auth/OIDCAuth.ts
@@ -270,7 +270,11 @@ export class OIDCAuth {
     }
   > {
     const start = await this.deviceStart();
-    const interval = Math.max(start.interval ?? 1, intervalSec);
+    // Coerce the provider's interval to a number and validate it is finite and positive.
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #936

PR #930, opened 2026-08-08
fix(ws): ignore socket events from a replaced connection

robwoodgate:
## Description

`WSConnection` event handlers mutate shared state (`connectionPromise`, `rpcListeners`, message handling) without checking the event came from the socket that is still current. `close()` clears `this.ws` synchronously, so a following `connect()` can install a replacement before the old socket's `onclose` arrives. When it does, the stale event clears the replacement's connection promise and fails or drops its RPC listeners: a fresh subscription's acknowledgement is suppressed and event delivery can be denied while the replacement appears open. Ordinary network delay can produce the same ordering.

## Changes

- `connect()` captures the socket it creates; `onopen`/`onerror`/`onmessage` no-op unless it is still `this.ws`.
- `onclose` bails only when a replacement socket is already current, so an explicit `close()` (which clears `this.ws`) still runs teardown and notifies `onClose` subscribers.

## Reviewer Notes

- No change for a normal single-connection lifecycle or for custom injected sockets: same socket interface, and the guard is plain object-identity. Only the reconnect-race path changes.
- One deliberate semantic: `onClose` no longer fires for a socket already superseded by a reconnect (it previously could signal "closed" right after a successful reconnect). A live socket closing still notifies.

### DIFF
--- a/src/transport/WSConnection.ts
+++ b/src/transport/WSConnection.ts
@@ -118,23 +118,31 @@ export class WSConnection {
         settle(() => reject(err));
       };
 
+      let socket: WebSocket;
       try {
-        this.ws = new this._WS(this.url.toString());
+        socket = new this._WS(this.url.toString());
+        this.ws = socket;
       } catch (e) {
         fail(e);
         return;
       }
 
+      // A later connect() replaces this.ws with a new socket. This socket's own events must then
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #937

PR #938, opened 2026-08-09
fix(crypto): bound pow2height in createNewMintKeys at runtime

robwoodgate:
## Description

`createNewMintKeys(pow2height: IntRange<0, 65>, ...)` relies only on the erased TypeScript type. A plain-JS caller, JSON-derived value, or cast can pass an arbitrary or infinite count; the loop then performs one key derivation and EC public-key multiply per iteration, so a large or `Infinity` value exhausts CPU/memory. Mint-side, reachable when an admin endpoint, plugin, or config loader forwards an attacker-influenced count.

## Changes

- Reject a non-integer or out-of-`[0, 64]` `pow2height` with a `CTSError` before the loop (the type admits 0..64).

## Reviewer Notes

- no behaviour change for valid callers.

### DIFF
--- a/src/crypto/NUT01.ts
+++ b/src/crypto/NUT01.ts
@@ -71,6 +71,11 @@ export function createNewMintKeys(
     versionByte?: number;
   },
 ): KeysetPair {
+  // The IntRange type is erased at runtime; a plain-JS or JSON-derived caller can pass any number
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #940

PR #939, opened 2026-08-09
fix(auth): detach the OIDC token listener when the provider is replaced

robwoodgate:
## Description

`attachOIDC()` sets `this.oidc` and registers a token listener, but never removes the listener from a previously attached provider. A delayed authorization/device-flow completion from a detached provider therefore still runs `updateFromOIDC`, installing that provider's access and refresh tokens. Because `ensureCAT()` refreshes against the currently attached provider, the old provider's refresh token can then be handed to the new provider (cross-identity credential disclosure), or a stale callback can switch subsequent requests back to the wrong identity.

## Changes

- `AuthManager` keeps a reference to the listener it registers and detaches it from the previous provider before attaching a new one.
- New `OIDCAuth.removeTokenListener(fn)` (by reference) to support this.

## Reviewer Notes

- Distinct from #927 (in-flight refresh resurrecting a cleared session). This is the listener retained across provider replacement.
- After replacement the old provider holds no listener into the manager, so its late deliveries are inert; the current provider still updates state.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1444,6 +1444,7 @@ export class OIDCAuth {
     passwordGrant(username: string, password: string): Promise<TokenResponse>;
     // (undocumented)
     refresh(refresh_token: string): Promise<TokenResponse>;
+    removeTokenListener(fn: (t: TokenResponse) => void | Promise<void>): void;
     // (undocumented)
     setClient(id: string): void;
     // (undocumented)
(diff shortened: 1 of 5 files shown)


robwoodgate:
- #943 with remaining conflicts!

```bash
git fetch origin backport-939-to-v4-dev
git worktree add --checkout .worktree/backport-939-to-v4-dev backport-939-to-v4-dev
cd .worktree/backport-939-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x ae9ea8b2544ce1d07afcccc62af77f4e167f7a1c
```

PR #942, opened 2026-08-09
feat(token): add getTokenMetadata proofAmounts and deprecate incompleteProofs

robwoodgate:
## Description

v4 companion to #941 (which removes `incompleteProofs` on v5). Handing back partial proofs from `getTokenMetadata()`, the pre-wallet inspection helper, was a recurring source of confusion, and the field carried more than inspection needs. This LTS change is additive and non-breaking: it adds `proofAmounts` and marks `incompleteProofs` deprecated rather than removing it.

## Changes

- `TokenMetadata`: new `proofAmounts: Amount[]`; `incompleteProofs` kept and `@deprecated` (removed in v5).
- `getTokenMetadata` populates both fields.
- Tests updated.

## Reviewer Notes

- Non-breaking: existing `incompleteProofs` consumers are unaffected; the deprecation surfaces in editors/tsc via the JSDoc tag.
- To obtain actual proofs, decode with a loaded wallet.

### DIFF
--- a/src/model/types/token.ts
+++ b/src/model/types/token.ts
@@ -81,6 +81,13 @@ export type TokenMetadata = {
   amount: Amount;
   /**
    * The incomplete proofs of the token.
+   *
+   * @deprecated Removed in v5. Use {@link TokenMetadata.proofAmounts} to inspect amounts, or decode
+   *   the token with a loaded wallet to obtain proofs.
    */
   incompleteProofs: Array<Omit<Proof, 'id'>>;
+  /**
+   * The per-proof amounts of the token, in order.
+   */
+  proofAmounts: Amount[];
 };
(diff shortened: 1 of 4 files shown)


src/model/types/token.ts
```diff
@@ -81,6 +81,14 @@ export type TokenMetadata = {
   amount: Amount;
   /**
    * The incomplete proofs of the token.
+   *
+   * @deprecated Returns partial proofs (secrets and signatures included), which a metadata helper
+   *   should not. Use {@link TokenMetadata.proofAmounts} to inspect amounts, or decode the token
+   *   with a loaded wallet to obtain proofs. Removed in v5.
```
robwoodgate:
```suggestion
   * @deprecated Removed in v5. Use {@link TokenMetadata.proofAmounts} to inspect amounts,
   *  or decode the token with a loaded wallet to obtain proofs.
```

PR #946, opened 2026-08-10
fix(p2bk): share one ephemeral key across a blinded SIG_ALL batch

robwoodgate:
## Description

NUT-28 requires a SIG_ALL batch to reuse one ephemeral keypair: NUT-11's equality rule means every proof in the batch must carry identical `data` and `tags`, which only holds if all outputs are blinded with the same `e`. `createP2PKData` was generating a fresh key per split output, so the minted proofs could never be spent together.

## Changes

- `OutputData.createP2PKData` generates one ephemeral key (via the existing `createRandomSecretKey`) when `blindKeys` is set with `sigFlag: 'SIG_ALL'`, and threads it through a new optional `eBytes` param on `createSingleP2PKData`. SIG_INPUTS batches keep per-output keys.
- `DefaultOutputDataCreator.createP2PKData` delegates to the batch path unless the single-output hook is overridden, mirroring the deterministic path's existing pattern.
- Tests cover the shared-key SIG_ALL batch, per-key SIG_INPUTS batch, creator delegation, and subclassed hooks.

## Reviewer Notes

- `createSingleP2PKData` gains an optional trailing param (additive, API report updated).
- Subclasses that override only `createSingleP2PKData` keep the per-output loop; per the `OutputDataCreator` docs, compatibility outside the default implementation is the integrator's responsibility.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -1510,7 +1510,7 @@ export class OutputData implements OutputDataLike {
     // (undocumented)
     static createSingleDeterministicData(amount: AmountLike, seed: Uint8Array, counter: number, keysetId: string): OutputData;
     // (undocumented)
-    static createSingleP2PKData(p2pk: P2PKOptions, amount: AmountLike, keysetId: string): OutputData;
+    static createSingleP2PKData(p2pk: P2PKOptions, amount: AmountLike, keysetId: string, eBytes?: Uint8Array): OutputData;
     // (undocumented)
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #948 with remaining conflicts!

```bash
git fetch origin backport-946-to-v4-dev
git worktree add --checkout .worktree/backport-946-to-v4-dev backport-946-to-v4-dev
cd .worktree/backport-946-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 9189f36f7ac0ed0b8ed834a2b38ea2556f78f131
```

PR #947, opened 2026-08-10
refactor(sigall): recompute signing digests from package contents

robwoodgate:
## Description

The SigAll signing package carried precomputed digests that `signPackage` signed directly. Recomputing them from the package's own inputs, outputs and quote is simpler and more robust: a signer always signs exactly what the package describes, and the transport no longer duplicates data it already contains. With that, the `digests` field and the `validateDigest` option are redundant and removed.

The `sigallA` prefix is kept: a package from an older build still parses, with its `digests` field ignored and stripped, and the fields that matter (inputs, outputs, quote, witness) are unchanged.

## Changes

- `signPackage` recomputes digests from package contents and emits one signature per supported format.
- `SigAllSigningPackage` drops `digests`; `deserializePackage` drops the `validateDigest` option and rebuilds the package from validated fields only. New guards: melt packages require a non-empty `quote` (it is part of the signed transcript), and `witness.signatures` must be a string array.
- Digest names are version-anchored: `SigAllDigests` is now `{ v0 }`, leaving room for the length-framed `v1` format proposed in cashubtc/nuts#404.
- The deprecated pre-v0 message builder is removed; the wallet SIG_ALL path and `computeDigests` now produce the v0 format only. Pre-v0 was only ever accepted by Nutshell <= 0.20.2 and CDK < 0.14.0.
- `buildP2PKSigAllMessage` renamed `buildP2PKSigAllMessageV0` (internal, not in the public rollup).

## Reviewer Notes

- Breaking for the experimental SigAll surface only: `SigAllDigests` shape, `deserializePackage` signature, no digests in emitted packages. Wallets on old mints that only accept the pre-v0 SIG_ALL format will no longer be able to spend SIG_ALL-locked proofs there.
- Net deletion overall; SigAll.ts coverage 98.9% statements / 95.2% branches from its own suite.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -2095,19 +2095,16 @@ export type SigAllApi = {
     extractSwapPackage: (preview: SwapPreview) => SigAllSigningPackage;
     extractMeltPackage: <TQuote extends Pick<MeltQuoteBaseResponse, 'quote'>>(preview: MeltPreview<TQuote>) => SigAllSigningPackage;
     serializePackage: (pkg: SigAllSigningPackage) => string;
-    deserializePackage: (input: string, options?: {
-        validateDigest?: boolean;
-    }) => SigAllSigningPackage;
+    deserializePackage: (input: string) => SigAllSigningPackage;
     signPackage: (pkg: SigAllSigningPackage, privkey: string) => SigAllSigningPackage;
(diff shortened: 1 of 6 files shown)


robwoodgate:
- #951

KvngMikey (commented):
ACK !

PR #964, opened 2026-08-13
perf(wallet): restore only the counters a scan needs

robwoodgate:
Seed restore asks NUT-07 before NUT-09. A state check is the cheaper oracle and the only one that can say "do not bother": once a batch's counters are known SPENT, their signatures are never fetched. On a wallet with ~16k used counters and 29 live proofs that is 4.4s instead of 118.6s, 42 requests instead of 68, and the mint sees 29 blinded messages it recognises instead of 15,816.

## Why

Restore is normally done the other way round, and the reference wallets still do it that way: fetch every signature the mint ever issued, unblind them all, then check which are spent. The work scales with history, and on an old wallet over 99% of it is thrown away.

Asking the state check first inverts that:

- **Work scales with balance, not history.** A fully spent batch costs one derivation and one `Y` per counter and a single request. Signatures are fetched and unblinded for live proofs alone.
- **It runs on every mint deployed today.** No new endpoint, no gap negotiation, no mint-side change. It composes with the efficient-recovery NUT discussion rather than competing with it.
- **The spent history never has both halves revealed.** The mint records `B_` at issuance and `Y` at spend and cannot connect them. The old order handed it both for every spent counter; this order hands it `B_` for the live set only.

|                                            | `B_` recognised | `Y` recognised |
| ------------------------------------------ | --------------- | -------------- |
| `filterSpent: false` (restore everything)   | 15,816          | 15,787         |
| default (state check first)                 | **29**          | 15,787         |

The `Y` column does not move: the scan still walks `[0, T]`, so the mint learns the shape of the history. This changes what is revealed per counter, not how many counters are touched.

## Changes

- `batchRestore` state checks each batch and restores only the counters that are not spent; spent proofs are dropped and pending ones kept. The unfiltered mode is gone with it: `restore(start, count)` stays as the raw NUT-09 replay for audits and comparisons.
- The stop signal is "was any counter in this batch ever issued", proven by a `SPENT` state or a returned signature, since a fully spent batch now restores nothing.
- The scan opens with one probe as wide as the gap limit, so an unused keyset costs two requests, then continues in batches through a pool that doubles per used wave.
- Batch size and pool width default by keyset kind: 500/4 for HMAC, 200/2 for BIP32, 100/2 for BLS, capped at `max_array_length`. Every scanned counter costs a derivation, a `Y` and a blinded message on the JS thread (about 0.1ms, 0.7ms and 1.1ms per kind), so a batch is sized to roughly one round trip of that work, and the dearer kinds keep fewer in flight. Measured against a nutshell BLS mint: a 35-counter BLS keyset recovers in 2.7s against 11.6s with the previous 4-wide pool of 500-batches, an unused BLS keyset in 1.5s and two requests, and an 800-counter HMAC history in 0.6s.
- The scan derives a batch once for the state check and blinds only the non-spent counters, through the wallet's `outputDataCreator`, so a custom crypto backend is honoured in restore as it is in a swap. `checkProofsStates` addresses results by request index. The restore-response mapping moves to `_internal`.
- `examples/restore_example.ts` churns a wallet and recovers it three ways, with `SCENARIO=` presets for the shapes that tune differently (aged mostly spent, aged sparsely spent, newer mostly unspent) and a never-used-seed leg.

## Impact

**Behaviour changes.** Returned proofs and `lastCounterWithSignature` are identical to the v5 default before this change; the request pattern is not, and `filterSpent` is removed rather than kept as a second mode. A loop over `restore()` reproduces the old scan, as the example does.

**The trust surface moves.** A mint falsely reporting `SPENT` now causes the wallet to skip that counter rather than filter a returned proof. That is weaker than what a hostile mint can already do (withhold the signature, refuse the proof later), but it is a new place where correctness depends on the mint's honesty; `restore()` is the route round it.

**It assumes an unknown `Y` reports `UNSPENT`**, which both reference mints do by construction and NUT-07 leaves implicit: of its three states, the other two cannot apply to a proof the mint has never seen.

**The gap rule applies uniformly from counter 0.** The old opening wave speculated four batches wide, so proofs past a longer-than-`gapLimit` hole at the start of a keyset could be rescued by accident. The scan now stops where the documented contract says.

**The next `gapLimit` counters after a restore have been probed**, so their `Y` and `B_` reached the mint before the proofs exist. That is inherent to any gap-rule scan: the wallet must resume inside the gap to stay recoverable, since skipping the probed range would open a hole a later scan stops in. The old path already revealed their `B_`; this adds `Y`. The ramp keeps the overshoot beyond the gap to one wave.


robwoodgate:
Run from the demo in `examples/restore_example.ts`:

```
CHURN_ROUNDS=500 SPEND_FRAC=0.2 npx tsx examples/restore_example.ts

Wallet state: 20890 sats live + 21 sats sent-unclaimed, counters used 0..13022

--- Device lost! Recovering from seed on a fresh wallet ---

filterSpent off     : recovered 20911 sats (T={"0187b2e...6074":13022})
  filterSpent off     : 55 requests in 89.2s (28 restore, 27 checkstate)
  revealed 14000 B_ and 13013 Y; mint recognised 13013 B_ and 12991 Y (26004 linked)
  0187b2e8…: 22 live at counters 6119..13022, so the oldest sits 6903 below T

filterSpent on      : recovered 20911 sats (T={"0187b2e...6074":13022})
  filterSpent on      : 33 requests in 2.5s (5 restore, 28 checkstate)
  revealed 1009 B_ and 14000 Y; mint recognised 22 B_ and 12991 Y (13013 linked)
  0187b2e8…: 22 live at counters 6119..13022, so the oldest sits 6903 below T

filterSpent on @100 : recovered 20911 sats (T={"0187b2e...6074":13022})
  filterSpent on @100 : 145 requests in 2.5s (9 restore, 136 checkstate)
  revealed 609 B_ and 13600 Y; mint recognised 22 B_ and 12991 Y (13013 linked)
  0187b2e8…: 22 live at counters 6119..13022, so the oldest sits 6903 below T

Success: both recovered 20911 sats = 20890 live + 21 unclaimed

```

PR #967, opened 2026-08-14
feat(utils): findCashuPayload to locate tokens and payment requests in text

### LINKED ISSUE #849: feat(utils): findCashuPayload to locate tokens and payment requests in text
## Problem

Wallets receive cashu payloads buried in arbitrary text: chat messages, clipboard blobs, `bitcoin:` URI params, wallet-site URLs like `wallet.example/#token=cashuB...`. Each wallet hand-rolls detection, typically a curated table of known wallet URL prefixes plus a regex, and those tables rot as the ecosystem churns.

cashu-ts already handles the *clean* cases: `getDecodedToken` strips `cashu:`, `cashu://`, and `web+cashu://` before decoding. What is missing is locating a payload inside text that contains other content.

## Proposal

One exported finder:

```ts
export function findCashuPayload(
  text: string,
): { kind: 'token' | 'paymentRequest'; payload: string } | null;
```

- Scan for candidate substrings by prefix, each with its own charset:
  - `cashuA` (V3 token), `cashuB` (V4 token), `creqA` (NUT-18 payment request): base64/base64url charset.
  - `creqb1` / `CREQB1` (payment request, bech32m TLV): bech32 charset, matched case-insensitively since QR alphanumeric mode produces the uppercase form. `PaymentRequest.fromEncodedRequest` already accepts both.
- Decode-to-validate each candidate with the existing decoders; a match that fails to decode is skipped, first valid match wins. Regex alone false-positives; the decoder is the authority.
- Return the validated payload string, not a decoded object. The wallet feeds it to `getDecodedToken` / `PaymentRequest.fromEncodedRequest` as usual, so the finder needs no keyset knowledge and adds no second decode surface.

Scanning for the payload itself subsumes every wrapper form: URI schemes, query params, URL fragments, surrounding prose. No per-wallet knowledge to maintain.

## Robustness

Input is attacker-influenced (pasted from anywhere), so the scan must stay linear:

- Prefix-anchored matches with a single character-class quantifier only. No alternation ambiguity, no nested quantifiers, so the regex cannot backtrack catastrophically.
- A hard cap on decode attempts per call. A failed candidate can contain a later prefix (eg a valid `cashuB` swallowed inside junk that starts `cashuA`), so the scan re-enters at the next prefix occurrence; without a cap, text stuffed with prefixes could force quadratic decode work.

## Non-goals

- **No curated third-party wallet URL table** (`wallet.nutstash.app/#`, `wallet.cashu.me/?token=`, ...). That is ecosystem trivia that churns and rots inside a protocol library; the payload scan covers those forms anyway.
- **No multi-match variant.** Paste flows carry one payload; an array-returning variant can come later if a consumer asks.
- **No clipboard or DOM integration.** Input is a string; how wallets obtain it is their business.

## Placement

`src/utils/core.ts`, beside `getDecodedToken` and the private `removePrefix` it complements.

KvngMikey:
## TL;DR

Adds `findCashuPayload(text)`, which locates the first cashu token or payment request inside arbitrary text and returns the encoded payload verbatim. Wallets can drop their hand-rolled URL-prefix tables and regexes.

Fixes #849

## Why

Wallets receive payloads buried in whatever the user pasted: chat messages, clipboard blobs, `bitcoin:` URI parameters, wallet-site URLs. `getDecodedToken` already handles the clean cases (it strips `cashu:`, `cashu://`, `web+cashu://`), but nothing in the library locates a payload inside text that contains other content. Every wallet ends up maintaining its own curated table of known wallet URL prefixes plus a regex, and those tables rot as the ecosystem churns. Scanning for the payload itself subsumes every wrapper form, so there is no per-wallet knowledge to maintain.

## Changes

Two-stage, exactly as specced in #849:

1. **Candidate scan**: one scanner per prefix (`cashuA`/`cashuB`, `creqA`, `creqb1`), each a literal prefix plus a single character-class quantifier. No alternation, no nested quantifiers, so no catastrophic backtracking. `creqb1` (NUT-26) is matched case-insensitively because QR alphanumeric mode emits the uppercase form.
2. **Decode to validate**: `handleTokens` for tokens, `PaymentRequest.fromEncodedRequest` for requests. A regex alone false-positives; the decoder is the authority. First candidate that decodes, by position, wins.

Notes for review:

- **The charset is base64url only (`[A-Za-z0-9=_-]`), not base64.** This is the one place the implementation departs from the issue text, which says "base64/base64url charset". Including `+` and `/` is actively wrong: a token in a URL path (`example.com/cashuB…/more`) has the trailing segment swallowed into the candidate, and base64 decoding ignores what follows the padding while CBOR ignores trailing bytes, so the candidate **decodes successfully** and the payload comes back as `cashuB…/more`. The decoders validate the prefix, not the boundary. The cashu specs are base64url throughout (NUT-00 tokens, NUT-18 `creqA`), so nothing spec-compliant is lost. Regression test pins all three payload types. Happy to revisit if you want the wider charset back.
- **Re-entry on failed decode resumes at `matchIndex + 1`**, not match end, so a payload swallowed inside a failed greedy span is still found, the `cashuB`-inside-`cashuA` example is a test case.
- **Decode-attempt cap**: `MAX_PAYLOAD_DECODE_ATTEMPTS = 16` in `utils/limits.ts`, following the house pattern there. Bounds the adversarial prefix-stuffed case to linear total work. The number is a suggestion and is a one-integer change if needed.
- **Token validation uses the module-private `handleTokens` via `removePrefix`**: keyset-free, and avoids `getDecodedToken`'s 5.0 `keysetIds` requirement, which the finder has no business needing.
- `CashuPayloadKind` is exported as a named alias for `'token' | 'paymentRequest'` so consumers can type against it. Structurally identical to the inline signature in the issue.
- Non-goals honored: no wallet URL table, no multi-match variant, no clipboard/DOM integration, and the returned payload is never normalized (uppercase `CREQB1…` comes back uppercase — test case).

Docs: new section on the Helpers page; the usage-index Helpers row previously named mint URL normalization as the page's only content, so it is broadened.

## Impact

Purely additive, one new function and one new exported type, no existing behavior touched. The API report diff is those two entries and nothing else.

### DIFF
--- a/docs-src/usage/helpers.md
+++ b/docs-src/usage/helpers.md
@@ -22,14 +22,22 @@ normalizeMintUrl('ftp://mint.example.com'); // throws CTSError
 
 Finds the first token or payment request out of a block of text which could be a chat message, a wallet URL with the token in its fragment, a `bitcoin:` URI carrying a request in a query parameter.
 
-A match is found by prefix and then decoded to verify they are correct, so anything that is not really a payload gets skipped and the scan carries on, because we are looking for a correct payload and not the wrapper. What it finds comes back exactly as it appeared, ready for `getDecodedToken` or `decodePaymentRequest`, or `null` if nothing decodes.
+A match is found by prefix and then decoded to verify they are correct, so anything that is not really a payload gets skipped and the scan carries on, because we are looking for a correct payload and not the wrapper. What it finds comes back as it appeared, apart from bech32m requests, which are lowercased to their canonical form, and you get `null` if nothing decodes.
 
 ```ts
-import { findCashuPayload, getDecodedToken, decodePaymentRequest } from '@cashu/cashu-ts';
+import {
+  findCashuPayload,
+  getTokenMetadata,
+  getDecodedToken,
+  decodePaymentRequest,
+} from '@cashu/cashu-ts';
 
 const found = findCashuPayload(pastedText);
 
 if (found?.kind === 'token') {
+  const meta = getTokenMetadata(found.payload); // mint, unit and amount, no keysets needed
+  // getDecodedToken resolves short keyset IDs against the keysets you pass it, so it throws for a
+  // mint you have not loaded, which is the common case for a stranger's paste.
   const token = getDecodedToken(found.payload, myKeyChain.getAllKeysetIds());
 } else if (found?.kind === 'paymentRequest') {
   const pr = decodePaymentRequest(found.payload);
--- a/docs-src/usage/helpers.md
+++ b/docs-src/usage/helpers.md
@@ -20,12 +20,9 @@ normalizeMintUrl('ftp://mint.example.com'); // throws CTSError
 
 ## `findCashuPayload`
 
-Finds the first cashu token or payment request inside arbitrary text: a chat message, a clipboard
-blob, a `bitcoin:` URI parameter, a wallet URL fragment. Candidates are matched by prefix and then
-decoded to validate, so a prefix that does not decode is skipped and the search continues. The
-payload is returned exactly as it appeared, ready to hand to `getDecodedToken` or
-`decodePaymentRequest`. Scanning for the payload itself covers the wrapper forms, so there is no
-table of known wallet URL prefixes to maintain.
+Finds the first token or payment request out of a block of text which could be a chat message, a wallet URL with the token in its fragment, a `bitcoin:` URI carrying a request in a query parameter.
+
+A match is found by prefix and then decoded to verify they are correct, so anything that is not really a payload gets skipped and the scan carries on, because we are looking for a correct payload and not the wrapper. What it finds comes back exactly as it appeared, ready for `getDecodedToken` or `decodePaymentRequest`, or `null` if nothing decodes.
 
 ```ts
 import { findCashuPayload, getDecodedToken, decodePaymentRequest } from '@cashu/cashu-ts';
@@ -38,5 +35,3 @@ if (found?.kind === 'token') {
   const pr = decodePaymentRequest(found.payload);
 }
 ```
-
-Returns `null` when the text carries no payload that decodes.
--- a/docs-src/usage/helpers.md
+++ b/docs-src/usage/helpers.md
@@ -17,3 +17,26 @@ import { normalizeMintUrl } from '@cashu/cashu-ts';
 normalizeMintUrl('https://Mint.Example.COM/'); // 'https://mint.example.com'
 normalizeMintUrl('ftp://mint.example.com'); // throws CTSError
 ```
+
+## `findCashuPayload`
+
+Finds the first cashu token or payment request inside arbitrary text: a chat message, a clipboard
+blob, a `bitcoin:` URI parameter, a wallet URL fragment. Candidates are matched by prefix and then
+decoded to validate, so a prefix that does not decode is skipped and the search continues. The
+payload is returned exactly as it appeared, ready to hand to `getDecodedToken` or
+`decodePaymentRequest`. Scanning for the payload itself covers the wrapper forms, so there is no
+table of known wallet URL prefixes to maintain.
+
+```ts
+import { findCashuPayload, getDecodedToken, decodePaymentRequest } from '@cashu/cashu-ts';
+
+const found = findCashuPayload(pastedText);
+
+if (found?.kind === 'token') {
+  const token = getDecodedToken(found.payload, myKeyChain.getAllKeysetIds());
+} else if (found?.kind === 'paymentRequest') {
+  const pr = decodePaymentRequest(found.payload);
+}
+```
+
+Returns `null` when the text carries no payload that decodes.
--- a/docs-src/usage/usage_index.md
+++ b/docs-src/usage/usage_index.md
@@ -39,7 +39,7 @@ If you are building a wallet integration from scratch, read these in order:
 | [Logging](./logging.md)                             | Enable and route library logs while debugging wallet or mint behavior.          |
 | [Amounts](./amounts.md)                             | Work with the `Amount` and `AmountWithUnit` value objects.                      |
 | [Fees](./fees.md)                                   | Pick the right fee helper: input fees, sender-pays-fees, send-max, NUT-18.      |
-| [Helpers](./helpers.md)                             | Standalone public utility functions — mint URL normalization, token detection.  |
+| [Helpers](./helpers.md)                             | Standalone helpers: normalize mint URLs, find tokens and payment requests.      |
 
 ## Related docs
 
--- a/docs-src/usage/usage_index.md
+++ b/docs-src/usage/usage_index.md
@@ -39,7 +39,7 @@ If you are building a wallet integration from scratch, read these in order:
 | [Logging](./logging.md)                             | Enable and route library logs while debugging wallet or mint behavior.          |
 | [Amounts](./amounts.md)                             | Work with the `Amount` and `AmountWithUnit` value objects.                      |
 | [Fees](./fees.md)                                   | Pick the right fee helper: input fees, sender-pays-fees, send-max, NUT-18.      |
-| [Helpers](./helpers.md)                             | Standalone public utility functions — currently mint URL normalization.         |
+| [Helpers](./helpers.md)                             | Standalone public utility functions — mint URL normalization, token detection.  |
 
 ## Related docs
 
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -33,7 +33,7 @@ import { encodeBase64ToJson, encodeBase64toUint8, encodeUint8toBase64Url } from 
 import { Bytes } from './Bytes';
 import { decodeCBOR, encodeCBOR } from './cbor';
 import { JSONInt } from './JSONInt';
-import { MAX_PAYLOAD_DECODE_ATTEMPTS, MAX_SPLIT_OUTPUTS } from './limits';
+import { MAX_PAYLOAD_DECODE_ATTEMPTS, MAX_PAYLOAD_LENGTH, MAX_SPLIT_OUTPUTS } from './limits';
 
 /**
  * Splits the amount into denominations of the provided keyset.
@@ -390,23 +390,32 @@ export function getTokenMetadata(token: string): TokenMetadata {
 export type CashuPayloadKind = 'token' | 'paymentRequest';
 
 /**
- * One scanner per payload prefix: a literal prefix plus a single character class, so matching
- * cannot backtrack catastrophically. The base64 class spans both alphabets so it never truncates a
- * real payload, the decoder decides validity. `creqb1` (NUT-26) is matched case-insensitively
- * because QR alphanumeric mode emits the uppercase form.
+ * One scanner per payload prefix: a literal prefix plus a single character-class quantifier, so
+ * matching cannot backtrack catastrophically, bounded by {@link MAX_PAYLOAD_LENGTH} so a hostile run
+ * cannot produce one huge match for the decoders to chew on. Tokens and `creqA` use the base64url
+ * class only, `+` and `/` are excluded deliberately, since `/` lets a token in a URL path swallow
+ * the following segment and the decoders accept the trailing junk rather than reject it. `creqb1`
+ * (NUT-26) uses the bech32 class and is matched case-insensitively because QR alphanumeric mode
+ * emits the uppercase form.
  */
 const PAYLOAD_SCANNERS: ReadonlyArray<{ regExp: RegExp; kind: CashuPayloadKind }> = [
-  { regExp: /cashu[AB][A-Za-z0-9=_-]+/g, kind: 'token' },
-  { regExp: /creqA[A-Za-z0-9=_-]+/g, kind: 'paymentRequest' },
-  { regExp: /creqb1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]+/gi, kind: 'paymentRequest' },
+  { regExp: new RegExp(`cashu[AB][A-Za-z0-9=_-]{1,${MAX_PAYLOAD_LENGTH}}`, 'g'), kind: 'token' },
+  {
+    regExp: new RegExp(`creqA[A-Za-z0-9=_-]{1,${MAX_PAYLOAD_LENGTH}}`, 'g'),
+    kind: 'paymentRequest',
+  },
+  {
+    regExp: new RegExp(`creqb1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{1,${MAX_PAYLOAD_LENGTH}}`, 'gi'),
+    kind: 'paymentRequest',
+  },
 ];
 
 /**
- * Finds the first cashu token (v3 or v4) or payment request (`creqA` NUT-18, `creqB` NUT-26)
- * embedded in arbitrary text: chat messages, clipboard blobs, `bitcoin:` URI parameters, wallet URL
- * fragments. Candidates are matched by prefix and then decoded to validate, so a prefix that does
- * not decode is skipped. The match is returned verbatim, ready for {@link getDecodedToken} or
- * `PaymentRequest.fromEncodedRequest`.
+ * Finds the first cashu token (v3 or v4) or payment request (`creqA` NUT-18, `creqb1` / `CREQB1`
+ * NUT-26) embedded in arbitrary text: chat messages, clipboard blobs, `bitcoin:` URI parameters,
+ * wallet URL fragments. Candidates are matched by prefix and then decoded to validate, so a prefix
+ * that does not decode is skipped. The match is returned verbatim, ready for {@link getDecodedToken}
+ * or `PaymentRequest.fromEncodedRequest`.
  *
  * @example
  *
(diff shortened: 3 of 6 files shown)


src/utils/core.ts
```diff
@@ -384,6 +384,68 @@ export function getTokenMetadata(token: string): TokenMetadata {
...
+/**
+ * One scanner per payload prefix: a literal prefix plus a single character class, so matching
+ * cannot backtrack catastrophically. The base64 class spans both alphabets so it never truncates a
+ * real payload, the decoder decides validity. `creqb1` (NUT-26) is matched case-insensitively
+ * because QR alphanumeric mode emits the uppercase form.
+ */
+const PAYLOAD_SCANNERS: ReadonlyArray<{ regExp: RegExp; kind: CashuPayloadKind }> = [
+  { regExp: /cashu[AB][A-Za-z0-9=_-]+/g, kind: 'token' },
+  { regExp: /creqA[A-Za-z0-9=_-]+/g, kind: 'paymentRequest' },
+  { regExp: /creqb1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]+/gi, kind: 'paymentRequest' },
```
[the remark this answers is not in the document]
KvngMikey: handled.

docs-src/usage/helpers.md
robwoodgate, line 1: Can we wrap the prose naturally please - this has AI fixed width hard line breaks, which don't flow, eg:

```
... a chat message, a clipboard
blob, ...
```

KvngMikey: not a problem.

robwoodgate: Thank you sir! I know some of the other docs are still afflicted with these, but good not to add more :)

KvngMikey: I'll go through everything again and make necessary updates in a min.

robwoodgate:
Thanks, this is a useful addition. I ran an adversarial pass with executed probes rather than just reading; one structural note first, then items to fix in this PR.

**Decoder strictness (#986)**

Candidate validation inherits leniency from the underlying decoders, so trailing same-charset junk was either included in the returned payload (v4/creqA), missed entirely (v3), or platform-dependent (Node `Buffer` vs browser `atob`). #986 is queued ahead of this PR and makes junk-suffixed candidates fail uniformly on both platforms. Once it merges, please rebase, then revert the `/` and `+` charset carve-out and its justifying comment: the plain base64url class works after that. A junk-suffix regression test (eg `` `${V4_TOKEN}abc` `` returns null) is worth adding at the same time.

**Behavior**

- The `i` flag on the creqb1 scanner accepts spec-invalid mixed-case bech32 (`Creqb1...` from mobile autocapitalize) and returns it verbatim; strict BIP-173 decoders downstream will reject it. Match all-lower/all-upper explicitly and return lowercase.
- `text` is not runtime-guarded, so `exec(null)` scans the string "null". Public API arguments should fail fast per the contributing guide.

**Limits**

- `MAX_PAYLOAD_LENGTH` (256 KiB): a valid 3,000-proof v4 token encodes to ~432k chars and decodes fine via `getDecodedToken`, but is invisible to `findCashuPayload`. Either raise the cap or correct the "clears any real payload" comment and document the limit.
- The "bounds a single candidate at MAX_PAYLOAD_LENGTH" test passes with the bound removed (a single character class cannot backtrack, so the hostile input matches in ~1ms either way). Assert the truncated match length instead of null plus timing.

**Docs**

- helpers.md example: `getDecodedToken(found.payload, keysetIds)` throws for v4 tokens with modern short keyset IDs from any mint the caller has not loaded, which is the typical paste. Suggest `getTokenMetadata` first.
- TSDoc: multi-entry v3 tokens are silently skipped, and null is also returned when 16 failed candidates precede a valid payload; both deserve a line on the public docblock.
- The usage index row says "token detection" only; the function also finds payment requests.

**Nit**

- The V3/V4 token fixtures are re-declared in the new describe block; the file already carries copies (#959 cleaned up this class of drift).

src/utils/core.ts
```diff
@@ -384,6 +384,77 @@ export function getTokenMetadata(token: string): TokenMetadata {
...
+ * What {@link findCashuPayload} located: a cashu token, or a NUT-18 / NUT-26 payment request.
+ */
+export type CashuPayloadKind = 'token' | 'paymentRequest';
+
+/**
+ * One scanner per payload prefix: a literal prefix plus a single character-class quantifier, so
+ * matching cannot backtrack catastrophically, bounded by {@link MAX_PAYLOAD_LENGTH} so a hostile run
+ * cannot produce one huge match for the decoders to chew on. Tokens and `creqA` use the base64url
+ * class only, `+` and `/` are excluded deliberately, since `/` lets a token in a URL path swallow
+ * the following segment and the decoders accept the trailing junk rather than reject it. `creqb1`
```
robwoodgate: Comment nit:

After the #986 rebase this justification stops being true: the decoders will reject trailing junk. The class itself is right, but for a stronger reason than the comment gives, so please rewrite the comment rather than widen it: this is simply the base64url alphabet, which NUT-00 mandates for these payloads.

The two characters that distinguish it from standard base64 are delimiters in the contexts people paste tokens (`/` in URL paths, `+` as a space in query strings), so stopping at them keeps an embedded payload findable rather than swallowing its surroundings.

src/utils/core.ts
```diff
@@ -384,6 +384,77 @@ export function getTokenMetadata(token: string): TokenMetadata {
...
+    if (earliestMatch === null) {
+      return null;
+    }
+    try {
+      if (earliestMatch.kind === 'token') {
+        handleTokens(removePrefix(earliestMatch.text));
+      } else {
+        PaymentRequest.fromEncodedRequest(earliestMatch.text);
+      }
+      return { kind: earliestMatch.kind, payload: earliestMatch.text };
```
robwoodgate: The creqb1 match is returned verbatim, so a mixed-case paste (`Creqb1...` from mobile autocapitalize) comes back as spec-invalid bech32 that strict BIP-173 decoders elsewhere will reject. Case carries no data in bech32 and `fromEncodedRequest` already lowercases before decoding, so keep the tolerant match but normalize on return (the docblock's "returned verbatim" wants the same tweak):
```suggestion
      const payload = /^creqb1/i.test(earliestMatch.text)
        ? earliestMatch.text.toLowerCase()
        : earliestMatch.text;
      return { kind: earliestMatch.kind, payload };
```

robwoodgate (commented):
Adversarial pass with executed probes; details inline (this supersedes my earlier summary comment, now removed). Sequencing note: #986 is queued ahead of this PR and makes junk-suffixed candidates fail uniformly on both platforms. Previously they were returned with the junk included (v4/creqA), missed entirely (v3), or platform-dependent (Node Buffer vs browser atob). Please rebase once it merges.

src/utils/core.ts
```diff
@@ -384,6 +384,77 @@ export function getTokenMetadata(token: string): TokenMetadata {
...
+ *
+ * @example
+ *
+ *     findCashuPayload('paying you back cashuBo2Ft… thanks!');
+ *     // { kind: 'token', payload: 'cashuBo2Ft…' }
+ *
+ * @returns The first valid payload by position, or `null` if the text carries none.
+ */
+export function findCashuPayload(text: string): { kind: CashuPayloadKind; payload: string } | null {
+  let searchFrom = 0;
```
robwoodgate, line 434: Public API: guard the argument at runtime (types are erased at the boundary). `exec` coerces, so a plain-JS caller passing `null` scans the string "null" and gets a meaningless result instead of failing fast:
```suggestion
  if (typeof text !== 'string') {
    throw new CTSError('text must be a string');
  }
  let searchFrom = 0;
```

src/utils/core.ts
```diff
@@ -384,6 +384,77 @@ export function getTokenMetadata(token: string): TokenMetadata {
...
+ * wallet URL fragments. Candidates are matched by prefix and then decoded to validate, so a prefix
+ * that does not decode is skipped. The match is returned verbatim, ready for {@link getDecodedToken}
+ * or `PaymentRequest.fromEncodedRequest`.
+ *
+ * @example
+ *
+ *     findCashuPayload('paying you back cashuBo2Ft… thanks!');
+ *     // { kind: 'token', payload: 'cashuBo2Ft…' }
+ *
+ * @returns The first valid payload by position, or `null` if the text carries none.
```
robwoodgate: Comment nit:

`null` has two more causes than "carries none": a valid payload beyond `MAX_PAYLOAD_DECODE_ATTEMPTS` failed candidates (your decoy test proves this), and a multi-entry v3 token, which the decoder rejects and the catch swallows. Both deserve a mention here:
```suggestion
 * @returns The first valid payload by position, or `null` if the text carries none, a valid
 *   payload sits beyond {@link MAX_PAYLOAD_DECODE_ATTEMPTS} failed candidates, or the only
 *   candidate is a multi-entry v3 token (unsupported).
```

src/utils/limits.ts
```diff
@@ -49,3 +49,20 @@ export const MAX_MINT_INFO_LIST = 1_024;
...
+ * and a few near-misses, so 16 clears them; scans that exhaust the budget return null.
+ */
+export const MAX_PAYLOAD_DECODE_ATTEMPTS = 16;
+
+/**
+ * Upper bound on the encoded payload a single `findCashuPayload` candidate may span. The scan
+ * quantifiers are bounded by it, so a hostile paste of millions of charset characters yields a
+ * capped candidate instead of one huge match handed to the base64/CBOR decoders, the attempt cap
+ * alone bounds how many candidates are tried, not the cost of each. 256 KiB is ~90x the largest
+ * token in the test tree and holds roughly a thousand proofs, so it clears any real payload; a
```
robwoodgate: There is no length cap on the decode path itself, which creates an odd middle ground: a ~432k-char 3,000-proof consolidation token decodes fine when passed directly to `getDecodedToken` but is invisible to `findCashuPayload`.

The transport layer's yardstick for absurd size is 8 MiB (`DEFAULT_MAX_RESPONSE_BYTES`, ">10x any realistic mint response"), so suggest raising this to `8_388_608` to match: at ~57k minimal proofs it genuinely clears any real payload, and worst-case scan work stays bounded (16 attempts over fast-failing decodes).

A follow-up can apply the same constant as an input guard on the decode side, so the two agree by construction.

test/utils/core.test.ts
```diff
@@ -371,6 +373,124 @@ describe('test getTokenMetadata', () => {
...
+  test('returns null when there is no payload', () => {
+    expect(findCashuPayload('')).toBeNull();
+    expect(findCashuPayload('just a normal sentence about cashu wallets')).toBeNull();
+  });
+
+  test('does not match an uppercased token prefix', () => {
+    expect(findCashuPayload(`CASHUB${V4_TOKEN.slice('cashuB'.length)}`)).toBeNull();
+  });
+
+  test('bounds a single candidate at MAX_PAYLOAD_LENGTH', () => {
```
robwoodgate, line 481: This test passes with the bound removed: a single character class cannot backtrack, so the hostile input matches in ~1ms either way, burns one attempt, and returns null. Deleting `MAX_PAYLOAD_LENGTH` would go undetected. Pin the cap directly instead, eg build a valid payload just over the limit and assert it is missed while the same payload under the limit is found.

test/utils/core.test.ts
```diff
@@ -371,6 +373,124 @@ describe('test getTokenMetadata', () => {
   });
 });
 
+describe('findCashuPayload', () => {
+  // Payload fixtures reused from the decode tests above and from paymentRequests.test.ts.
+  const V3_TOKEN =
```
robwoodgate: Nit: these literals already appear elsewhere in this file (the v4 token twice more, the v3 token once). Hoisting shared consts stops them drifting apart; #959 just cleaned up this class of problem.

docs-src/usage/helpers.md
```diff
@@ -17,3 +17,21 @@ import { normalizeMintUrl } from '@cashu/cashu-ts';
...
+
+A match is found by prefix and then decoded to verify they are correct, so anything that is not really a payload gets skipped and the scan carries on, because we are looking for a correct payload and not the wrapper. What it finds comes back exactly as it appeared, ready for `getDecodedToken` or `decodePaymentRequest`, or `null` if nothing decodes.
+
+```ts
+import { findCashuPayload, getDecodedToken, decodePaymentRequest } from '@cashu/cashu-ts';
+
+const found = findCashuPayload(pastedText);
+
+if (found?.kind === 'token') {
+  const token = getDecodedToken(found.payload, myKeyChain.getAllKeysetIds());
```
robwoodgate, line 41: Documentation nit:

The call is correct as written, but only for tokens from mints already in `myKeyChain`: the keyset list can only resolve short keyset IDs it actually contains. The typical `findCashuPayload` catch is a stranger's paste from a mint the reader has not loaded, and there this same call throws in the short-ID mapping. 

Worth showing `getTokenMetadata` first (needs no keysets) and noting the constraint; add it to the import above:
```suggestion
  const meta = getTokenMetadata(found.payload); // mint/unit/amount, no keysets needed
  // getDecodedToken resolves short keyset IDs against your keysets, so it throws for mints you have not loaded
  const token = getDecodedToken(found.payload, myKeyChain.getAllKeysetIds());
```

docs-src/usage/usage_index.md
```diff
@@ -38,7 +38,7 @@ If you are building a wallet integration from scratch, read these in order:
 | [Logging](./logging.md)                             | Enable and route library logs while debugging wallet or mint behavior.          |
 | [Amounts](./amounts.md)                             | Work with the `Amount` and `AmountWithUnit` value objects.                      |
 | [Fees](./fees.md)                                   | Pick the right fee helper: input fees, sender-pays-fees, send-max, NUT-18.      |
-| [Helpers](./helpers.md)                             | Standalone public utility functions — currently mint URL normalization.         |
+| [Helpers](./helpers.md)                             | Standalone public utility functions — mint URL normalization, token detection.  |
```
robwoodgate: The row undersells it; payment request detection is the half people will search for:
```suggestion
| [Helpers](./helpers.md)                             | Standalone public utility functions — mint URL normalization, token and payment request detection. |
```

src/utils/core.ts
```diff
@@ -384,6 +384,77 @@ export function getTokenMetadata(token: string): TokenMetadata {
...
+export function findCashuPayload(text: string): { kind: CashuPayloadKind; payload: string } | null {
+  let searchFrom = 0;
+  for (let attempts = 0; attempts < MAX_PAYLOAD_DECODE_ATTEMPTS; attempts++) {
+    let earliestMatch: { index: number; text: string; kind: CashuPayloadKind } | null = null;
+    for (const scanner of PAYLOAD_SCANNERS) {
+      // Global regexes resume from lastIndex, so point each one at the current search position.
+      scanner.regExp.lastIndex = searchFrom;
+      const match = scanner.regExp.exec(text);
+      if (match && (earliestMatch === null || match.index < earliestMatch.index)) {
+        earliestMatch = { index: match.index, text: match[0], kind: scanner.kind };
```
robwoodgate: A mixed-case creqb1 paste (`Creqb1...` from mobile autocapitalize) is matched tolerantly but returned verbatim, handing callers spec-invalid bech32 that strict BIP-173 decoders elsewhere will reject. Case carries no data in bech32 and `fromEncodedRequest` lowercases before decoding anyway, so canonicalize at capture: a scanner that matches case-insensitively should emit lowercase. The docblock's "returned verbatim" wants the same tweak.
```suggestion
        earliestMatch = {
          index: match.index,
          // A case-insensitive scanner emits canonical lowercase (bech32 case carries no data).
          text: scanner.regExp.flags.includes('i') ? match[0].toLowerCase() : match[0],
          kind: scanner.kind,
        };
```

src/utils/core.ts
robwoodgate: A junk-suffix regression test is worth adding at the same time, eg `expect(findCashuPayload(V4_TOKEN + 'abc')).toBeNull()`.

src/utils/limits.ts
KvngMikey: Good catch but raising the cap outright would've traded a slow bug for a crashable one, since V8's bounded quantifiers recurse with match length, not the bound, and exec() throws outside our try/catch. 
1 MiB makes sense as it has solid headroom over the real 432k case, cheap worst-case cost, and safely below the 4–6M crash zone. Appreciate you documenting the ceiling in the comment.

src/utils/core.ts
KvngMikey: both to be added when #986 is merged.

robwoodgate: Thanks for taking care of all this @KvngMikey - the PR 986 is merged.

KvngMikey: done now!

robwoodgate (approved):
Awesome job, great utility, thanks @KvngMikey .

This will be v5 only, as it may change decoding experience on some platforms, throwing cases that previously passed etc.

KvngMikey:
> Awesome job, great utility, thanks @KvngMikey .
> 
> This will be v5 only, as it may change decoding experience on some platforms, throwing cases that previously passed etc.

no problem, as soon as V5 lands, i can make the downstream PRs to capture this.

PR #969, opened 2026-08-14
fix(wallet): allow minting paid expired quotes

Egge21M:
# NUT

- https://github.com/cashubtc/nuts/pull/427

## What changed

- Stop rejecting mint quotes solely because their payment-request expiry is in the past.
- Keep enforcing the quote's available balance (`amount_paid - amount_issued`).
- Clarify the `expiry` field documentation.
- Replace the old expiry rejection tests with a regression covering an expired quote with mintable balance.

## Why

NUT-23 defines `expiry` as the deadline for paying the BOLT11 request, not a deadline for issuing ecash after payment. NUT-04 defines the currently mintable amount as `amount_paid - amount_issued`.

The previous wall-clock guard in `Wallet.validateMintQuote` rejected already-paid quotes after their invoice expiry, even when the quote still had mintable balance.

Legacy `state: "PAID"` responses remain supported: quote normalization derives `amount_paid` from `amount` and sets `amount_issued` to zero.

## Impact

Wallets can mint remaining paid balance after the payment request has expired. Requests exceeding the remaining balance are still rejected locally.

## Validation

- `npx vitest run --project node` — 2,389 tests passed
- Focused wallet and mint suites — 238 tests passed
- `npm run check-lint` — passed
- `npm run check-format` — passed
- `npm run api:update` — passed

`npm run check-types` is currently blocked by unrelated existing test mocks that do not implement the newly typed `fetch.preconnect` property. Browser tests could not start on the development host because Playwright system libraries are missing.

### DIFF
--- a/src/model/types/NUT04.ts
+++ b/src/model/types/NUT04.ts
@@ -68,7 +68,8 @@ export type MintQuoteBaseResponse = {
    */
   updated_at: number | null;
   /**
-   * Timestamp of when the quote expires. `null` when the mint does not set an expiry.
+   * Timestamp until which the payment request can be paid. `null` when the mint does not set an
+   * expiry.
    */
   expiry: number | null;
   /**
--- a/src/wallet/Wallet.ts
+++ b/src/wallet/Wallet.ts
@@ -2281,7 +2281,7 @@ class Wallet {
   /**
    * @internal
    */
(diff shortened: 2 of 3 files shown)


robwoodgate:
The preconnect error is likely Bun’s fetch typings leaking in, playwright has a script install. Type checks ok 🫡

robwoodgate (approved):
ACK - thank you!

PR #986, opened 2026-08-18
fix(utils): make base64 and CBOR payload decoding consistent across platforms

robwoodgate:
## TLDR

Base64 and CBOR payload decoding now gives the same verdict on Node and in browsers: well-formed input decodes, malformed input throws `CTSError`.

## Why

The two platform decoders (`Buffer`, `atob`) disagree on some edge cases, so the same string could decode on one platform and fail on the other. #967 validates scanned text candidates by decoding them, so verdicts must be uniform across platforms; this PR is sequenced ahead of it.

## Changes

- `Bytes.fromBase64` normalizes (trim, whitespace strip, base64url mapping, padding) and validates in one place, then decodes the validated string, so both backends see identical input.
- `decodeCBOR` requires a payload to consume the whole buffer.
- `encodeBase64ToJson` delegates to `Bytes.fromBase64` directly; the redundant private pre-mapping helper is removed and the base64url output mapping has a single implementation.
- Tests run a shared accept/reject corpus under both decoder backends.

## Impact

`getDecodedToken`, `getDecodedTokenBinary` and `PaymentRequest.fromEncodedRequest` throw `CTSError` for malformed payloads on every platform. Line-wrapped, unpadded, and edge-whitespace input still decodes. Full suite and prtasks pass. v5 only, as it may now throw on platforms that once passed.

### DIFF
--- a/src/utils/Bytes.ts
+++ b/src/utils/Bytes.ts
@@ -86,17 +86,24 @@ export class Bytes {
   }
 
   static fromBase64(base64: string): Uint8Array {
-    base64 = base64.trim();
-    // normalize base64url to base64 and pad
-    let normalizedBase64 = base64.replace(/-/g, '+').replace(/_/g, '/');
-    while (normalizedBase64.length % 4) {
-      normalizedBase64 += '=';
+    // Normalize: trim, strip ASCII whitespace (line-wrapped input is common),
+    // map base64url to base64, drop trailing padding.
+    const unpadded = base64
+      .trim()
+      .replace(/[\t\n\f\r ]+/g, '')
+      .replace(/-/g, '+')
+      .replace(/_/g, '/')
+      .replace(/={1,2}$/, '');
+    if (/[^A-Za-z0-9+/]/.test(unpadded) || unpadded.length % 4 === 1) {
+      throw new CTSError('Invalid base64 string');
     }
+    // Re-pad canonically so both decoder backends see the same validated string.
+    const normalized = unpadded + '='.repeat((4 - (unpadded.length % 4)) % 4);
     const bufferConstructor = getBufferConstructor();
     if (bufferConstructor) {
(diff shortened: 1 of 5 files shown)


robwoodgate:
@KvngMikey - related to your work in #967, would value your thoughts.

KvngMikey:
Post-merge follow-up:

- This one is adjacent to this change rather than a problem with it. The new validation covers charset, padding and length, but not the unused low bits of the final base64 character. On the v4 fixture the last char carries two spare bits, so ...Q, ...R, ...S and ...T all decode to byte-identical proofs, and all four pass getTokenMetadata. CBOR can't catch it — the bytes genuinely are the same, so the trailing-bytes check doesn't see it either.
Not a security issue as far as I can tell, but it means the token string isn't a stable identity: anything keyed on the string rather than the decoded proofs (dedup, a spent-token cache, "already claimed" checks) sees four keys for one token. It also lands on #967, since findCashuPayload returns the matched substring verbatim.
One thing I noticed while digging: isBase64String already canonicalizes via re-encode-and-compare, so keyset IDs were never exposed to this, it's only the token/PR decode paths that have no canonical form check. If it's ever worth closing, that same re-encode-and-compare (or just masking the spare bits of the final char against the expected value) is the fix, and it could live behind the scanner/validator rather than the hot decode path. Happy to open a PR if you think it's worth tracking.

- Small one for the release notes: the CBOR trailing-bytes check also reaches getDecodedTokenBinary, since it feeds bytes.slice(5) straight into decodeCBOR. A craw token sitting in a padded or fixed-size buffer (BLE frame, QR byte-mode segment, that kind of thing) used to decode the prefix and now throws "CBOR data has trailing bytes".
I think strict is the right default here, but the Impact section reads as "same behavior, now consistent across platforms", and this is a new failure mode for previously-decodable input, so anyone scanning the notes to work out if they're affected would miss it. Might be worth a line in the v5 notes even if the PR body stays as-is.

PR #998, opened 2026-08-26
test: pin the NUT-13 V2 secret derivation vectors

robwoodgate:
Pins counters 0-4 of [nuts/tests/13-tests.md "Version 2: Secret derivation"](https://github.com/cashubtc/nuts/blob/main/tests/13-tests.md#version-2-secret-derivation) against `deriveSecretAndBlindingFactor`, alongside the existing V3 lock-in test.

The V1 and V3 vector sets were already exercised, but the V2 set never was, so a regression in the V2 message construction would not have failed a test. The spec's mnemonic is pre-derived to its BIP39 seed, keeping the test in seed space like its neighbours (no bip39 dependency).

All vectors pass against current main.

### DIFF
--- a/test/crypto/NUT13.test.ts
+++ b/test/crypto/NUT13.test.ts
@@ -59,6 +59,39 @@ describe('v3 (BLS) derivation', () => {
   });
 });
 
+describe('v2 derivation spec vectors', () => {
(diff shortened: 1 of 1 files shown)


robwoodgate:
- #999

PR #1001, opened 2026-08-27
fix(wallet): keep large proofs in the close-match selection pool

robwoodgate:
Proof selection can return an empty send on a fee-charging mint even when the wallet holds ample funds.

`selectProofsRGLI` pre-trims proofs that are too large to be useful. For a close match it kept proofs with `exFee <= amountToSend`. But `exFee` floors the proof's own fee, while the real fee ceils once over the whole selection, so a proof whose `exFee` equals the target can net a unit below it. Such a proof capped the pool, every larger proof was trimmed away, and the trial loop then had nothing that could cover the target.

Concrete case, a mint at 100 ppk, wallet holding a 32 and an 8:

```ts
wallet.selectProofsToSend(proofs, 8, true); // includeFees
// before: { send: [] }
// after:  { send: [32] }
```

The 8 has `exFee` 8 but nets 7 against a target of 8, so the 32 is the only workable input. Callers see this as `Not enough funds available to send` from `send`/`prepareSwapToSend`, which is misleading: the funds are there.

The exact-match branch immediately above already trims against `amountToSend + 1`, with a comment describing this same floor-versus-ceil slack. This applies the same bound to the close-match branch.

Found while testing a wallet against a 100 ppk mint; it is not specific to any keyset version.

### Tests

New case in `test/wallet/selectProofsRGLI.test.ts` pinning the `[32, 8]` shape at 100 ppk. `npm run prtasks` green.

### DIFF
--- a/src/wallet/SelectProofs.ts
+++ b/src/wallet/SelectProofs.ts
@@ -180,7 +180,11 @@ export function selectProofsRGLI(
       const rightIndex = binarySearchIndex(spendableProofs, targetAmountBig + 1n, true);
       endIndex = rightIndex !== null ? rightIndex + 1 : 0;
     } else {
-      const biggerIndex = binarySearchIndex(spendableProofs, targetAmountBig, false);
+      // The same floor-vs-ceil slack as exact match above: a proof with exFee equal to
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #1002

PR #1003, opened 2026-08-27
feat(model): expose NUT-06 urls, time and tos_url on MintInfo

robwoodgate:
Types and exposes the remaining NUT-06 top-level fields: `urls`, `time` and `tos_url`.

## Why
NUT-06 defines these optional fields and some mints send them, but `GetInfoResponse` did not declare them. They survived into `MintInfo.cache` at runtime yet were invisible to TypeScript consumers without a cast.

## Changes
- `GetInfoResponse` gains optional `urls: string[]`, `time: number` (Unix seconds) and `tos_url: string`.
- `MintInfo` gains passthrough getters for these plus the already-typed `icon_url` (which had no accessor). `urls` returns a snapshot like the other array getters.
- Tests cover exposure, cache round-trip and the omitted case; API report updated.

## Impact
Additive, no behavior change. The library still does nothing with `urls`: whether to trust or switch to an advertised endpoint is a consumer decision, and the getter's doc says so.

### DIFF
--- a/etc/cashu-ts.api.md
+++ b/etc/cashu-ts.api.md
@@ -541,6 +541,9 @@ export type GetInfoResponse = {
     description?: string;
     description_long?: string;
     icon_url?: string;
+    urls?: string[];
+    time?: number;
+    tos_url?: string;
     max_array_length?: number;
     contact: MintContactInfo[];
     nuts: {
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #1004

PR #1032, opened 2026-08-29
Authenticate protected websockets in-band

crodas:
## Why

Cash-TS implementation of the NUT improvement https://github.com/cashubtc/nuts/pull/413

A mint can protect GET /v1/ws under NUT-21 or NUT-22, but the WebSocket path sent no credentials, so every NUT-17 subscription failed against an auth-enabled mint. Headers are not an option: browsers cannot set them and the library injects a standard WebSocket constructor, so no target can. Send the NUT-22 authenticate command on the open socket instead.

The token is fetched lazily, just before the first subscribe, because it is spent the moment it leaves the pool: a connection that never subscribes costs nothing. The mint's reply is awaited so a rejected token fails where it happens rather than as a 31001 on the subscribe that follows. Consecutive failures are bounded so a mint that keeps rejecting cannot drain the pool one reconnect at a time.

Clear auth travels in the same command as the raw access token. That is forward looking: mints today expect a CAT in the upgrade header and refuse a header-less upgrade outright.

### DIFF
--- a/src/mint/Mint.ts
+++ b/src/mint/Mint.ts
@@ -1074,7 +1074,11 @@ class Mint {
       const wsUrl = mintUrl.toString();
 
       if (!this.ws) {
-        this.ws = new WSConnection(wsUrl, this._logger);
+        this.ws = new WSConnection(
+          wsUrl,
+          this._logger,
+          this._authProvider ? () => this.getWsAuthToken() : undefined,
+        );
       }
 
       await this.ws.ensureConnection();
@@ -1147,6 +1151,23 @@ class Mint {
     if (!info.requiresBlindAuthToken(method, path)) return undefined;
     const bat = await this._authProvider.getBlindAuthToken({ method, path });
     return bat;
+  }
+
+  /**
+   * Returns the token for the in-band NUT-17 WebSocket `authenticate` command, or undefined when
+   * the mint does not protect `/v1/ws`.
+   *
+   * @remarks
+   * A blind auth token is preferred because the command carries exactly one credential and the mint
+   * recognises a BAT by its `authA` prefix. The clear auth branch is forward looking: mints today
+   * expect a CAT in the upgrade header and refuse a header-less upgrade outright, which fails
+   * before any frame can be sent, so it only engages on a mint that accepts one.
+   */
+  private async getWsAuthToken(): Promise<string | undefined> {
+    const info = await this.getLazyMintInfo();
+    const bat = await this.handleBlindAuth('GET', '/v1/ws', info);
+    if (bat) return bat;
+    return this.handleClearAuth('GET', '/v1/ws', info);
   }
 
   private async requestWithAuth<T>(
--- a/src/transport/WSConnection.ts
+++ b/src/transport/WSConnection.ts
@@ -1,5 +1,5 @@
 import { type Logger, NULL_LOGGER } from '../logger';
-import { CTSError } from '../model/Errors';
+import { CTSError, WsAuthError, WsRpcError } from '../model/Errors';
 import { type JsonRpcMessage, type JsonRpcReqParams, type RpcSubId } from '../model/types';
 import { generateUuidV7 } from '../utils/uuid.js';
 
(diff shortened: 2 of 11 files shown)


src/transport/WSConnection.ts
```diff
@@ -302,6 +336,133 @@ export class WSConnection {
...
+   * Fetches a token and authenticates, counting the attempt against the failure bound.
+   *
+   * @remarks
+   * A supplier failure is not counted: it throws before a token leaves the pool, so nothing was
+   * spent and retrying cannot drain it. Everything after that is counted once, in a single place,
+   * because `sendRpcMessage` already fails pending RPCs before rethrowing and a narrower catch
+   * would count a send failure twice.
+   */
+  private async authenticate(timeoutMs: number): Promise<void> {
+    if (this.consecutiveAuthFailures >= MAX_CONSECUTIVE_AUTH_FAILURES) {
```
robwoodgate, line 379: [P1] Reset terminal auth state when creating a fresh connection

`consecutiveAuthFailures` persists across WebSocket instances and eventually rejects before requesting another token. `Mint.disconnectWebSocket()` retains this same `WSConnection`, so disconnecting and reconnecting cannot recover: after three temporary timeouts, server restarts, or rejections, all future wallet subscriptions fail until the consumer constructs a new `Mint`. Please remove the cross-connection cap, reset it for each socket, or discard the `WSConnection` on disconnect.

src/transport/WSConnection.ts
```diff
@@ -302,6 +336,133 @@ export class WSConnection {
...
+
+    const socket = this.ws;
+    if (socket?.readyState !== this._WS.OPEN) {
+      this._logger.error('Attempted ensureAuthenticated, but socket was not open');
+      throw new CTSError('Socket not open');
+    }
+
+    if (this.authState?.socket === socket) return this.authState.promise;
+
+    const promise = this.authenticate(timeout);
```
robwoodgate, line 364: [P2] Bind the authentication attempt to this captured socket

The socket is captured here only for the cache key; `authenticate()` later awaits the token supplier and `sendAuthenticate()` sends through the then-current `this.ws`. If this socket closes during token acquisition and a reconnect starts, the old attempt can send its BAT on the replacement socket alongside the replacement socket’s own attempt, consuming two BATs and issuing two authenticate commands for one connection. Please thread the captured socket through the attempt and reject it if that socket is no longer current.

src/mint/Mint.ts
```diff
@@ -1149,6 +1153,23 @@ class Mint {
...
+   * A blind auth token is preferred because the command carries exactly one credential and the mint
+   * recognises a BAT by its `authA` prefix. The clear auth branch is forward looking: mints today
+   * expect a CAT in the upgrade header and refuse a header-less upgrade outright, which fails
+   * before any frame can be sent, so it only engages on a mint that accepts one.
+   */
+  private async getWsAuthToken(): Promise<string | undefined> {
+    const info = await this.getLazyMintInfo();
+    const bat = await this.handleBlindAuth('GET', '/v1/ws', info);
+    if (bat) return bat;
+    return this.handleClearAuth('GET', '/v1/ws', info);
```
robwoodgate, line 1170: [P2] Do not send a CAT in the NUT-22 authenticate command

cashubtc/nuts#413 defines `params.token` specifically as a serialized `authA…` BAT. The in-band command is not defined for NUT-21 clear auth. A NUT-21-only mint will either reject the headerless upgrade or has no obligation to accept this command/CAT, and this also places a reusable CAT into a command intended for a single-use BAT. Please consult only the NUT-22 protected endpoints here and return `undefined` otherwise.

If we wish to keep this forward-looking extension, it should be specified in the NUT first.

robwoodgate (commented):
Agent review against cashubtc/nuts#413 found three correctness issues that should be addressed before merge.

robwoodgate:
@crodas — Nice work on your first Cashu-TS PR. This is a thoughtful, robust implementation, and the failure-path coverage is impressive. I ran an agent review against NUTS #413 and it surfaced a few lifecycle and protocol points inline. Once those are resolved, I’d also like to take a light simplification pass (not a rewrite), just checking whether some of the auth-specific state and policy machinery can be reduced.

PR #1039, opened 2026-09-03
fix(transport): fail rather than follow redirects on requests with a body

robwoodgate:
A mint request that carries a body now fails on an HTTP redirect instead of following it, matching what already happened for requests carrying an auth token.

## Why

A 307 or 308 preserves the method and the body, so following one re-sends the whole request to the redirect target. `redirect: 'error'` was enforced only for requests carrying NUT-21/22 `Blind-auth` / `Clear-auth` headers; every other request used the fetch default of `follow`. The body is the more consequential half of a mint request, and it had no such guard.

## Changes

- `redirect: 'error'` is now the default whenever a request has a body. It sits before the caller's fetch options, so a deployment that legitimately redirects can still override it. The auth-header guard sits after and stays absolute.
- `Mint.requestWithAuth` mirrors the same predicate, so consumers passing a custom `RequestFn` inherit the invariant.
- Bodiless GETs are untouched and still follow redirects, so http to https upgrades and trailing-slash normalisation on `/v1/info` and `/v1/keys` keep working.

## Impact

No behaviour change for a mint reached through a reverse proxy, which passes requests through rather than redirecting. A mint origin answering 307/308 to a POST now fails immediately instead of silently completing elsewhere. A 301/302/303 already broke POSTs, since fetch rewrites those to GET and drops the body, so nothing that worked before stops working. Consumers who need the old behaviour can pass `redirect: 'follow'` per call or globally.

### DIFF
--- a/src/mint/Mint.ts
+++ b/src/mint/Mint.ts
@@ -1177,8 +1177,9 @@ class Mint {
       endpoint: joinUrls(this._mintUrl, path),
       method,
       headers,
-      // A CAT/BAT must never be replayed to a redirect target: fail rather than follow.
-      ...(bat || cat ? { redirect: 'error' as const } : {}),
+      // A redirect target must never receive a replay of the request: bodies carry
+      // proofs and headers carry CAT/BAT tokens. Fail rather than follow.
+      ...(bat || cat || init.requestBody ? { redirect: 'error' as const } : {}),
       ...(nut19?.supported && nut19.params ? nut19.params : {}),
       onResponseMeta: this._captureResponseMetadata,
     });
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #1041 with remaining conflicts!

```bash
git fetch origin backport-1039-to-v4-dev
git worktree add --checkout .worktree/backport-1039-to-v4-dev backport-1039-to-v4-dev
cd .worktree/backport-1039-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 8bc33833e5235b15686f5664e81edee924be6121
```

PR #1040, opened 2026-09-03
fix(wallet): omit unselected proofs from serialized swap previews

robwoodgate:
A serialized swap preview no longer carries the proofs that were not selected for the swap.

## Why

`unselectedProofs` take no part in the NUT-19 replay: `completeSwap` passes them straight through to the returned `keep` array without touching them. Persisting them writes spendable proofs to a second location for no benefit, and nothing in the docs said the serialized blob was sensitive at all.

## Changes

- `unselectedProofs` removed from `SerializedSwapPreview`, in both `serializeSwapPreview` and `deserializeSwapPreview`.
- The serializer docblock and the send, receive and NUT-19 recipes now say a serialized preview holds `inputs` in the clear and needs the same protection as the proof database.
- The send recipe already returned unselected proofs to storage before serializing; that comment now explains why it matters.

## Impact

`SerializedSwapPreview` loses one optional field. Previews stored by 4.10.0 still deserialize, as the extra key is ignored. Callers who replay a persisted preview get `keep` without the unselected proofs and should return those to storage themselves, which is what the send recipe already showed.

### DIFF
--- a/docs-src/usage/nut19.md
+++ b/docs-src/usage/nut19.md
@@ -149,7 +149,8 @@ const receivePreview = await wallet.prepareSwapToReceive(token);
 const sendPreview = await wallet.prepareSwapToSend(21, proofs, { includeFees: true });
 
 // Swap previews have a serialize helper: persist the JSON-safe form, then
-// rehydrate it with deserializeSwapPreview to replay after a restart.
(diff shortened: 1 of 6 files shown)


robwoodgate:
- #1043

PR #1044, opened 2026-09-03
fix(nut20): match quote lock keys by x coordinate

robwoodgate:
`findSigningKey` compared compressed pubkeys byte for byte. A key imported from an x-only context (any nostr key, including NIP-61 nutzap keys) is published as `02 || x`, but its scalar derives the odd-y twin half the time, so minting a locked quote failed with "No private key matches quote pubkey" for half of all such users.

The fix matches on the x coordinate and returns the scalar for the published parity (`n - d` when the derived point is the other twin), so the NUT-20 signature verifies against the pubkey the quote names. Exact matches behave as before.

Tests cover the exact match, the negated-scalar path across several keys, and the no-match error.

### DIFF
--- a/src/crypto/core.ts
+++ b/src/crypto/core.ts
@@ -152,16 +152,25 @@ function toXOnlyPubkey(pubkey: string): string {
 /**
  * Find the private key that can sign for a given compressed public key.
  *
+ * @remarks
+ * Matches on the x coordinate: a key imported from an x-only context (any nostr key) is published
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #1045 with remaining conflicts!

```bash
git fetch origin backport-1044-to-v4-dev
git worktree add --checkout .worktree/backport-1044-to-v4-dev backport-1044-to-v4-dev
cd .worktree/backport-1044-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 0f4d0253ef23953cb5a2211510538df850f6b5f3
```

PR #1060, opened 2026-09-04
chore: drop the unread release-please config

robwoodgate:
Deletes `release-please-config.json`, which release-please has never read on any branch.

## Why

The workflow passes `release-type: node` to `googleapis/release-please-action@v4`, which builds its configuration from the action inputs and never opens the config file. Its `changelog-sections` block has therefore had no effect anywhere, while looking exactly like the file you would edit to change release behaviour. It has already cost two rounds of debugging on v3-dev.

Run logs confirm it: an inputs-mode run never prints `Fetching release-please-config.json`, whereas a manifest-mode dry run does.

Making it live instead would mean switching to manifest mode, which also needs `include-component-in-tag` moved into the file (the tags here have no component prefix) and leaves two config files per branch to keep in sync across three release branches. That buys a changelog heading. Not worth it for a single-package repo.

## Changes

Removes the file. `.release-please-manifest.json` stays: that one is read and updated on every release.

## Impact

None on release behaviour, since nothing read it. Needs backporting so the same trap is not left sitting on the LTS branches.


robwoodgate:
- #1066

robwoodgate:
Successfully created backport PR for `v3-dev`:
- #1067

Issue #1079, opened 2026-09-05
v4 Wallet: melt change reconstruction crashes on zero-value returned signatures (NUT-08) — MeltChangeError on a PAID melt


Amperstrand:
## Description

`Wallet.meltProofsBolt11()` with NUT-08 change fails during client-side reconstruction when the mint returns **zero-value blind signatures** for the change blanks: `OutputData.toProof()` looks up `keyset.keys[amount.toString()]` for the returned signature amount, and for `amount: 0` there is no key — `pointFromHex(undefined)` throws (`Expected hex string, got undefined`). The throw is wrapped into `MeltChangeError` by `completeMelt`, even though the melt itself **settled and was paid** on-chain.

Per NUT-08, Alice sets each blank output's amount to `0` and Bob imprints the real refund amount; Bob *"should omit all blind signatures with value 0"*. A wallet that receives a zero-value signature anyway (mint-side omission is a SHOULD, not a MUST) should ignore it rather than crash the whole change reconstruction — especially since the melt is irreversible at that point and the error is indistinguishable from an actual payment failure for most integrators.

## Environment

- cashu-ts 4.10.0 (verified the v4.10.0...v4.10.1 diff does not touch this path)
- Any mint that returns an `amount: 0` entry in the melt response `change` array (we hit this in production against our mint before it began omitting zero-value signatures; NUT-08 allows a conformant mint to include them)

## Repro

1. Wallet with proofs, melt quote with change expected (`inputs total > amount + fee_reserve`).
2. `prepareMelt` builds the blanks — all with `amount: 0` (spec-correct) — count `max(ceil(log2(overpay)), 1)`.
3. Mint returns `change: [{ amount: 0, C_: "...", id: "..." }, ...]` (zero-value signature included).
4. `completeMelt` → `createMeltChangeProofs(outputData, change)` → `validateReturnedSignatures(..., { checkAmounts: false })` passes → `e[n].toProof(t, keyset)`:

```js
// lib dist, OutputData.toProof
const r = e.amount.toString();        // "0"
const i = pointFromHex(t.keys[r]);    // t.keys["0"] === undefined → throws
```

Minimal sketch:

```js
const outputData = wallet.createOutputData(0, keyset, { denominations: [0] });
try {
  wallet.createMeltChangeProofs(outputData, [{ amount: 0, C_: validC_, id: keyset.id }]);
} catch (e) {
[...]

PR #1084, opened 2026-09-05
fix(utils): guard the keysetIds argument of getDecodedToken

robwoodgate:
`getDecodedToken` now fails with a message naming its required `keysetIds` argument instead of a `TypeError` from inside the keyset id mapping. Refs #1079 (secondary report).

## Why

The second argument has been required since v4 and the migration guide covers it, but plain JS callers on the old single-argument form get `Cannot read properties of undefined (reading 'map')` with no hint at the cause. Public functions are called from JS as well as TS, so the boundary should say what is missing.

## Changes

One `Array.isArray` check at the top of `getDecodedToken`, throwing a `CTSError` that names the argument and points at the migration guide and `wallet.decodeToken()`. Correct callers are unaffected.

## Impact

None for callers passing an array. Callers already broken get a readable error.

### DIFF
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -362,6 +362,12 @@ function tokenFromTemplate(template: TokenV4Template): Token {
  * @returns Cashu token object.
  */
 export function getDecodedToken(tokenString: string, keysetIds: readonly string[]): Token {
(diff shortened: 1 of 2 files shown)


robwoodgate:
- #1085

PR #1099, opened 2026-09-08
fix(nut26): reject repeated singular TLV tags and untyped policy flags

robwoodgate:
A creqB carrying a repeated field, or a payment request whose policy flags are not booleans, is now rejected rather than read differently by different decoders.

## Why

`decodeTLV` kept the last value of a repeated singular tag, so two readers of one creqB could disagree on its id, amount, unit or transport target. [NUT-26](https://github.com/cashubtc/nuts/blob/main/26.md) marks only `mint`, `transport`, `supported_method` and tag tuples as repeatable, and cdk's decoder already rejects a repeat of anything else. The `s` and `mp` flags went through `Boolean()`, so a string such as `'false'` read as `true`.

## Changes

A repeated singular tag, a repeated transport `kind`/`target` and a second nprofile identity pubkey now throw. `encodeTLV` refuses a non-`sat` unit whose encoding is the single `0x00` byte NUT-26 reserves for `sat`; decoding is unchanged, since `0x00` already maps to `sat`.

The flags accept `true`/`false`/`0`/`1`. CBOR `null` reads as absent, which is how cdk serialises an unset `Option` field: its canonical NUT-18 vector carries `s: null` and `d: null`, and is added as a fixture.

## Impact

Inputs that were previously misread now throw a `CTSError` from `decodePaymentRequest` and `PaymentRequest.fromRawRequest`. No API surface change.

### DIFF
--- a/src/model/PaymentRequest.ts
+++ b/src/model/PaymentRequest.ts
@@ -43,6 +43,21 @@ import { CTSError } from './Errors';
 import type { Proof } from './types/proof';
 
 /**
+ * Normalizes a wire boolean flag, admitting true/false, the numeric 0/1 form, and null.
+ *
+ * @remarks
+ * `null` is treated the same as an absent key: some encoders (eg cdk) serialize an unset optional
+ * field as CBOR null rather than omitting it.
+ * @throws If `value` is present, non-null, and not one of those four values.
+ */
+function normalizePaymentRequestFlag(value: unknown, field: 's' | 'mp'): boolean | undefined {
+  if (value === undefined || value === null) return undefined;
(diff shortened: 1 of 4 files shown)


robwoodgate:
- #1100 with remaining conflicts!

```bash
git fetch origin backport-1099-to-v4-dev
git worktree add --checkout .worktree/backport-1099-to-v4-dev backport-1099-to-v4-dev
cd .worktree/backport-1099-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 1fa657ea0214fe9256b4a91e3964255749042868
```

PR #1101, opened 2026-09-08
fix(nut11): reject malformed P2PK tag values and bound key lists

robwoodgate:
A P2PK secret whose `n_sigs`, `n_sigs_refund` or `sigflag` tag is malformed, or carries more than one value, is now rejected instead of being read as if the tag were absent.

## Why

`getTagInt` returned `undefined` for a value that failed to parse, and every caller then applied its default, so a malformed `n_sigs` read as 1-of-n. `getTagScalar` returned the first of several values, so two readers of one secret could disagree. [NUT-11](https://github.com/cashubtc/nuts/blob/main/11.md) requires a non-positive-integer `n_sigs` or `n_sigs_refund` to be rejected as unspendable, and cdk already errors on both shapes.

## Changes

The arity and integer checks live in the two accessors every consumer reads through, so there is no second parser left to differ. `"01"` and JSON numbers still parse.

`locktime` is the deliberate exception: NUT-11 says a value that is not a valid unix time makes the lock permanent, so `getLocktime` reads it that way and a foreign proof is passed on for the mint to judge. Lock creation from a payment request stays strict.

`MAX_P2PK_PUBKEYS` and `MAX_P2PK_SIGNATURES` move to `utils/limits.ts`; the key bound now also applies to a payment request's `pubkeys` and `refund` tags before any per-key curve work; it sits above the 11-slot rule NUT-28 sets, which the builder enforces on creation. A nutroot blind-me list is bounded by the requested tree's own key count instead, since a tree can carry far more keys than a P2PK tag.

## Impact

`verifyP2PKSpendingConditions`, `getP2PKExpectedWitnessPubkeys`, `Wallet.spendOptions` and `PaymentRequest.toP2PKOptions` now throw a `CTSError` on the malformed shapes above; all already documented `@throws` for a malformed secret. No API surface change.

### DIFF
--- a/src/crypto/NUT10.ts
+++ b/src/crypto/NUT10.ts
@@ -211,32 +211,42 @@ export function getTag(secret: Secret | string, key: string): string[] | undefin
 }
 
 /**
- * Get the first scalar value of a tag as a string, or undefined if missing.
+ * Get the single value of a scalar tag as a string, or undefined if missing.
  *
  * @param secret - The Proof secret.
  * @param key - Tag key to lookup.
  * @returns - Tag value or undefined if not present.
+ * @throws If the tag carries more than one value.
  */
 export function getTagScalar(secret: Secret | string, key: string): string | undefined {
   const vals = getTag(secret, key);
-  return vals && vals.length > 0 ? vals[0] : undefined;
+  if (vals === undefined) return undefined;
+  if (vals.length !== 1) {
+    throw new CTSError(`Invalid NUT-10 tag "${key}": must carry a single value`);
+  }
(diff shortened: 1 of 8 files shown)


robwoodgate:
- #1102 with remaining conflicts!

```bash
git fetch origin backport-1101-to-v4-dev
git worktree add --checkout .worktree/backport-1101-to-v4-dev backport-1101-to-v4-dev
cd .worktree/backport-1101-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x f088921bccbe27fd50605fe66c9a11d27403e34a
```

PR #1104, opened 2026-09-09
fix(nut11): bind SIG_INPUTS signatures to the secret and honour overlapping HTLC refunds

robwoodgate:
A SIG_INPUTS signature is now always made and checked over the proof secret, an HTLC whose sender key is also a receiver key can be refunded after expiry, and P2PK/HTLC secrets and witnesses are size-bounded before they are parsed.

## Why

`signP2PKProof` accepted a `message` override for a SIG_INPUTS proof and the verifiers used one when given, so signer and verifier could disagree on what a signature covers. [NUT-11](https://github.com/cashubtc/nuts/blob/main/11.md) fixes the message as `Proof.secret` unless the sigflag says otherwise. Separately, an HTLC with the same key in `pubkeys` and `refund` reported `MAIN` from the P2PK verdict, so its refund path was never reached; cdk selects the refund path whenever the preimage check fails.

## Changes

Signing is symmetric: a message on SIG_INPUTS throws, a missing one on SIG_ALL throws. Verification derives the message from the sigflag. `verifyHTLCSpendingConditions` lost a `message ?? secret` default that had let a SIG_ALL HTLC verify with no message.

`createHTLCsecret` validates and lowercases the hashlock. Verification deliberately does not shape-check it: the reference mints authorise the expiry path without inspecting `Secret.data`, so the wallet's verdict now matches theirs, and a garbage hash simply never matches a preimage.

`MAX_SECRET_LENGTH` moves to `utils/limits.ts` and applies before parsing; `MAX_WITNESS_LENGTH` (16,384) bounds witnesses. A maximal 11-slot lock is 968 characters, a 64-signature witness plus preimage 8,478. Malformed witness JSON reads as no witness, without a `console.error`.

## Impact

`signP2PKProof` throws on the two mismatched sigflag/message shapes; `signP2PKProofs` warns and skips instead. `attachHTLCPreimage` throws on an oversized existing witness. No API surface change.

### DIFF
--- a/src/crypto/NUT11.ts
+++ b/src/crypto/NUT11.ts
@@ -5,7 +5,12 @@ import { type Logger, NULL_LOGGER } from '../logger';
 import { CTSError } from '../model/Errors';
 import { type OutputDataLike } from '../model/OutputData';
 import { type HTLCWitness, type P2PKWitness, type Proof } from '../model/types';
-import { MAX_P2PK_PUBKEYS, MAX_P2PK_SIGNATURES } from '../utils/limits';
+import {
+  MAX_P2PK_PUBKEYS,
+  MAX_P2PK_SIGNATURES,
+  MAX_SECRET_LENGTH,
+  MAX_WITNESS_LENGTH,
+} from '../utils/limits';
 import { type NUT10Option } from '../wallet/types/payment-requests';
 
 import { getValidSigners, schnorrSignMessage, schnorrVerifyMessage, type PrivKey } from './core';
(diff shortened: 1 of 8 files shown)


robwoodgate:
- #1105 with remaining conflicts!

```bash
git fetch origin backport-1104-to-v4-dev
git worktree add --checkout .worktree/backport-1104-to-v4-dev backport-1104-to-v4-dev
cd .worktree/backport-1104-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 52fc6c2628ab42f78e798a134894f0121ebddd96
```

PR #1130, opened 2026-09-10
fix(utils): guard missing DLEQ blinding factor on v4 token decode

Hardeezah:
Decode v4 tokens whose DLEQ proof omits the optional blinding factor `r`, preventing a raw `TypeError` crash on valid peer tokens.

## Why

`tokenFromTemplate` called `bytesToHex(p.d.r)` without guarding for `undefined`. NUT-00 and NUT-12 make `r` optional in the DLEQ proof, it is only needed for reblinding. Peer implementations (e.g. cdk) ship tokens whose DLEQ block contains only `{e, s}`.

The encode path already tolerates this (`p.dleq.r ?? '00'`); the decode path did not, creating an asymmetric wire contract that caused `getDecodedToken`, `getTokenMetadata`, and `wallet.receive` to crash on valid tokens. `findCashuPayload` silently swallowed the error, making valid pasted/QR-scanned tokens unfindable.

## Changes

- `src/utils/core.ts:392` - guard `p.d.r` with `?? new Uint8Array([0])`, matching the encode side's fallback behavior.
- `src/model/types/token.v4.ts:13` - mark `V4DLEQTemplate.r` as optional (`r?: Uint8Array`) so the type reflects the CBOR wire format from other implementations.

## Impact

Non-breaking. Consumers decoding v4 tokens from other Cashu implementations will no longer crash on well-formed tokens with an omitted DLEQ `r` field. Existing tokens with `r` present continue to decode identically. No public API surface change.

### DIFF
--- a/src/model/types/token.v4.ts
+++ b/src/model/types/token.v4.ts
@@ -10,7 +10,7 @@ export type V4DLEQTemplate = {
   /**
    * Blinding factor.
    */
-  r: Uint8Array;
+  r?: Uint8Array;
 };
 
 /**
--- a/src/utils/core.ts
+++ b/src/utils/core.ts
@@ -377,7 +377,7 @@ function tokenFromTemplate(template: TokenV4Template): Token {
         id: bytesToHex(t.i),
         ...(p.d && {
           dleq: {
-            r: bytesToHex(p.d.r),
+            r: bytesToHex(p.d.r ?? new Uint8Array([0])),
             s: bytesToHex(p.d.s),
             e: bytesToHex(p.d.e),
           },


robwoodgate:
Thank you for making your first contribution! You've certainly found a couple of bugs - that `TypeError: expected Uint8Array, got type=undefined` should definitely be guarded / impossible, and the `?? '00'` fudge in the encoder side is also incorrect.

Your rationale and suggested solution, are unfortunately not correct, and stem from a misunderstanding of the NUT spec, which is understandable as the spec is unfortunately slightly ambiguous.

- NUT-00 says the DLEQ `d` "block" is optional.  If included, however, all three elements are required.

```json
"d": { <optional> // optional DLEQ proof
    "e": bytes,
    "s": bytes,
    "r": bytes
},
```

- NUT-12 says that a returned `BlindSignature` from Mint to User has just `e` (challenge) and `s` (response), which is correct, because a mint is not privy to the blinding factor `r`.

The **WALLET** is expected to add its `r` (blinding factor) when creating the `Proof` object (per User to User section). 

So a wallet that creates a `Proof` with just `e` and `s` is **BUGGY** and not to spec. We should NOT fudge a `00` blinding factor, because that a) creates a DLEQ that will not verify and b) suggests the original `B_` (blinded message) was not in fact blinded (`r =0 means B = B_`). 

CDK is correct  at both mint and wallet side, so your experience must be a buggy wallet talking to a CDK mint and not hydrating Proofs properly.

The ambiguity in NUT-12 is the privacy note:

```
**Privacy:** The blinding factor `r` should not be shared with the mint
or otherwise, the mint will be able to associate the `BlindSignature`
with the `Proof`.
```

That implies it's ok to keep `e` and `s` (a partial DLEQ) in the `Proof`, which is incorrect. 

If a wallet sends `e` and `s` back to the mint, it is also a privacy leak as the mint can calculate the DLEQ nonce used:

```
s = k + e·a  mod n
k = s - e·a  mod n
```

Once it has `k`, it can, if it wishes, sample the `B_`'s in its issuance table to find the one that links the secret:

```
C' = a·B'
R1 = k·G
R2 = k·B'
test  hash(R1, R2, A, C') == e
```

That is why the core implementations (Nutshell, CDK, Cashu-TS) **REMOVE** the entire DLEQ block when sending proof inputs for swap and melt. 

In summary, there should NEVER be a situation where a `Proof` validly carries only `e` and `s` and not `r`.

Your suggestion would just bake in tolerance of a buggy and privacy leaking wallet.

So NACK for this contribution, I'm afraid, am closing in favour of the solution in #1135

PR #1139, opened 2026-09-13
chore(ci): run the auth demo against a live mint

robwoodgate:
Runs the auth demo in CI, so the blind auth path has live coverage for the first time.

## Why

Every test under `test/auth` mocks fetch, and neither integration suite mentions auth, so nothing exercised NUT-21 / NUT-22 against a real mint and a real OIDC provider. A regression there surfaced only when someone next ran the demo by hand. #1137 is the example: the demo had been broken since mintd 0.18 shipped and CI stayed green throughout.

## Changes

A PR job runs `make demo-ci` against the pinned mint image, so a failure implicates the PR that caused it. Upstream drift is the other half and no PR can catch it, so the nightly job runs the same demo with the image overridden to `:latest`, matching how the integration mints are already covered; it opens or comments on a `drift` issue as the existing job does.

The demo waited on a fixed sleep, fine locally and a coin toss on a cold runner. It now polls the discovery document and the mint's `/v1/info`, and dumps container logs on timeout.

## Impact

One job per PR, measured at 23s plus install. `MINT_IMAGE` overrides the compose pin. Local `make up` and the nutshell targets share the same poll.


robwoodgate:
- #1140 with remaining conflicts!

```bash
git fetch origin backport-1139-to-v4-dev
git worktree add --checkout .worktree/backport-1139-to-v4-dev backport-1139-to-v4-dev
cd .worktree/backport-1139-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 12ce8b240b85f95f03eac51add53f238c385c2fa
```

PR #1153, opened 2026-09-16
fix(utils): resolve short keyset IDs in getDecodedTokenBinary

### LINKED ISSUE #1134: getDecodedTokenBinary does not resolve short keyset IDs
**Describe the bug**

`getDecodedTokenBinary` does not resolve short keyset IDs, so a binary (`craw`) token from another implementation can decode to proofs whose `id` is still the 8-byte short form.

[NUT-00](https://github.com/cashubtc/nuts/blob/main/00.md) makes the short keyset ID optional for senders ("wallets **MAY** use the short keyset ID representation") but mandatory for receivers:

> Wallets receiving a Token **MUST** support both short and full keyset ID representations. When a short keyset ID is encountered, the wallet **MUST** resolve it to the corresponding full keyset ID before processing the contained `Proof` objects.

A binary token is the same object as a `cashuB` token, just framed differently (`utf8("craw") || utf8("B") || cbor(token_v4_object)`), so the same rule applies to both.

`getDecodedToken` resolves via `mapShortKeysetIds`, which is its only caller. `getDecodedTokenBinary` returns straight from the CBOR reader, and its signature takes no keyset ID list, so it has nothing to resolve against.

**To Reproduce**

Decode a binary token whose `i` field holds the 8-byte prefix of a v1 (or later) keyset ID:

```ts
const token = getDecodedTokenBinary(bytes);
token.proofs[0].id; // 16 hex chars, not the 66-char ID the mint's endpoints use
```

Anything keyed on the full ID then misses.

**Expected behavior**

Short IDs resolve the same way they do for `cashuB` tokens, including the ambiguity rule: a short ID matching more than one known keyset must fail parsing.

**Additional context**

Not reachable through tokens this library produces: `getEncodedTokenBinary` emits full IDs, and v0 keyset IDs are 8 bytes in full form, so neither needs resolving. It affects binary tokens from implementations that do use the short form on v1+ keysets.

Fixing it changes a public signature, since the decoder needs the keyset ID list `getDecodedToken` already takes. An optional second parameter keeps current callers compiling. Worth deciding before the v5 line goes GA.

Hardeezah:
Closes #1134

`getDecodedTokenBinary` now resolves 8-byte short keyset IDs, matching how `cashuB` tokens are handled.

- Adds an **optional** `keysetIds` second argument (`readonly string[]`, the same list `getDecodedToken` takes). When provided, decoded proofs are routed through `mapShortKeysetIds`, so short IDs resolve to their full form — identical logic and ambiguity rule to `cashuB` tokens.
- The parameter is optional so existing single-argument callers keep compiling: binary tokens this library produces always carry full IDs (or 8-byte v0 full-form IDs), so they never need resolving. Omitted, IDs are left untouched.

A binary token is the same `token_v4_object` as `cashuB`, just framed as `utf8("craw") || utf8("B") || cbor(...)` (NUT-00), so the resolution requirement applies to both.

## Tests

- resolves an 8-byte short keyset ID to the full ID when `keysetIds` are given
- throws `Short keyset ID ... cannot be resolved` when the list is empty
- throws on an ambiguous short ID (matches more than one known keyset)
- leaves the short ID untouched when `keysetIds` are omitted (backward compatibility)
- full-ID round-trip through `getEncodedTokenBinary` unchanged

## Verification

prettier, `tsc`, eslint, `api:check` all pass; node test suite `3347/3347` (86 files) green.


robwoodgate (approved):
Thanks for your first contribution! PR ACK.

robwoodgate:
- #1154

PR #1157, opened 2026-09-16
test: raise the vitest timeout and shrink two heavy prepareMint fixtures

robwoodgate:
Node CI intermittently times out on CPU-bound tests. This raises the vitest budget to 20s and shrinks the two prepareMint fixtures that sat closest to the 5s cliff.

## Why

The `prepareMint mutants` tests minted 1000 sats against the shared dummy keyset, which only carries the 1 and 2 denominations. That is 500 blinded outputs at roughly 0.4ms each, about 190ms locally. CI runs the node project alongside three Playwright browsers and coverage on one runner, so CPU-bound tests stretch far past vitest's 5s default. Nothing in these tests waits; they are simply heavy and starved.

## Changes

- `testTimeout: 20_000` on each vitest project (a root-level value did not reach the node project when I checked), the same accommodation #1120 made for one test by hand. A genuine hang still fails, just later.
- The two prepareMint mutant tests mint 8 sats instead of 1000, with the quote amounts and the `amount_paid` assertion kept in step. Each drops from about 250ms to 3ms. The assertions concern null accounting fields, not the amount.

Other CPU-bound tests (BLS batch verify, ScriptPath signing, the v3 OutputData round-trip) sit at 200-330ms locally and are covered by the new budget. The dummy keyset itself is left alone: its ID and pubkeys are hardcoded across many fixtures, so widening its denominations is a separate refactor.


robwoodgate:
- #1158 with remaining conflicts!

```bash
git fetch origin backport-1157-to-v4-dev
git worktree add --checkout .worktree/backport-1157-to-v4-dev backport-1157-to-v4-dev
cd .worktree/backport-1157-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 46e5fc3fda63edc5986c5e6b447dfc46a7b22b46
```

PR #1161, opened 2026-09-16
fix(mint): check a bolt11 melt preimage against the invoice payment hash

robwoodgate:
A bolt11 melt quote or melt response that supplies a `payment_preimage` now has it checked against the invoice's payment hash. A preimage that does not match becomes `null` with a warning.

## Why

`payment_preimage` was passed through as any string. Apps use it as proof of payment, so a value that does not hash to the invoice's payment hash is worse than no value. Nutshell's CLI already does this check. Not every backend returns a preimage, so the check runs only when one is supplied.

## Changes

- `bolt11PaymentHash` and `bolt11PreimageMatches` in utils (`@internal`): bech32-decode the invoice, read the `p` tagged field, compare. No signature check, that is the payer's job. Mint stays free of noble imports.
- `Mint.normalizeMeltBoltFields` verifies the preimage on bolt11 responses that carry both `request` and `payment_preimage`. bolt12 is skipped (its `request` is an offer, not the invoice). An unparseable invoice leaves the preimage as supplied with a debug log.
- Tests: the BOLT 11 spec vector for the reader, and constructed invoices for the mint paths.

## Impact

Consumers reading `payment_preimage` may now see `null` where a mint sent a wrong value, including LND's all-zero not-yet-known string if a mint passes it through. An execution response that omits `request` (cdk does) is not checked at that point; the next quote check, which carries the invoice, is. Candidate for a v4 label backport.


robwoodgate:
- #1165 with remaining conflicts!

```bash
git fetch origin backport-1161-to-v4-dev
git worktree add --checkout .worktree/backport-1161-to-v4-dev backport-1161-to-v4-dev
cd .worktree/backport-1161-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x f9e74b116dff7068d49e4d516f51929921336661
```

PR #1176, opened 2026-09-17
fix(model): require boolean true for advertised NUT support

KvngMikey:
`MintInfo.isSupported()` now requires boolean `true` for generic NUT support. A mint advertising `supported: "false"` previously produced `{ supported: true }`.

## Why

Found while reviewing [Minibits' capability migration](https://github.com/minibits-cash/minibits_wallet/pull/203). Its existing NUT-20 check required `=== true`, but routing it through `MintInfo.isSupported(20)` admitted truthy non-booleans. Checking the returned `.supported === true` cannot recover the original value after coercion. Coco's NUT-11/NUT-20 preflight has the same strict contract.

## Changes

- Use strict equality in the shared generic-NUT check (7, 8, 9, 10, 11, 12, 14, 20).
- Test booleans, missing flags, strings, numbers, objects, arrays, and null; document the behavior.

## Impact

Valid boolean metadata behaves identically. Malformed truthy flags now report unsupported. Method-based capability checks and NUT-04/05 disabled semantics are unchanged; there is no public API shape change.

Rob, could this also receive the `backport v4-dev` label if accepted? Minibits uses v4; the earlier constructor hardening backport (#1117) addresses a separate problem and does not fix this coercion. Consumers will need a release containing this change and a lockfile update.

## Validation

- New malformed-truthy regression cases fail before the fix.
- Lint, changed-file formatting, typecheck, build, and API extraction passed; no API report change.
- Full Node and browser suite: 249 files passed; 9,801 tests passed and 3 skipped. Required Playwright binaries were installed in a temporary directory before the successful run.


docs-src/usage/mint_capabilities.md
```diff
@@ -88,6 +88,10 @@ network calls, use [`wallet.loadMintFromCache(storedInfo, keyChainCache)`](./cre
 
 `isSupported(num)` reports the mint's advertised support for a given NUT.
 
+For NUTs 7, 8, 9, 10, 11, 12, 14, and 20, support requires the boolean `true`.
+Missing flags and non-boolean values, including the strings `"true"` and `"false"`,
+are treated as unsupported.
+
```
robwoodgate:
```suggestion
```

Drop this hunk please. A consumer doesn't need to know that a spec defined `bool` is treated as a `bool`, and the text will age badly as more NUTS are published.

robwoodgate (dismissed):
Great catch, @KvngMikey - thank you!

One minor nit below, otherwise ACK and yes, we will backport to v4.

I note that Nutshell wallet also uses key prescence as true (in `supports_nut`), so you might want to raise a PR there too.  CDK wallet is even stricter - it fails to parse a MintInfo if `supported` contains a string value.

Both reference mints emit a `bool`, so this really only addresses a buggy mint implementation.

docs-src/usage/mint_capabilities.md
KvngMikey: got it, will fix !

KvngMikey:
ready @robwoodgate

robwoodgate:
- #1177

PR #1179, opened 2026-09-17
fix(core): keep a token's witness verbatim when re-encoding

robwoodgate:
`getEncodedToken` stringifies `Proof.witness` unconditionally, but `witness` is typed `string | P2PKWitness | HTLCWitness` and a token decoded with `getDecodedToken` hands back the wire string. So decode then encode nests it:

```
{"signatures":["cdec…"]}          →  "{\"signatures\":[\"cdec…\"]}"
```

Any wallet that decodes a locked token and re-encodes it, to store it, forward it, or merge two tokens, silently corrupts the witness, and the mint then refuses the spend. Found by merging a P2PK-locked token with another one against a live mint.

`Wallet._normalizeWitness` has always had exactly this guard for the mint payload, so spends the wallet itself signs were never affected; the token codec just never got the same rule. The asymmetry dates to "Support for proof witness in token serialization" (2025-03-27) and is present on `v4-dev` too, so a backport is probably worth it.

The existing test only asserted the witness was defined after a round trip, which is how it survived. It now asserts the exact string, and a second test round-trips three times. Both fail without the one-line fix. `npm run prtasks` green.


robwoodgate:
- #1181 with remaining conflicts!

```bash
git fetch origin backport-1179-to-v4-dev
git worktree add --checkout .worktree/backport-1179-to-v4-dev backport-1179-to-v4-dev
cd .worktree/backport-1179-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 356efa0b23d0e9dcc346357efcaf96211071750e
```

PR #1195, opened 2026-09-20
fix(wallet): refuse pre-v3 locks the mint does not advertise

Kelbie:
I was testing a tool that looks for inconsistencies between the NUTs and Cashu implementations, and it flagged this one. It seems like a sensible change to me.

`Wallet` now refuses to build a P2PK or HTLC locked output on a pre-v3 keyset unless the mint advertises NUT-11 (and NUT-14 for a hashlock).

## Why

cashu-ts will currently create locked ecash on a mint that does not support the spending condition. The mint signs outputs blind, so it never sees the lock and nothing fails. The condition is only read at spend time, and [NUT-10](https://github.com/cashubtc/nuts/blob/main/10.md), [NUT-11](https://github.com/cashubtc/nuts/blob/main/11.md) and [NUT-14](https://github.com/cashubtc/nuts/blob/main/14.md) all say what a non-supporting mint may do with it:

> proofs may be treated as regular anyone-can-spend proofs. Applications need to make sure to check whether the mint supports a specific kind of spending condition by checking the mint's info endpoint.

`MintInfo.isSupported(11 | 14)` exists, but nothing consults it before a lock is built.

Where this could matter: a [NIP-61](https://github.com/nostr-protocol/nips/blob/master/61.md) nutzap is P2PK-locked ecash published in a public event, on a mint the recipient chose. If that mint does not enforce P2PK, anyone reading the relay could redeem the token before the recipient. I have not seen this happen; most mints advertise NUT-11.

Some downstream wallets already do this check at the app layer (coco refuses a P2PK send when NUT-11 is not advertised, cashubtc/coco#309).

## Changes

- `Wallet.createOutputData` refuses a `lock` output on a pre-v3 keyset when `isSupported(11)` is false, and when the lock has a hashlock and `isSupported(14)` is false. The `asLocked()` builders (send, receive, mint, melt) all reach this point.
- v3 keysets are not checked. Under the nutroot revision of NUT-10 (cashubtc/nuts#443) every v3 secret is a public key and spending always takes a signature, and the NUT-10 info flag is defined for pre-v3 JSON secrets only.
- Test fixtures that lock on a pre-v3 keyset now advertise NUT-11.

## Scope

- Creation side only. It does not change how received locked tokens are verified or judged.
- Only locks built through `Wallet` (`OutputType` `lock`). `OutputData.createP2PKData()` called directly, and `custom` / `factory` output types, are not checked.

## Impact

No API change. A pre-v3 lock now throws against a mint that does not advertise NUT-11/14. That includes a mint that enforces P2PK but does not advertise it: the mint's info is taken at its word. Building a locked output now needs mint info loaded (`loadMint` / `loadMintFromCache`).

## Validation

Lint, format, typecheck and the Node suite (3,407 tests) pass locally. I did not run the browser suite locally; CI is green on this PR.


Kelbie:
I was testing a tool to try and find inconsistencies between NUTs and various cashu implementations and it found this one which seems like a sensible change to me. 

I think the main point is cashu-ts will currently create locked ecash (P2PK/HTLC) on a mint that doesn't support those spending conditions.

robwoodgate:
- #1196 with remaining conflicts!

```bash
git fetch origin backport-1195-to-v4-dev
git worktree add --checkout .worktree/backport-1195-to-v4-dev backport-1195-to-v4-dev
cd .worktree/backport-1195-to-v4-dev
git reset --hard HEAD^
git cherry-pick -x 167b2bbdb9288a07c90d9c750f0901bb93ec786d
```

robwoodgate:
Nice addition, thanks @Kelbie !

